PR #2, opened 2017-03-01
Allow running basic OpenShift testing scenarios

omron93:
I would like add some simple scripts to test images in OpenShift environment.

It is quite easy to deploy OpenShift with `oc cluster up` and having these simple tests would allow developers/users to easily test changes or to add this testing to CI.

This PR adds new make target `test-openshift` - by it images are built and `test/run-openshift` is run (if exists), it is similar how `make test` is handled.

Testing changes in OpenShift is not necessary for all code changes, so IMPOV it could be better to haveseparate make target for it.

@praiskup Please take a look.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -100,5 +100,14 @@ for dir in ${dirs}; do
     fi
   fi
 
+  if [[ -v TEST_OPENSHIFT_MODE ]]; then
+    if [[ -x test/run-openshift ]]; then
+      VERSION=$dir IMAGE_NAME=${IMAGE_NAME} test/run-openshift
+    else
+      echo "-> OpenShift tests are not present, skipping"
+    fi
+  fi
+
+
   popd > /dev/null
 done
--- a/common.mk
+++ b/common.mk
@@ -28,3 +28,7 @@ build:
 .PHONY: test
 test:
 	$(script_env) TAG_ON_SUCCESS=$(TAG_ON_SUCCESS) TEST_MODE=true $(build)
+
+.PHONY: test-openshift
+test-openshift:
+	$(script_env) TAG_ON_SUCCESS=$(TAG_ON_SUCCESS) TEST_OPENSHIFT_MODE=true $(build)


build.sh
```diff
@@ -100,5 +100,14 @@ for dir in ${dirs}; do
...
 
+  if [[ -v TEST_OPENSHIFT_MODE ]]; then
+    if [[ -x test/run-openshift ]]; then
+      VERSION=$dir IMAGE_NAME=${IMAGE_NAME} test/run-openshift
+    else
+      echo "-> OpenShift tests are not present, skipping"
+    fi
+  fi
+
+
```
praiskup: What I dislike about build.sh script is that it does both building and testing.  And now it would serve another task.  Such things are hard to be parallelized in future.

It would be much nicer if we had self-standing script for this without the need for  TEST_OPENSHIFT_MODE variable.

praiskup: > What I dislike about build.sh script is that it does both building and testing.

By that I mean that usually this is solved a declarative way;  by separate targets in Makefile and dependencies between targets.

omron93: @praiskup Please check the update. Is that what you thought?

I could create separate scripts for normal testing too.

> By that I mean that usually this is solved a declarative way; by separate targets in Makefile and dependencies between targets.

Is somehow possible to have `build:` target. And in `test-openshift:` target have dependency on it but pass some parameters for it.
What I think is setting `TEST_MODE`. When user wants to build images, it should not be set. When using this target as dependency of test targets, it should be set. Is it somehow possible or is is necessary to have separate `build-test` target? Or how to do this?

common.mk
```diff
@@ -5,26 +5,33 @@ ifndef common_dir
...
+build_env = \
 	SKIP_SQUASH=$(SKIP_SQUASH)                      \
-	UPDATE_BASE=$(UPDATE_BASE)                      \
+	UPDATE_BASE=$(UPDATE_BASE)
+
+script_env = \
+	VERSION="$(VERSION)"                            \
 	VERSIONS="$(VERSIONS)"                          \
 	OS=$(OS)                                        \
-	VERSION="$(VERSION)"                            \
```
praiskup: I'm curious why VERSION was moved above;  and while we are on it .. is the VERSION && VERSIONS needed, or is that just leftover?

praiskup:
> I could create separate scripts for normal testing too.

From long term POV, it would be neat.

> Is somehow possible to have build: target. And in test-openshift: target have dependency on it but pass some parameters for it.

It is usually done so `build` depends on `build-VERSIONA` and `build-VERSIONB`;  `test` depends on `test-VERSIONA` and `test-VERSIONB` and `test-VERSIONA` depends on `build-VERSIONA`.

I believe that's what you ask for?

praiskup:
You can of course run `'$(MAKE) sub-target` which preserves make variables, but it is sub-optimal and less readable.

common.mk
omron93: > I'm curious why VERSION was moved above

I was thinking about grouping these two similar variables together:-) Sry, not needed...

> is the VERSION && VERSIONS needed, or is that just leftover

At least one is needed. So what about removing `VERSION`? 
So in case user wants only one version to build, `VERSIONS` variable has to be rewritten. But what about backward compatibility?

omron93:
> It is usually done so build depends on build-VERSIONA and build-VERSIONB; test depends on test-VERSIONA and test-VERSIONB and test-VERSIONA depends on build-VERSIONA.
>
> I believe that's what you ask for?

Probably yes.

I thought about `build:` and `build-test:` targets (see last commit). They are almost same, so I wanted to know if it can be done somehow differently that having two targets differing only in one additional variable. But they are only one liners, so it is not important :-)

omron93:
> From long term POV, it would be neat.

Is separate pull request needed for this, from your point of view?

omron93:
> I thought about build: and build-test: targets (see last commit). They are almost same, so I wanted to know if it can be done somehow differently that having two targets differing only in one additional variable. But they are only one liners, so it is not important :-)

It was not working, so I removed it.

Perhaps it would be easier not to use '-candidate' suffix for image names for testing (it conflicts when running same command in two different directories) and use for example git hash tag for these 'temporary' images.

omron93:
After offline discussion with @praiskup we decided to return back to first implementation - in same way as for already implemented local docker tests.

@praiskup What about target name - is it better to use `-` or `_`?

praiskup:
> @praiskup What about target name - is it better to use - or _

I think it is matter of taste;  but dashes (as you used it) are more common.

omron93:
OK.

Anything else why not to merge?

praiskup:
LGTM, though I'm curious whether we don't want to have `tests` subdir instead of `test`.

praiskup:
Ah, Marek corrected me that the 'test/' dir is relative to the "version" directory.
LGTM -> merging.

PR #7, opened 2017-05-03
Generate help.1 file

omron93:
Generate/update help.1 file from README.md on every build.

@praiskup @pkubatrh Please take a look and merge.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -32,7 +32,6 @@ function docker_build_with_version {
   # Use perl here to make this compatible with OSX
   DOCKERFILE_PATH=$(perl -MCwd -e 'print Cwd::abs_path shift' $dockerfile)
   cp ${DOCKERFILE_PATH} "${DOCKERFILE_PATH}.version"
-  go-md2man -in README.md -out help.1
   echo "COPY help.1 /" >> "${dockerfile}.version"
   git_version=$(git rev-parse --short HEAD)
   echo "LABEL io.openshift.builder-version=\"${git_version}\"" >> "${dockerfile}.version"
--- a/build.sh
+++ b/build.sh
@@ -32,6 +32,8 @@ function docker_build_with_version {
   # Use perl here to make this compatible with OSX
   DOCKERFILE_PATH=$(perl -MCwd -e 'print Cwd::abs_path shift' $dockerfile)
   cp ${DOCKERFILE_PATH} "${DOCKERFILE_PATH}.version"
+  go-md2man -in README.md -out help.1
+  echo "COPY help.1 /" >> "${dockerfile}.version"
   git_version=$(git rev-parse --short HEAD)
   echo "LABEL io.openshift.builder-version=\"${git_version}\"" >> "${dockerfile}.version"
   if [[ "${UPDATE_BASE}" == "1" ]]; then
--- a/common.mk
+++ b/common.mk
@@ -24,13 +24,18 @@ script_env = \
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"
 
 .PHONY: build
-build:
+build: manpages
 	$(script_env) $(build)
 
 .PHONY: test
-test:
+test: manpages
 	$(script_env) TAG_ON_SUCCESS=$(TAG_ON_SUCCESS) TEST_MODE=true $(build)
 
 .PHONY: test-openshift
-test-openshift:
+test-openshift: manpages
 	$(script_env) TAG_ON_SUCCESS=$(TAG_ON_SUCCESS) TEST_OPENSHIFT_MODE=true $(build)
+
+manpages = $(shell for version in $(VERSIONS); do echo "$$version/help.1"; done)
+manpages: $(manpages)
+$(manpages): %help.1: %README.md
+	go-md2man -in "$^" -out "$@"


praiskup:
I would much rather see this implemented in common.mk, do you think this would be possible?

omron93:
@praiskup Why in `common.mk`?

I've put it in `build.sh` because `LABEL io.openshift.builder-version=` is also generated there. IMPOV it is better to "change Dockerfile" in one place.

praiskup:
That's because it feels a bit more natural ... `Makefile` is for building things in parallel, for declarative way of defining dependencies for targets  (while building of help.1 sounds exactly like such task).

omron93:
README files differs for every version (`9.4`, `9.5`,...). But dealing with versions is in `build.sh` so I don't see any elegant way how to add this to current `common.mk`. Suggestions?

> Makefile is for building things in parallel, for declarative way of defining dependencies for targets (while building of help.1 sounds exactly like such task).

I totally agree. But IMPOV build scripts rewrite is needed for this and is welcomed ;-)

pkubatrh:
I agree with @praiskup it makes more sense to build the `help.1` file via Makefile as a dependency of the image build. The addition of `COPY` to the Dockerfile can imo stay in `build.sh` if you want to modify the Dockerfile in one place only.

>README files differs for every version (9.4, 9.5,...). But dealing with versions is in build.sh so I don't see any elegant way how to add this to current common.mk. Suggestions?

It has been a while since I looked at the Makefiles but was there not a `VERSIONS` argument that could be used in `common.mk`?

omron93:
> It has been a while since I looked at the Makefiles but was there not a VERSIONS argument that could be used in common.mk?

Yes. There are `VERSIONS` and `VERSION` variables - the first is used when second is empty. Iterating over versions specified in these variables is done in `build.sh`.

All specified versions are now built in ONE `build.sh` invocation. So there is no parallelism. But I agree that ability to build versions in parallel would be nice feature.

praiskup:
What about this?
```Make
manpages = $(shell for version in $(VERSIONS); do echo "$$version/help.1"; done)
manpages: $(manpages)
$(manpages): %help.1: %README.md
        go-md2man -in "$^" -out "$@"
```

omron93:
Now using what @praiskup suggested.

praiskup:
Thanks.

I'm just having a look at postgresql-container, and we have help.1 inserted into `root/help.1`.  Isn't this common pattern?

pkubatrh:
>I'm just having a look at postgresql-container, and we have help.1 inserted into root/help.1. Isn't this common pattern?

After a quick look it seems like the s2i-* images do not use `root/` in their repositories at all. They also do not have their own built `help.1` file either but it would be a good idea to store the `help.1` file in the locations expected by the images that do use it. And I guess introduce `root/help.1` for those that do not? Which would also let us remove the `COPY` Dockerfile addition since that would already be present in the original Dockerfiles.

omron93:
> I'm just having a look at postgresql-container, and we have help.1 inserted into root/help.1. Isn't this common pattern?

Only following images have this file:
```
mariadb-container
devtoolset-container
ror-container
mongodb-container
mysql-container
postgresql-container
```

The idea is to remove `help.1` files from these repositories. And only generate it on build (same as OSBS do).

praiskup:
Ok, I'll merge this then.  Thank you!

praiskup:
> After a quick look it seems like the s2i-* images do not use root/ in their repositories at all. They also do not have their own built help.1 file either but it would be a good idea to store the help.1 file in the locations expected by the images that do use it. And I guess introduce root/help.1 for those that do not? Which would also let us remove the COPY Dockerfile addition since that would already be present in the original Dockerfiles

Sorry I missed your comment.  I think from long perspective, osbs should not generate help.1 at all when it is available (suggested by @pkubatrh at some meeting), and osbs should provide easily-reusable script which does the build.  But for now this looks pretty good, and perhaps somebody can write a better patch which will install help.1 under `root/`?

praiskup:
I mean, install `help.1` into `root/` when it is expected (so some configuration option would be needed, just guessing).

pkubatrh:
I think you can ignore my previous comment. I see no problems with not having the built `help.1` in git other than it not being present when the images are build just using `docker build`.

omron93:
>  I think from long perspective, osbs should not generate help.1 at all when it is available (suggested by @pkubatrh at some meeting), and osbs should provide easily-reusable script which does the build.

The goal of this PR is to make maintenance of README easier. Now in some repositories there is `help.1` and `README.md` with same content (but in different format). So if one is updated, second has to be too - this isn't comfortable for developers. This is why it is better to have only `README.md` and generate `help.1` automatically.

torsava:
I agree that placing the help.1 file under `root/` would be better—clearer and less cluttering. I'll open a PR unless someone is opposed to the idea.

omron93:
> I agree that placing the help.1 file under root/ would be better—clearer and less cluttering. I'll open a PR unless someone is opposed to the idea.

Sorry, I misunderstood pkubatrh suggestions  :-(

+1 for generating `help.1` into place which is copied to image `/` instead of modifying Dockerfile and adding `COPY` instruction.

But "common approach" for s2i based images is having only `contrib` (copied to `/opt/app-root`) and `s2i` copied to `$STI_SCRIPTS_PATH`. So nothing copied to `/`...

So it would be possible to rename `contrib` to `/root/opt/app-root`.

Or we could use `ONBUILD` docker instruction in `s2i-base` image to enforce presence of `help.1` in build content. (for example `ONBUILD COPY help.1 /`)

pkubatrh:
>Or we could use ONBUILD docker instruction in s2i-base image to enforce presence of help.1 in build content. (for example ONBUILD COPY help.1 /)

I don't like this approach as it would force every image (users` images) based upon `s2i-base` to add the `help.1` file. Also existing image builds based on `s2i-base` would fail due to this change. in my opinion we should keep this need for `help.1` contained to our images.

>But "common approach" for s2i based images is having only contrib (copied to /opt/app-root) and s2i copied to $STI_SCRIPTS_PATH. So nothing copied to /...
>So it would be possible to rename contrib to /root/opt/app-root.

This seems like the better way to do this.

PR #8, opened 2017-05-15
Move help.1 file into a root/ subdirectory

torsava:
Continuing upon the pull request #7, the help.1 file is now generated into a `root` subdirectory (which is first created should it not exist yet) and then an instruction is added to copy all contents of the `root` directory to the `/` root of the image.

s2i based images can now move the contents of their `contrib` directory into `root/opt/app-root` instead, which will then be copied automatically.

_I've also included a small bug fix of the `manpages` macro to account for the `VERSION` parameter correctly._

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -32,7 +32,6 @@ function docker_build_with_version {
   # Use perl here to make this compatible with OSX
   DOCKERFILE_PATH=$(perl -MCwd -e 'print Cwd::abs_path shift' $dockerfile)
   cp ${DOCKERFILE_PATH} "${DOCKERFILE_PATH}.version"
-  echo "COPY help.1 /" >> "${dockerfile}.version"
   git_version=$(git rev-parse --short HEAD)
   echo "LABEL io.openshift.builder-version=\"${git_version}\"" >> "${dockerfile}.version"
   if [[ "${UPDATE_BASE}" == "1" ]]; then
--- a/common.mk
+++ b/common.mk
@@ -35,7 +35,7 @@ test: manpages
 test-openshift: manpages
 	$(script_env) TAG_ON_SUCCESS=$(TAG_ON_SUCCESS) TEST_OPENSHIFT_MODE=true $(build)
 
-manpages = $(shell for version in $(VERSIONS); do echo "$$version/help.1"; done)
+manpages = $(shell for version in $(if $(VERSION), $(VERSION), $(VERSIONS)); do echo "$$version/help.1"; done)
 manpages: $(manpages)
 $(manpages): %help.1: %README.md
 	go-md2man -in "$^" -out "$@"
--- a/common.mk
+++ b/common.mk
@@ -35,7 +35,9 @@ test: manpages
 test-openshift: manpages
 	$(script_env) TAG_ON_SUCCESS=$(TAG_ON_SUCCESS) TEST_OPENSHIFT_MODE=true $(build)
 
-manpages = $(shell for version in $(if $(VERSION), $(VERSION), $(VERSIONS)); do echo "$$version/help.1"; done)
+manpages = $(shell for version in $(if $(VERSION), $(VERSION), $(VERSIONS)); \
+		do echo "$$version/root/help.1"; done)
 manpages: $(manpages)
-$(manpages): %help.1: %README.md
+$(manpages): %root/help.1: %README.md
+	mkdir -p $(@D)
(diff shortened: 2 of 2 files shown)


pkubatrh:
Copying important Dockerfile instructions into the builder script is not a good idea. If you do this then you run into the issue that when you build using `docker build` (instead of using the builder script), the `root/` directory will not be added to the image, resulting in an unusable image.

What the builder should actually do is just prepare the `help.1` file and copy it to the `root/` folder inside the git directory structure and leave it up to the specific image to do with it what it wants via their Dockerfile.

torsava:
@pkubatrh Good point! I've amended the commit accordingly.

pkubatrh:
That is better, thank you!
As for the VERSION fix, I am not sure if it is really needed. Eventually we should get rid of the VERSION variable in the Makefile, as VERSION and VERSIONS are basically used for the same purpose (= you can just pass a single version number/string to VERSIONS in order to build it).

In any way, this should only be merged after all of the images add the copy instruction to their Dockerfiles, else we run into failing CI tests.

torsava:
> That is better, thank you!
> As for the VERSION fix, I am not sure if it is really needed. Eventually we should get rid of the VERSION variable in the Makefile, as VERSION and VERSIONS are basically used for the same purpose (= you can just pass a single version number/string to VERSIONS in order to build it).

I agree that the VERSION vs. VERSIONS split is confusing and unnecessary, however, until we amend the scripts to respect only one or the other and update all the docs accordingly, I think it should be fixed.

> In any way, this should only be merged after all of the images add the copy instruction to their Dockerfiles, else we run into failing CI tests.

The help.1 patch was merged into common-scripts only 12 days ago, so I suppose not all the images already switched to it. Thus I think we only need to make sure those that did are compliant. Is there a way to find out which of them already updated their common submodule?

torsava:
Hm, so 14 images already updated, only a handful didn't. Do we open 14 issues/PRs against all of them, or how is this handled?

omron93:
> Hm, so 14 images already updated, only a handful didn't. Do we open 14 issues/PRs against all of them, or how is this handled?

Yes, separate PR for each is needed :-(

These images have to be updated to use `root` directory (more than half of all sclorg/ images)
```
devtoolset-container
nginx-container
passenger-container
s2i-base-container
s2i-nodejs-container
s2i-perl-container
s2i-php-container
s2i-python-container
s2i-ruby-container
varnish-container
```

@torsava So the plan is to rename `contrib` directory to `root/opt/app-root` and in Dockerfile use `COPY root/ /` instead of copying `contrib`, right? @bparees Are you fine with this change?

If so, I can create PRs for these images.

torsava:
> @torsava So the plan is to rename contrib directory to root/opt/app-root and in Dockerfile use COPY root/ / instead of copying contrib, right?

Indeed.

> If so, I can create PRs for these images.

I can do it, since I opened the PR. I'm helping to take care of the Python container so I need to do it there anyway.

omron93:
> I can do it since I opened the PR. I'm helping to take care of the Python container so I need to do it there anyway.

Ok, thanks. But please wait for @bparees answer if this change is not compatibility/documentation problem or something else...

bparees:
> Ok, thanks. But please wait for @bparees answer if this change is not compatibility/documentation problem or something else...

if i understand this correctly you're just talking about changing the repo structure, not the final image result.  If that's true, it's no problem.

praiskup:
LGTM.

praiskup:
I guess there's a type though -> missing backslash after newline split?

torsava:
> I guess there's a type though -> missing backslash after newline split?

Ah, thanks! It somehow got lost when I was rebasing the order of the commits. Fixed!

praiskup:
As there's no voice against it, merging. LGTM.

torsava:
> These images have to be updated to use root directory (more than half of all sclorg/ images)
> 
> devtoolset-container
> nginx-container
> passenger-container
> s2i-base-container
> s2i-nodejs-container
> s2i-perl-container
> s2i-php-container
> s2i-python-container
> s2i-ruby-container
> varnish-container

All of the containers have been successfully updated to follow the new `root/` subdirectory structure with the exception of the base container where it's part of an upcoming PR: https://github.com/sclorg/s2i-base-container/pull/121. ∎

praiskup:
Thanks for taking care of it!

PR #10, opened 2017-05-17
[WIP] Do not use VERSION and VERSIONS for the same purpose in common.mk

pkubatrh:
Instead only use VERSIONS to choose which versions of the image to build in the makefile
and depend on its rules for populating the VERSION variable to be used in build.sh

Based on https://github.com/sclorg/postgresql-container/pull/152

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -17,26 +17,24 @@ endif
 script_env = \
 	SKIP_SQUASH=$(SKIP_SQUASH)                      \
 	UPDATE_BASE=$(UPDATE_BASE)                      \
-	VERSIONS="$(VERSIONS)"                          \
 	OS=$(OS)                                        \
-	VERSION="$(VERSION)"                            \
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"
 
 .PHONY: build
-build: manpages
-	$(script_env) $(build)
+build: $(VERSIONS)
+
+.PHONY: $(VERSIONS)
+$(VERSIONS): % : %/root/help.1
+	VERSION="$@" $(script_env) $(build)
 
 .PHONY: test
-test: manpages
-	$(script_env) TAG_ON_SUCCESS=$(TAG_ON_SUCCESS) TEST_MODE=true $(build)
+test: script_env += TEST_MODE=true
+test: $(VERSIONS)
 
 .PHONY: test-openshift
-test-openshift: manpages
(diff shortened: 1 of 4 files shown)


omron93:
@bparees What backward compatibility we have to keep in build scripts?

common.mk
```diff
@@ -17,25 +17,24 @@ endif
...
-	VERSIONS="$(VERSIONS)"                          \
 	OS=$(OS)                                        \
-	VERSION="$(VERSION)"                            \
 	BASE_IMAGE_NAME=$(BASE_IMAGE_NAME)              \
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"
 
 .PHONY: build
-build: manpages
-	$(script_env) $(build)
+build: $(VERSIONS)
```
praiskup: This will cause parallel build of multiple image versions, right?  What's the status of docker bug discussed in  sclorg/postgresql-container/pull/149 ?

pkubatrh: Correct, if the user has setup to run make in parallel by default. But I agree this is still an issue. 

The bug mentioned in  sclorg/postgresql-container#149 happens when testing, not building, so to work around this issue for the time being we could leave the build rule as-is and modify the test rule to not be able to be run in parallel (possibly by separating `build.sh` into building and testing parts)

pkubatrh:
Pushed a new version of the changes, this time making sure test suites for different versions of the image are not able to be run in parallel by splitting off the testing code located in `build.sh` into `test.sh`.

That said I have once again tried running the test suites in parallel (on a rhel-7.3 machine) using code from sclorg/postgresql-container#149 (without any of the "hack" commits) and have yet to hit a failure over ~200 runs.

Eventually it would make sense to move most, if not all, of `test.sh` code into the `common.mk`.

omron93:
Not only tests of postgresql-container are not able to run in parallel. Also a lot of s2i based images runs `s2i build` in tests and name of resulted image does not differ among versions.

So they have to be fixed too before enabling parallel testing.

praiskup:
Sorry, I totally forgot about this PR;  and the PR #12 caused collisions :(.  Can this be rebased please?

pkubatrh:
Rebased. Additionally I split off the tagging code to a separate script as well. I am wondering if it makes sense to introduce a `tag` rule to the Makefile or if it's ok to leave it as is.

@praiskup @omron93 PTAL

omron93:
It looks really good. Thanks for a lot of improvements.

+1 for merging

praiskup:
Agreed, this is very nice and clean first step!  Thanks.

PR #11, opened 2017-05-31
Use docker build --label to add information about git hash

omron93:
Use `docker build --label`  to add information about repository git version. With this `Dockerfile` doesn't have to be modified.

@praiskup @pkubatrh @torsava Please take a look and merge.


praiskup:
Nice, LGTM.  Thanks!

PR #12, opened 2017-06-02
Get tag of resulted image from name and version labels

omron93:
Get tag of resulted image from `name` and `version` labels

Earch sclorg/ image should contain `name` and `version` labels (Atomic Best Practices recommends it, Fedora requires it). These labels can be used to tag the image - OSBS does the same.

So with this change there is no need for getting `NAMESPACE` from `TARGET`, `BASE_IMAGE_NAME` from `pwd`,...

@pkubatrh @torsava @praiskup Please take a look.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -16,8 +16,6 @@ VERSION=${2-$VERSION}
 
 DOCKERFILE_PATH=""
 
-BASE_DIR_NAME=$(echo $(basename `pwd`) | sed -e 's/-[0-9]*$//g')
-
 # Perform docker build but append the LABEL with GIT commit id at the end
 function docker_build_with_version {
   local dockerfile="$1"
@@ -29,9 +27,18 @@ function docker_build_with_version {
     BUILD_OPTIONS+=" --pull=true"
   fi
 
-  IMAGE_NAME="$BASE_DIR_NAME-$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 4 | head -n 1
-):${git_version}"
-  docker build ${BUILD_OPTIONS} -t ${IMAGE_NAME} -f "${dockerfile}" .
+  docker build ${BUILD_OPTIONS} -f "${dockerfile}" . | tee build.log
+  local IMAGE_ID=$(cat build.log | awk '/Successfully built/{print $NF}')
+  rm build.log
+
+  name=$(docker inspect -f "{{.Config.Labels.name}}" $IMAGE_ID)
+  version=$(docker inspect -f "{{.Config.Labels.version}}" $IMAGE_ID)
+
+  IMAGE_NAME=$name
+  if [[ -v TEST_MODE ]]; then
+    IMAGE_NAME+="-candidate"
+  fi
+  docker tag $IMAGE_ID $IMAGE_NAME
 
   if [[ "${SKIP_SQUASH}" != "1" ]]; then
     squash "${dockerfile}"
@@ -79,15 +86,10 @@ for dir in ${dirs}; do
   fi
 
   if [[ $ok_to_tag -eq 1 ]]; then
-    name=$(docker inspect -f "{{.Config.Labels.name}}" $IMAGE_NAME)
-    version=$(docker inspect -f "{{.Config.Labels.version}}" $IMAGE_NAME)
-
     echo "-> Tagging image to '$name:$version' and '$name:latest'"
     docker tag $IMAGE_NAME "$name:$version"
     docker tag $IMAGE_NAME "$name:latest"
   fi
-
-  docker rmi --no-prune $IMAGE_NAME
 
   popd > /dev/null
 done
--- a/build.sh
+++ b/build.sh
@@ -1,37 +1,38 @@
 #!/bin/bash -e
 # This script is used to build, test and squash the OpenShift Docker images.
 #
-# Name of resulting image will be: 'NAMESPACE/BASE_IMAGE_NAME-VERSION-OS'.
+# Resulting image will be tagged: 'name:version' and 'name:latest'. Name and version
+#                                  are values of labels from resulted image
 #
-# BASE_IMAGE_NAME - Usually name of the main component within container.
-# OS - Specifies distribution - "rhel7" or "centos7"
+# OS - Specifies distribution - "rhel7", "centos7" or "fedora"
 # VERSION - Specifies the image version - (must match with subdirectory in repo)
 # TEST_MODE - If set, build a candidate image and test it
 # TAG_ON_SUCCESS - If set, tested image will be re-tagged as a non-candidate
 #       image, if the tests pass.
 # VERSIONS - Must be set to a list with possible versions (subdirectories)
-# OPENSHIFT_NAMESPACES - Which of available versions (subdirectories) should be
-#       put into openshift/ namespace.
 
 OS=${1-$OS}
 VERSION=${2-$VERSION}
 
 DOCKERFILE_PATH=""
 
-test -z "$BASE_IMAGE_NAME" && {
-  BASE_DIR_NAME=$(echo $(basename `pwd`) | sed -e 's/-[0-9]*$//g')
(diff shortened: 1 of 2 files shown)


praiskup:
I'm actually not sure here.  Probably more comments about "history" of removed code (and why removing is safe) is needed?

Historically, we tagged images as `openshift/*`, then we moved to `rhscl/*` namespace.  Those tags are not needed anymore?

praiskup:
What mechanism will be used to tag the image correctly as before?

omron93:
> Historically, we tagged images as openshift/*, then we moved to rhscl/* namespace. Those tags are not needed anymore?

Yes, this is true. Some versions of RHEL variants were build with `openshift/` namespace, some with `rhscl/` one. But each image version has only one namespace.

"Removed code" was used to determine namespace depending on OS and VERSION (`Makefile` contains `NAMESPACE_OPENSHIFT` to get versions with `openshift/` namespace, rest version has `rhscl/` namespace).

> What mechanism will be used to tag the image correctly as before?

Now, almost all images contains `name` label [1] (some PRs are not merged yet). For each OS and VERSION combination there is a Dockerfile. So it can contain proper name with proper namespace.

So there is not need for getting proper namespace from OS and VERSION and `NAMESPACE_OPENSHIFT`. This is because proper image name is stored in Dockerfile the image is built from.

If new version is added and it should be built to different namespace, this has to be changed only in the newly added Dockerfile.

@praiskup Is it clear? Or anything to explain more?

To note: I am not sure, but I think all supported/built images are in `rhscl/` namespace.

[1] http://docs.projectatomic.io/container-best-practices/#_descriptive_labels

build.sh
```diff
@@ -1,37 +1,38 @@
...
 # VERSIONS - Must be set to a list with possible versions (subdirectories)
-# OPENSHIFT_NAMESPACES - Which of available versions (subdirectories) should be
-#       put into openshift/ namespace.
 
 OS=${1-$OS}
 VERSION=${2-$VERSION}
 
 DOCKERFILE_PATH=""
 
-test -z "$BASE_IMAGE_NAME" && {
```
praiskup: What was the 'test -z "$BASE_IMAGE_NAME"' check for before?  Was that an option?

build.sh
```diff
@@ -1,37 +1,38 @@
...
   local dockerfile="$1"
+  echo "-> Version ${dir}: building image from '${dockerfile}' ..."
+
   git_version=$(git rev-parse --short HEAD)
   BUILD_OPTIONS+=" --label io.openshift.builder-version=\"${git_version}\""
   if [[ "${UPDATE_BASE}" == "1" ]]; then
     BUILD_OPTIONS+=" --pull=true"
   fi
+
+  IMAGE_NAME="$BASE_DIR_NAME-$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 4 | head -n 1
```
praiskup: What is the random string for?  Can you document that in code?

build.sh
omron93: Not sure.

If user set BASE_IMAGE_NAME in environment and run make XY, could this value be propagated by make to `build.sh`?

praiskup:
It basically looks good to me, can anyone else have a look too and ACK please? @pkubatrh @hhorak 

I have two warnings:
1. The 'BASE_IMAGE_NAME' is now not replaceable from environment.
2. Previously we had predictable output.  We either directly tagged the image into $IMAGE_NAME, or into ${IMAGE_NAME}-candidate (with make test) and we never removed anything.  Now we first tag into semi-unique tag name, and then we conditionally tag into $IMAGE_NAME (which makes no predictable image name for `make test` unless TAG_ON_SUCCESS is done), while the semi-unique image is always removed.

To me, the `build.sh` was never logically implemented, and from this POV this pull request is neutral for me (I don't use anything from what is changed).  So unless we know the changed stuff is used by someone, I'm fine, thanks!

Long term POV: We should separate testing and building in Makefile, but that's a different thing.

build.sh
omron93: User now can't customize resulted image name by `BASE_IMAGE_NAME` (to note: as I know, it was not documented anywhere) - name of the image is get from `name` label.

But user can customize the name by adding `BUILD_OPTIONS="--label name=my_image"` to `make build/test`

omron93:
> 2. Previously we had predictable output. We either directly tagged the image into $IMAGE_NAME, or into ${IMAGE_NAME}-candidate (with make test) and we never removed anything. Now we first tag into semi-unique tag name, and then we conditionally tag into $IMAGE_NAME (which makes no predictable image name for make test unless TAG_ON_SUCCESS is done), while the semi-unique image is always removed.

This change is reverted now - to not block this PR and to allow broader discussion.


For warning 1:
- name is get from `name` label, so it does not make sense what `BASE_IMAGE_NAME` could mean... Name of resulted image can be configured by `BUILD_OPTIONS="--label name=my_image"`

build.sh
omron93: It was used to allow to run `make test` in two separate copies of image directory in parallel.

build.sh
```diff
@@ -1,37 +1,45 @@
...
   git_version=$(git rev-parse --short HEAD)
   BUILD_OPTIONS+=" --label io.openshift.builder-version=\"${git_version}\""
   if [[ "${UPDATE_BASE}" == "1" ]]; then
     BUILD_OPTIONS+=" --pull=true"
   fi
-  docker build ${BUILD_OPTIONS} -t ${IMAGE_NAME} -f "${dockerfile}" .
+
+  docker build ${BUILD_OPTIONS} -f "${dockerfile}" . | tee build.log
+  local IMAGE_ID=$(cat build.log | awk '/Successfully built/{print $NF}')
+  rm build.log
```
praiskup: Could we use:
```bash
local docker_cmd=(the command)
{ IMAGE_ID=$("${docker_cmd[@]}" | tee /dev/fd/$fd | awk '/Successfully built/{print $NF}'); } {fd}>&1
```

Or something similar, to avod creating temporary "build.log" file?

build.sh
praiskup: This makes sense to me.  Please mention the removal in commit message, and also add 'Fixes #12' line at the end.

build.sh
```diff
@@ -88,12 +69,11 @@ for dir in ${dirs}; do
...
   fi
 
+  ok_to_tag=1
   if [[ -v TEST_MODE ]]; then
     VERSION=$dir IMAGE_NAME=${IMAGE_NAME} test/run
-
-    if [[ $? -eq 0 ]] && [[ "${TAG_ON_SUCCESS}" == "true" ]]; then
-      echo "-> Re-tagging ${IMAGE_NAME} image to ${IMAGE_NAME%"-candidate"}"
-      docker tag $IMAGE_NAME ${IMAGE_NAME%"-candidate"}
+    if [[ $? -ne 0 ]] || [[ "${TAG_ON_SUCCESS}" != "true" ]]; then
```
praiskup: I think '[[ $? -eq 0 ]]' used to be no-op before, because if 'test/run' failed, the 'set -e' would already exit the shell anyways right before the condition....   And now the reversed version '[[ $? -ne 0 ]]' might never succeed too.

build.sh
omron93: Yes. Thanks for suggestion.

build.sh
omron93: @praiskup But `build.sh` does not set `set -e`, this is only set in `test/run`.

praiskup: There's `#!/bin/bash -e` shebang.

omron93:
Rebased to change commit message of first commit - to include https://github.com/sclorg/container-common-scripts/pull/12#discussion_r120374521

build.sh
omron93: You are right, I didn't see this.

Checking of return code is now removed.

praiskup:
LGTM.  Thanks.  As nobody complained so far, I'm taking the liberty to merge this (and if this causes some real troubles, we can partly revert or fix later).

PR #13, opened 2017-06-05
Fix squashing - docker-squash does not tag image by default

omron93:
Fix squashing - docker-squash does not tag image by default (it proclaims it - `By default it'll be set to 'image' argument`)

In current situation squashing is done, but the resulted image is not tagged. So testing is done with unsquashed image.

Also updated docker-squash

https://github.com/sclorg/s2i-base-container/pull/110 aims also to update to docker squash. It also brings a  change that image is build with `-unsquashed` suffix and after squashing this image is removed.
I'm against this change, because it "cleans" docker build cache and slows down builds. But it option too. Opinions?

@pkubatrh @praiskup @torsava Please take a look and merge.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -42,9 +42,9 @@ function docker_build_with_version {
 function squash {
   # FIXME: We have to use the exact versions here to avoid Docker client
   #        compatibility issues
-  easy_install -q --user docker_py==1.7.2 docker-squash==1.0.1
+  easy_install -q --user docker_py==1.10.6 docker-squash==1.0.5
   base=$(awk '/^FROM/{print $2}' $1)
-  ${HOME}/.local/bin/docker-squash -f $base ${IMAGE_NAME}
+  ${HOME}/.local/bin/docker-squash -f $base ${IMAGE_NAME} -t ${IMAGE_NAME}
 }
 
 # Versions are stored in subdirectories. You can specify VERSION variable


build.sh
```diff
@@ -42,9 +42,9 @@ function docker_build_with_version {
 function squash {
   # FIXME: We have to use the exact versions here to avoid Docker client
   #        compatibility issues
-  easy_install -q --user docker_py==1.7.2 docker-squash==1.0.1
+  easy_install -q --user docker_py==1.10.6 docker-squash==1.0.5
```
praiskup: Hms, I missed this ugly heritage :) .. could we rather remove the `easy_install` entirely and depend on system package?

praiskup: is the comment about "compatibility issues" still valid?

omron93: > Hms, I missed this ugly heritage :) .. could we rather remove the easy_install entirely and depend on system package?

I think that this package is not in RHEL7.

> is the comment about "compatibility issues" still valid?

I think not. Does someone know why it was added?

praiskup:
> Fix squashing - docker-squash does not tag image by default (it proclaims it - By default it'll be set to 'image' argument)

It looks OK to me, though I'm missing reference to upstream bugreport regarding this ... if docker-squash proclaims tagging, it seems to be a clear bug.

Also, it looks like a good material for @pkubat for the review, there were some good ideas about using `-c` option or so in sclorg/s2i-base-container#110

hhorak:
I agree with `-c` usage, it should help us keep disk on testing machine empty enough for longer period.

hhorak:
As for the version locking -- upstream of atomic-reactor uses `docker-squash>=1.0.0rc3` https://github.com/projectatomic/atomic-reactor/blob/master/requirements.txt#L2 but using latest greatest might not be entirely safe, so some better idea might give version chosen by Fedora's OSBS: https://koji.fedoraproject.org/koji/packageinfo?packageID=22304

Or, as proposed, we might not necessarily install it in case there is already the docker-squash installed from the system (detect this). We can then include a stable version on the system that matches preference of OSBS maintainers.

pkubatrh:
>I agree with -c usage, it should help us keep disk on testing machine empty enough for longer period.

I think there was a mention of this also cleaning the docker cache (have not tested it out yet), effectively making any subsequent testing builds slower.

praiskup:
Reported as: goldmann/docker-squash/issues/151

praiskup:
Thanks, merged.  We should consider later (different issue):
- not doing squash for centos images && fedora by default
- using `-c` option
- possibly taking the docker-squash out completely ... as that's used only by RHEL CI

PR #15, opened 2017-07-21
Add a common test file shared among all containers

trepik:
This file contains common bits for container tests, and should be sourced by each specific container test. It is created to keep the code on one place and not spread through the repositories.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -7,8 +7,8 @@
 # CID_FILE_DIR=$(mktemp --suffix=<container>_test_cidfiles -d)
 # reguires definition of TEST_LIST 
 # TEST_LIST="\
-# run_container_creation_tests
-# run_doc_test <words_to_look_for_in_the_doc>"
+# ctest_container_creation
+# ctest_doc_content"
 
 # Container CI tests
 # abbreviated as "ct"
@@ -79,32 +79,48 @@ function ct_wait_for_cid() {
   local attempt=1
   local result=1
   while [ $attempt -le $max_attempts ]; do
-    [ -f $cid_file ] && [ -s $cid_file ] && break
+    [ -f $cid_file ] && [ -s $cid_file ] && return 0
     : "Waiting for container start..."
     attempt=$(( $attempt + 1 ))
     sleep $sleep_time
   done
+  return 1
 }
 
 # ct_assert_container_creation_fails [container_args]
 # --------------------
 # The invocation of docker run should fail based on invalid container_args
 # passed to the function. Returns 0 when container fails to start properly.
-# Argument: container_args - all argument are passed directly to dokcer run
+# Argument: container_args - all arguments are passed directly to dokcer run
+# Uses: $CID_FILE_DIR - path to directory containing cid_files
 function ct_assert_container_creation_fails() {
-
-  # Time the docker run command. It should fail. If it doesn't fail,
-  # container will keep running so we kill it with SIGTERM to make sure
-  # timeout returns a non-zero value.
+  local ret=0
+  local max_attempts=10
+  local attempt=1
+  local cid_file=assert
   set +e
-  timeout -s SIGTERM --preserve-status 30s docker run --rm "$@" $IMAGE_NAME
-  ret=$?
-  set -e
+  ct_create_container $cid_file "$@"
+  if [ $? -eq 0 ]; then
+    local cid=$(ct_get_cid $cid_file)
 
-  # Timeout will exit with a high number.
-  if [ $ret -gt 128 ]; then
-    return 1
+    while [ "$(docker inspect -f '{{.State.Running}}' $cid)" == "true" ] ; do
+      sleep 2
+      attempt=$(( $attempt + 1 ))
+      if [ $attempt -gt $max_attempts ]; then
+        docker stop $cid
+        ret=1
+        break
+      fi
+    done
+    exit_status=$(docker inspect -f '{{.State.ExitCode}}' $cid)
+    if [ "$exit_status" == "0" ]; then
+      ret=1
+    fi
+    docker rm $cid
+    rm $CID_FILE_DIR/$cid_file
   fi
+  set -e
+  return $ret
 }
 
 # ct_create_container [name, additional_args]
@@ -121,7 +137,7 @@ function ct_create_container() {
   local cid_file="$CID_FILE_DIR/$1" ; shift
   # create container with a cidfile in a directory for cleanup
   docker run ${CONTAINER_ARGS:-} --cidfile="$cid_file" -d "$@" $IMAGE_NAME
-  ct_wait_for_cid $cid_file
+  ct_wait_for_cid $cid_file || return 1
   : "Created container $(cat $cid_file)"
 }
 
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -34,7 +34,7 @@ function ct_cleanup() {
       : "Dumping logs for $container"
       docker logs $container
     fi
-    docker rm $container
+    docker rm -v $container
     rm $cid_file
   done
   rmdir $CID_FILE_DIR
@@ -51,7 +51,7 @@ function ct_enable_cleanup() {
 # ct_get_cid [name]
 # --------------------
 # Prints container id from cid_file based on the name of the file.
-# Argument: name - name of cid_file containing the container id
+# Argument: name - name of cid_file where the container id will be stored
 # Uses: $CID_FILE_DIR - path to directory containing cid_files
 function ct_get_cid() {
   local name="$1" ; shift || return 1
@@ -99,7 +99,9 @@ function ct_assert_container_creation_fails() {
   local attempt=1
   local cid_file=assert
   set +e
-  ct_create_container $cid_file "$@"
+  local old_container_args="${CONTAINER_ARGS-}"
+  CONTAINER_ARGS="$@"
+  ct_create_container $cid_file
   if [ $? -eq 0 ]; then
     local cid=$(ct_get_cid $cid_file)
 
@@ -116,27 +118,28 @@ function ct_assert_container_creation_fails() {
     if [ "$exit_status" == "0" ]; then
       ret=1
     fi
-    docker rm $cid
+    docker rm -v $cid
     rm $CID_FILE_DIR/$cid_file
   fi
+  [ ! -z $old_container_args ] && CONTAINER_ARGS="$old_container_args"
   set -e
   return $ret
 }
 
-# ct_create_container [name, additional_args]
+# ct_create_container [name, command]
 # --------------------
 # Creates a container using the IMAGE_NAME and CONTAINER_ARGS variables. Also
 # stores the container id to a cid_file located in the CID_FILE_DIR, and waits
 # for the creation of the file.
-# Argument: name - name of cid_file containing the container id
-# Argument: additional_args - optional arguments passed directly to docker
+# Argument: name - name of cid_file where the container id will be stored
+# Argument: command - optional command to be executed in the container
 # Uses: $CID_FILE_DIR - path to directory containing cid_files
 # Uses: $CONTAINER_ARGS - optional arguments passed directly to docker run
 # Uses: $IMAGE_NAME - name of the image being tested
 function ct_create_container() {
   local cid_file="$CID_FILE_DIR/$1" ; shift
   # create container with a cidfile in a directory for cleanup
-  docker run ${CONTAINER_ARGS:-} --cidfile="$cid_file" -d "$@" $IMAGE_NAME
+  docker run --cidfile="$cid_file" -d ${CONTAINER_ARGS:-} $IMAGE_NAME "$@"
   ct_wait_for_cid $cid_file || return 1
   : "Created container $(cat $cid_file)"
 }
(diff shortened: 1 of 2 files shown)


tests.sh
```diff
@@ -0,0 +1,113 @@
...
+      : "Dumping logs for $CONTAINER"
+      docker logs $CONTAINER
+    fi
+    docker rm $CONTAINER
+    rm $cid_file
+  done
+  rmdir $CID_FILE_DIR
+  : "Done."
+}
+trap cleanup EXIT SIGINT
```
praiskup: As this is now library, we should explicitly ask for the cleanup action;  so I would suggest you to have some `enable_auto_cleanup` method.

praiskup:
Thanks for working on it!

- my first thought is as-clean-API-as-possible... please design naming conventions for the api calls and use that;  e.g. `cictests_*` prefix (ci-container-tests) for methods

- document how to use it at the beginning of the file

- (optional) provide one pull request against some container that would depend on this feature, that would simplify testing of the proposal :)

trepik:
- added prefix to all function names
- added extra function enabling the cleanup
- documentation will follow once it's working with more containers
- one can check the functionality with Mongo 3.2 container from my [mongo-container](https://github.com/trepik/mongodb-container) fork

test-lib.sh
```diff
@@ -0,0 +1,142 @@
...
+#
+# reguires definition of CID_FILE_DIR
+# CID_FILE_DIR=$(mktemp --suffix=<container>_test_cidfiles -d)
+# reguires definition of TEST_LIST 
+# TEST_LIST="\
+# run_container_creation_tests
+# run_doc_test <words_to_look_for_in_the_doc>"
+
+# Container CI tests
+# abbreviated as "concit"
```
praiskup: Could be OK, but @pkubatrh, @omron93, @hhorak FYI.  Ideas?

I have no strong preference, but maybe `ct_` prefix could be enough  (Container Testing).

Once we claim we provide "library", it will be much harder to migrate to different function prefix.  So we should rather agree that we use pretty namespace.

test-lib.sh
```diff
@@ -0,0 +1,142 @@
...
+
+# may be redefined in the specific container testfile
+EXPECTED_EXIT_CODE=0
+
+function concit_cleanup() {
+  for cid_file in $CID_FILE_DIR/* ; do
+    CONTAINER=$(cat $cid_file)
```
praiskup: The CONTAINER variable is not `local`, which creates shell variable pollution.

I would encourage you to document shell functions like:
```bash
# foo_action [ARGS...]
# --------------------
# Does "something" with all ARGS. Set's $X to $Z value.
# Defines: $X
# Uses: $Y
foo_actions ()
{
    X=$Y
    for arg ; do
        something "$arg"
    done
    local Z=baz # Z is not visible outside...
}
```

test-lib.sh
pkubatrh: +1 for `ct_`. The shorter the better.

test-lib.sh
```diff
@@ -0,0 +1,142 @@
...
+}
+
+# Make sure the invocation of docker run fails.
+function concit_assert_container_creation_fails() {
+
+  # Time the docker run command. It should fail. If it doesn't fail,
+  # container will keep running so we kill it with SIGKILL to make sure
+  # timeout returns a non-zero value.
+  set +e
+  timeout -s SIGTERM --preserve-status 10s docker run --rm "$@" $IMAGE_NAME
```
praiskup: I'm bit confused reading this call....   It returns '1' if the command timeouts, otherwise it returns 0.

So the name should be '*assert_container_start_timeouts' maybe?

test-lib.sh
```diff
@@ -0,0 +1,142 @@
...
+  fi
+}
+
+# to pass some arguments you need to specify CONTAINER_ARGS variable
+function concit_create_container() {
+  cid_file="$CID_FILE_DIR/$1" ; shift
+  # create container with a cidfile in a directory for cleanup
+  docker run ${CONTAINER_ARGS:-} --cidfile="$cid_file" -d "$@" $IMAGE_NAME
+  : "Created container $(cat $cid_file)"
+  concit_wait_for_cid
```
praiskup: Could `wait_for_cid` method have an argument, rather than global `$cid_file` variable?

test-lib.sh
```diff
@@ -0,0 +1,142 @@
...
+}
+
+function concit_scl_usage() {
```
praiskup: API of this function is really unclear..  Seems like it is `grep_container_output`, but I'm unsure TBH

test-lib.sh
```diff
@@ -0,0 +1,142 @@
...
+    return 1
+  fi
+  out=$(docker exec $(get_cid $name) /bin/sh -ic "${run_cmd}" 2>&1)
+  if ! echo "${out}" | grep -q "${expected}"; then
+    echo "ERROR[exec /bin/sh -ic "${run_cmd}"] Expected '${expected}', got '${out}'"
+    return 1
+  fi
+}
+
+function concit_run_doc_test() {
```
praiskup: This is actually a test, not library-call.  It should be obvious from the function name ... Proposals?

test-lib.sh
```diff
@@ -0,0 +1,142 @@
...
+  done
+  # Check whether the files use the correct format
+  if ! file ${tmpdir}/help.1 | grep -q roff ; then
+    echo "ERROR: /help.1 is not in troff or groff format"
+    return 1
+  fi
+  : "  Success!"
+}
+
+function concit_run_all_tests() {
```
praiskup: Could we leave this unsolved by container library, and do it as separate PR with something more "revolutionary"?

My thought would be to have separate file like `run-tests` for this method...  e.g.  And perhaps split particular tests into separate files too in future (so we can specify list of tests to be performed via arguments, or make variables).

praiskup:
Thanks a lot for the work.  My thought is to not necessarily move everything now, but start with smaller set of "clearly useful and well defined API functions".  Even if we move only five methods, it is good first step.

praiskup:
Sorry for closing, that was inadvertently ...

test-lib.sh
hhorak: I don't think the timeout is important here, it is just an implementation.. The result should be simply 0 if the container fails, 1 if it does not.

test-lib.sh
hhorak: `grep_container_output` makes sense to me.

test-lib.sh
hhorak: `ct_` prefix for library functions and `ctest_` or `run_ctest_` prefix for tests? Or just simply `run_test_` prefix for tests?

test-lib.sh
hhorak: I'd also increase the timeout to 30s since if the container is expected to fail after the initialization of the database, we can timeout too soon.. as mentioned above, if the container works (fails) as expected, it will fail much sooner in most cases..

test-lib.sh
hhorak: Actually, I've realized that this function really tests whether `scl enable` works... that's why there are three calls done: `docker run`, `docker exec` with `bash` and `docker exec` with `sh`... But something like `grep_container_output` may be handy as well.

test-lib.sh
praiskup: `ct_` for lib and `ctest_` for tests sounds OK to me.

trepik:
- changed the prefix from concit_ to ct_
- all the functions are now documented
- timeout in the ct_assert_container_creation_fails increased to 30s
- ct_wait_for_cid has an argument instead of using a global variable
- ctest_ prefix is used for single tests
- scl_usage name stayed as before
- run_doc_test renamed to ctest_doc_content
- run_all_tests renamed to run_test_list
I suggest labeling some of the functions as "use at your own risk" or "deprecated" but keep them until a new better ones are implemented

test-lib.sh
trepik: If you wanted a function `grep_container_command_output` you'd probably run the docker only once and not 3 times so therefore keeping the name. Once there is a better implementation of testing scl_usage it could be replaced.

test-lib.sh
```diff
@@ -0,0 +1,199 @@
...
+function ct_assert_container_creation_fails() {
+
+  # Time the docker run command. It should fail. If it doesn't fail,
+  # container will keep running so we kill it with SIGTERM to make sure
+  # timeout returns a non-zero value.
+  set +e
+  timeout -s SIGTERM --preserve-status 30s docker run --rm "$@" $IMAGE_NAME
+  ret=$?
+  set -e
+
```
praiskup: I think that if `docker run` fails with exit status `1`, this function is going to succeed.. which seems to be a bug (confirmed by @hhorak).

test-lib.sh
```diff
@@ -0,0 +1,199 @@
...
+# Uses: $IMAGE_NAME - name of the image being tested
+function ct_create_container() {
+  local cid_file="$CID_FILE_DIR/$1" ; shift
+  # create container with a cidfile in a directory for cleanup
+  docker run ${CONTAINER_ARGS:-} --cidfile="$cid_file" -d "$@" $IMAGE_NAME
+  ct_wait_for_cid $cid_file
+  : "Created container $(cat $cid_file)"
+}
+
+# ctest_scl_usage [name, command, expected]
```
praiskup: If this is the final test-case (`ctest_` prefix), avoiding arguments might be a better idea.

praiskup: That's because you can not simply list the function in $TEST_LIST, but you have to (somehow) specify the arguments, too.

test-lib.sh
```diff
@@ -0,0 +1,199 @@
...
+  out=$(docker exec $(get_cid $name) /bin/sh -ic "${command}" 2>&1)
+  if ! echo "${out}" | grep -q "${expected}"; then
+    echo "ERROR[exec /bin/sh -ic "${command}"] Expected '${expected}', got '${out}'"
+    return 1
+  fi
+}
+
+# ctest_doc_content [strings]
+# --------------------
+# Looks for occurence of stirngs in the documentation files and checks
```
praiskup: Similarly here, if this is "end-test" -- I'm not sure it should be parametrized;  that is -> `ctest_doc_content` function should be run against our all images, and guarantee that our images contain the same set of documentation files.

praiskup:
Looks much better to me, thanks!

On Thursday, August 17, 2017 4:00:44 PM CEST Tomas wrote:
> - ctest_ prefix is used for single tests

The tests should be IMO written so they don't accept arguments.  If they need
arguments, perhaps that's rather library call.

> - run_all_tests renamed to run_test_list

I would use the `ct_run_all_tests`.

> I suggest labeling some of the functions as "use at your own risk" or
> "deprecated" but keep them until a new better ones are implemented

Agreed, if you care enough please do.  E.g. `run_all_tests` was just an
work-around at the time when we were not allowed to do refactoring of testsuite.
I'll take a look at better mechanism once we have this PR merged.

praiskup:
Another note:  Consider that some of the methods are also useful for writing `build.sh` or `tag.sh` files, or some per-container build script in remote git repository;  IOW consider that some method is not strictly related to testing..   I'm curious whether we shouldn't rather go out with `cbl_` prefix (**general** container *build* library), while keeping the `ctest_*` prefix for the testcases.

test-lib.sh
trepik: Well I think the function should fail only if the docker run command is terminated by the timeout command... meaning that with the container_args provided the container starts properly

test-lib.sh
trepik: should I change the prefix to ct_ then? I don't get it

praiskup: Well, final test functions (those which can be listed in TEST_LIST) should be named `ctest_`, IMO.

test-lib.sh
hhorak: I'm puzzled now :) First of all, I though I already commented here, but now I don't see any comment from myself, so I might not send it.. so again..
Anyway, @praiskup, I might have been wrong when we talked about it.. IMO the function was meant to test whether the container does not start with invalid parameters. So, the behaviour should be this:
* if the container is still running after 30s --> function should fail
* if the container ends with 0 --> function should fail
* if the container ends with 1 --> function should succeed

What we might also consider (but I'm not sure whether it would bring some benefits), is running `docker inspect` and see `State.Running` and `State.ExitCode`. But maybe we're fine with the timeout above.

When I tried it myself a bit now, the only thing that did not work for me was that sometimes `SIGTERM` might not be enough.. I had to use `-s SIGKILL` to kill some daemons..

trepik:
- re-implemented `ct_assert_container_creation_fails` function without using `timeout`, inspecting rather the Status and ExitCode like @hhorak suggested 
- local variable in `cleanup` function decapitalized
- `ctest_` prefixed functions renamed

praiskup:
[test]

test-lib.sh
```diff
@@ -0,0 +1,215 @@
...
+    local container=$(cat $cid_file)
+
+    : "Stopping and removing container $container..."
+    docker stop $container
+    exit_status=$(docker inspect -f '{{.State.ExitCode}}' $container)
+    if [ "$exit_status" != "$EXPECTED_EXIT_CODE" ]; then
+      : "Dumping logs for $container"
+      docker logs $container
+    fi
+    docker rm $container
```
omron93: @trepik @praiskup What about removing also volumes? (`docker rm -v`)

Running tests several times creates a lot of unusefull stuff. For example older versions of mongodb preallocate ~300MB with each start.

praiskup: +1, but I would probably postpone the new features to new PRs.

praiskup: Ah, if that's only about `-v` option, it might work well...

test-lib.sh
```diff
@@ -0,0 +1,215 @@
...
+# for the creation of the file.
+# Argument: name - name of cid_file containing the container id
+# Argument: additional_args - optional arguments passed directly to docker
+# Uses: $CID_FILE_DIR - path to directory containing cid_files
+# Uses: $CONTAINER_ARGS - optional arguments passed directly to docker run
+# Uses: $IMAGE_NAME - name of the image being tested
+function ct_create_container() {
+  local cid_file="$CID_FILE_DIR/$1" ; shift
+  # create container with a cidfile in a directory for cleanup
+  docker run ${CONTAINER_ARGS:-} --cidfile="$cid_file" -d "$@" $IMAGE_NAME
```
omron93: Shouldn't the `$@` be placed after `$IMAGE_NAME`? Now `$@` and `${CONTAINER_ARGS:-}` do the same and there is no way to specify command for new container.

omron93:
Thanks for the work done in this PR.

Rebase is needed to make tests working. On the other hand, without PR for some container which add using of these functions, testing of this PR makes no sense.

test-lib.sh
trepik: good idea, I'll add the -v option to the `ct_cleanup` function as well

trepik:
I'm working on PR for mongo that would use these functions.

test-lib.sh
trepik: I didn't realize what the $@ part was meant to be used to. I'l fix that as you suggest. I'll also move `${CONTAINER_ARGS:-}` right infront of the `$IMAGE_NAME`

praiskup:
@trepik can you reference the PR against mongodb in commit-message, like:

    Required-by: mongodb-container#<PR_ID>

See #23 PR for more info.  Btw., seems like we should document "how to contribute".

praiskup:
[test]

praiskup:
Ah, this still needs a rebase.  But since Tomas is not going to continue on this PR (most probably(, we should submit a new PR starting from what has been done here.

praiskup:
Closing this, and continuing in #27.  Thanks!

PR #16, opened 2017-07-21
Add option to cleanup built images and containers created by tests

pkubatrh:
I have added the clean-up code to `tag.sh` since that seemed like the best place to me (either tag the results or remove them).

Currently the clean-up code removes:
- All leftover containers that have been created from the built image
- The built image (squashed image when not using SKIP_SQUASH)
- All untagged images (unsquashed image when not using SKIP_SQUASH)
- All volumes that are no longer referenced by any container

I know @omron93 mentioned somewhere that removing the un-squashed image might make the PR CI slower since the build cache is removed by doing this. Do we want to keep the un-squashed images for the sake of speed-up?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -37,6 +37,7 @@ function docker_build_with_version {
   docker tag $IMAGE_ID $IMAGE_NAME
 
   if [[ "${SKIP_SQUASH}" != "1" ]]; then
+    docker tag $IMAGE_ID "${IMAGE_NAME}-unsquashed"
     squash "${dockerfile}"
   fi
   # Narrow by repo:tag first and then grep out the exact match
--- a/build.sh
+++ b/build.sh
@@ -39,7 +39,8 @@ function docker_build_with_version {
   if [[ "${SKIP_SQUASH}" != "1" ]]; then
     squash "${dockerfile}"
   fi
-  docker images $IMAGE_NAME -q >.image-id
+  # Narrow by repo:tag first and then grep out the exact match
+  docker images "${IMAGE_NAME}:latest" --format="{{.Repository}} {{.ID}}" | grep "^${IMAGE_NAME}" | awk '{print $2}' >.image-id
 }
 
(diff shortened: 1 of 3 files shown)


torsava:
> I know @omron93 mentioned somewhere that removing the un-squashed image might make the PR CI slower since the build cache is removed by doing this. Do we want to keep the un-squashed images for the sake of speed-up?

How about making it an optional argument?

praiskup:
@torsava wrote:
> How about making it an optional argument?

It sort of is optional argument, implemented as environment variable .. I would vote +1 for merge, @torsava can you elaborate?

pkubatrh:
@torsava most likely meant to suggest to make the clean-up of unsquashed (= untagged) into a separate argument. So we could have CLEAN_AFTER and additionally CLEAN_UNTAGGED. Or something along those lines.

Or we can just check if CLEAN_AFTER is set (to any value) to clean everything except the untagged images and when it is set to a specific value (e.g. CLEAN_AFTER=all) then it would also clean untagged images.

torsava:
@praiskup Exactly what @pkubatrh said. I think both are useful use cases and thus giving the users the option instead of forcing one or the other would be better.

> Or we can just check if CLEAN_AFTER is set (to any value) to clean everything except the untagged images and when it is set to a specific value (e.g. CLEAN_AFTER=all) then it would also clean untagged images.

Good idea!

pkubatrh:
After discussing the changes with @praiskup offline I have opted to remove the clean-up of all unused volumes since it seems dangerous to do and could possibly make our CI fail in case multiple PRs were tested at once.

I have also modified the code to not remove all untagged images (for similar reasons as above) and removed the unsquashed image instead (if forced by setting the `CLEAN_AFTER` to `all`).

praiskup:
LGTM

pkubatrh:
Merging, thanks.

PR #17, opened 2017-07-24
Small compatability fix

jbornemann:
I was having bash compatibility issues here. Small change

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -30,7 +30,7 @@ function docker_build_with_version {
   name=$(docker inspect -f "{{.Config.Labels.name}}" $IMAGE_ID)
 
   IMAGE_NAME=$name
-  if [[ -v TEST_MODE ]]; then
+  if [ -n "${TEST_MODE}" ]; then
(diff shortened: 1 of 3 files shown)


pkubatrh:
Thanks for the PR!

Can you also add the same compatibility changes to `test.sh` and `tag.sh`?

jbornemann:
@pkubatrh sure. Done.

praiskup:
@jbornemann I'm just curious what shell are you using so the `[[ -v VAR ]]` construct is problematic...  Could you sum-up the original issue in commit message please?

praiskup:
Seems like `test` in bash 4.1.2 (RHEL6) doesn't have `-v` option yet.  So are you working on el6 containers? :)  I'm just curious.

jbornemann:
Yeah, [[ -v ]] is available after bash 4.2. I'm actually using a recent version of Darwin, which includes ancient bash 3.2.

praiskup:
Thanks!

PR #19, opened 2017-07-27
More logical buildsystem

praiskup:
Use more standardized "make" workflow
    
From now we don't blindly squash image if there are no changes
(aka docker build was no-op).  We also remove old (unreferenced
images if those are to be replaced by new images.
    
    `make` or `make build`
        Builds the images.
    
    `make tag`
        Tag built images (depends on `make build`).
    
    `make test`
        Tests the :latest image (build if necessary).  This depends on
        `make tag` because the existing test-cases out in the wild
        depend on tagged images.  The most importantly the s2i tool
        doesn't seem to work with $IMAGE_ID only.

The `make tag` also tags the image with `:squashed` or `:raw` suffixes,
while the `:latest` image points (only) to one of them.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,8 +3,6 @@
 #
 # OS - Specifies distribution - "rhel7", "centos7" or "fedora"
 # VERSION - Specifies the image version - (must match with subdirectory in repo)
-# TAG_ON_SUCCESS - If set, tested image will be re-tagged as a non-candidate
-#       image, if the tests pass.
 # VERSIONS - Must be set to a list with possible versions (subdirectories)
 
 OS=${1-$OS}
--- a/build.sh
+++ b/build.sh
@@ -3,7 +3,6 @@
 #
 # OS - Specifies distribution - "rhel7", "centos7" or "fedora"
 # VERSION - Specifies the image version - (must match with subdirectory in repo)
-# TEST_MODE - If set, build a candidate image and test it
 # TAG_ON_SUCCESS - If set, tested image will be re-tagged as a non-candidate
 #       image, if the tests pass.
 # VERSIONS - Must be set to a list with possible versions (subdirectories)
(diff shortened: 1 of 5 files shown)


pkubatrh:
All in all I really like the changes from the point of how everything is handled by make dependencies and I was considering to try writing something similar as this myself but unfortunately I am not a great make-magician as of yet.
+1 for `build-serial`! That is a really nice way to make sure we are not running in parallel by default and seems worthwhile to merge even by itself.

As for the qualms I have about this:
I think skipping the build stage when a build is already done can be omitted. It definitely is proper make work-flow but does not make much sense in the docker world since a build of an existing image is nearly instantaneous due to the docker cache. At least when the images are not squashed, which users will not do anyway if they are concerned for how much of time is spent on the build process.
Removing the build stage skip would make `common.mk` a bit easier to read as there would not be a reason to generate dependencies on each of the files inside the images' respective directories.
The issue then would be how to modify the makefile while making sure the work-flow (which is one of the best points about this change imo) stays the same, or at least very similar...

pkubatrh:
One additional thing - using image's ID as the content of the IMAGE_NAME variable used inside tests might break said tests for some images that rely on it being a name string that can be parsed properly (I have implemented something like this in thermostat16* tests - https://github.com/sclorg/thermostat-container/commit/a2c6dc3611823250508bdfe32912e59614548921#diff-8c6c67279d694dec8dcd75f8e276c4d9R34)

omron93:
> One additional thing - using image's ID as the content of the IMAGE_NAME variable used inside tests might break said tests for some images that rely on it being a name string that can be parsed properly (I have implemented something like this in thermostat16* tests - sclorg/thermostat-container@a2c6dc3#diff-8c6c67279d694dec8dcd75f8e276c4d9R34)

Also it is not possible for s2i based images. `s2i build` works only with tags and it is not possible to use image ID !

build.sh
```diff
@@ -68,33 +65,64 @@ function docker_build_with_version {
...
+# Sets:
+#   $IMAGE_ID
+squash ()
+{
+  local base squashed_from squashed= unsquashed=$IMAGE_ID
+  test "$SKIP_SQUASH" = 1 && return 0
+
   # FIXME: We have to use the exact versions here to avoid Docker client
   #        compatibility issues
   local squash_version=1.0.5
```
omron93: @praiskup I vote for not checking for exact `docker_squash` version.

Imagine a situation when we decides to change this version. We/user will have to update `docker_squash` on testing machine to make similar PR tests to pass.
But since other images won't use newest version of container-common-scripts its PRs will fail.

So common-scripts in images will have to updated all in once and all other PRs will fail till that time.

omron93:
> when doing make test the images are automatically tagged (TBH I don't claim that it is ideal approach, but at least for s2i tests we need IMAGE_NAME tag so that's sort of compromise)

@praiskup Or if you would like to avoid this, some "random" image name can be generated for tests only and removed after.

Overall, it looks very good. Thanks.

build.sh
praiskup: @omron93 wrote:
> I vote for not checking for exact docker_squash version.

+1.  I was not brave enough to drop that ... since this hack was important enough to be fixed many, many times ...  So the question is whether we really care about the version or not.  Maybe there should be `version >= 1.0.5`, but IMO this deserves separate PR and discussion.

> Imagine a situation when we decides to change this version. We/user will have to update docker_squash on testing machine to make similar PR tests to pass.

Right, but IMO it is not that huge issue, since we maintain the CI.  I don't think user will be somewhat bothered by this.

> But since other images won't use newest version of container-common-scripts its PRs will fail.

Agreed. IMO in such case it would be the right time to update `common` in the remote repos; and yes -> I don't claim that such hacks are not expensive.

So, since the version check was already merged and is not the part of this PR, what about to remove that later (and let others ack this?)?

omron93: > So, since the version check was already merged and is not the part of this PR, what about to remove that later (and let others ack this?)?

You are right

praiskup:
> @praiskup Or if you would like to avoid this, some "random" image name can be generated for tests only and removed after.

Can be, but it brings some complexity -- and I'm not good enough with s2i testing, so I would rather delay this for after-PR time :) if any.  Even as-is, the PR somewhat improves the resource management (major motivation for this PR) so I'm less motivated to solve more problems now.

common.mk
```diff
@@ -7,42 +10,69 @@ endif
...
+
+# The tests should ideally depend on $IMAGE_ID only, but see PR#19 for more info
+# while we need to depend on 'tag' instead of 'build'.
+test: tag
 	VERSIONS="$(VERSIONS)" $(script_env) $(test)
-	VERSIONS="$(VERSIONS)" $(script_env) $(tag)
 
 .PHONY: test-openshift
 test-openshift: script_env += TEST_OPENSHIFT_MODE=true
 test-openshift: $(VERSIONS)
```
pkubatrh: imo `test-openshift` should depend on `tag` as well

pkubatrh:
otherwise lgtm

common.mk
praiskup: right, good catch

praiskup:
Fixed the `test-openshift` target, and `build-all`target so it doesn't print not-right-now-built images.  Since we have basically 2xLGTM now, I'm merging this.

Issue #4, opened 2017-04-25
`latest` directory in container git repositories?


praiskup:
It is really ugly that we loose the git history all the time with new versions of our containers, speaking in examples - consider PostgreSQL container:

  1. working with `9.2` directory, we iterated a lot (git blame and git history gives us useful info
  2. but then we needed to work on `9.4`, so we copied and pasted whole `9.2` directory into `9.4`, which means that the whole `9.4` directory lost all the useful commit history
  3. moving to `9.5` made this even worse, and upcoming `9.6` won't help obviously

Well, e.g. in Fedora dist-git, we have `master` branch which guarantees that all the necessary info for `git blame` persists. It is fair also to keep the "authorship" of contributions.

Perhaps we should have e.g. `latest` directory now (instead of `9.6`) and make the `9.6` be a symbolic link to `latest`?

omron93:
+1

To note, it seems to me that change only point 2. would work too for some git commands (eg. `git blame`).

2. but then we needed to work on 9.4, so we rename whole 9.2 directory into 9.4 and commit this. Then copy 9.4 as 9.2 and add it. Push these two commits together.

But having `latest` directory seems more clean to me. Also
@bparees Are there any requirements for having directory for each version? I know that images should be able to build with `docker build` command.

In case we want to solve problem with multiple versions and would change the workflow - adding latest directory.
So why not to remove directories and have only one directory with dockerfiles: `Dockerfile.9.2`, `Dockerfile.9.4`, `Dockerfile.9.5`. Have only one scripts and in dockerfiles set some for example `VERSION` variable to allow differences in scripts.

pkubatrh:
+1 from me, having latest with the full git history would be really useful. I often find myself trying to find out what got changed where.

>So why not to remove directories and have only one directory with dockerfiles: Dockerfile.9.2, Dockerfile.9.4, Dockerfile.9.5. Have only one scripts and in dockerfiles set some for example VERSION variable to allow differences in scripts.

This would work in case there are no major differences outside of Dockerfiles between the images (which I think is the case for most of the images right now). But not sure if that can be guaranteed (=we want to work around) in future.

bparees:
> @bparees Are there any requirements for having directory for each version?

there are no requirements, since you guys own the CI/CD infrastructure and getting the images built on OSBS, you're welcome to structure the directories in whatever way you see fit.  As long as users can easily clone the repo and

1) see/understand which files are for a given version
2) build an image for a given version

[...]

PR #20, opened 2017-08-10
Add EXCLUDE_VERSIONS to explicitly set which versions not to build

pkubatrh:
This is connected to #4 . As we are now pushing towards having the whole git history available in a single directory (be-it `latest` or the highest version available so far) to use `git blame` easily, we need to make sure we are able to both keep the Dockerfile.version history when a new version is introduced (which is most likely going to be the latest one) and to make the tests not fail in case centos/rhel packages are still not ready for the new version.

Currently to make sure the tests do not fail we would have to remove the Dockerfile.version specific to the distribution which does not yet have the packages available. This would however get rid of the git history of the Dockerfile we have been trying to preserve. This PR aims to fix this problem by introducing `EXCLUDE_VERSIONS` to `common.mk` and `build.sh` that can be used to explicitly provide information on which versions for which distribution to ignore.

@hhorak @torsava @praiskup WDYT

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -16,6 +16,8 @@ DOCKERFILE_PATH=""
 # Perform docker build but append the LABEL with GIT commit id at the end
 function docker_build_with_version {
   local dockerfile="$1"
+  local exclude=.exclude-${OS}
+  [ -e $exclude ] && echo "-> $exclude file exists for version $dir, skipping build." && return
   [ ! -e "$dockerfile" ] && echo "-> $dockerfile for version $dir does not exist, skipping build." && return
   echo "-> Version ${dir}: building image from '${dockerfile}' ..."
 


pkubatrh:
I have also pushed this change to https://github.com/sclorg/container-common-scripts/tree/exclude_ver so that the CI can pickup the commit properly inside https://github.com/sclorg/postgresql-container/pull/182

build.sh
```diff
@@ -16,6 +16,7 @@ DOCKERFILE_PATH=""
 # Perform docker build but append the LABEL with GIT commit id at the end
 function docker_build_with_version {
   local dockerfile="$1"
+  [ -n "$(echo "${EXCLUDE_VERSIONS}" | grep ${dir}.${OS})" ] && echo "-> $OS build of version $dir found in EXCLUDED_VERSIONS, skipping build." && return
```
hhorak: I'm just thinking whether we can restrict the regexp a bit, since if we ever had something like such two files `Dockerfile.fedora-module` and `Dockerfile.fedora` in the repo, while having `Dockerfile.fedora-module` in `EXCLUDE_VERSIONS`, `echo "${EXCLUDE_VERSIONS}" | grep ${dir}.${OS}` would match both `Dockerfile.fedora` and `Dockerfile.fedora-module`..

`grep -e "^\(.*\s\)\?${dir}.${OS}\(\s.*\)\?$"` looks terrible though..

pkubatrh:
Modified the change to not look at a `make` argument but instead check if a `.exclude-$OS` file is present in the versioned directory. If it exists, do not build the corresponding distribution version.

praiskup:
lgtm

praiskup:
As there's no other response, I'm merging.  But even if there were future complaints -- with this one-liner implementation we can introduce other black-listing methods while keeping this one for compatibility reasons.

PR #21, opened 2017-08-18
build.sh: fix /dev/pts/ permission issues

praiskup:
The permissions/ownership of the /dev/pts/* file (symlinked from
/dev/fd/$fd) is uncertain, so we can not rely on the fact that we
can write to this file (happened to me when I ssh as 'root' to my
RHEL7 testing box and then I 'sudo su - foo').

Since we already depend on Bash features extensively, let's rather
let the shell allocate a temporary fifo file with >(cmd) feature.
While we are on it, separate the parsing method into self-standing
method which can be easily reused.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,6 +13,41 @@ VERSION=${2-$VERSION}
 
 DOCKERFILE_PATH=""
 
+# parse_output COMMAND FILTER_COMMAND OUTVAR [STREAM]
+# ---------------------------------------------------
+# Parse stdout (STREAM='stdout') or stderr (STREAM='stdout') of COMMAND with
+# FILTER_COMMAND and store the output into variable named OUTVAR.
+# The filtered output stays (live) printed to terminal.  This method doesn't
+# create any explicit temporary files.
+# Defines:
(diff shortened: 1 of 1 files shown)


praiskup:
@pkubatrh or @omron93, can you have a quick look at this?

praiskup:
Argh, I forgot about #19 -- I pushed updated version to comply.

pkubatrh:
lgtm

build.sh
```diff
@@ -13,6 +13,41 @@ VERSION=${2-$VERSION}
 
 DOCKERFILE_PATH=""
 
+# parse_output COMMAND FILTER_COMMAND OUTVAR [STREAM]
+# ---------------------------------------------------
+# Parse stdout (STREAM='stdout') or stderr (STREAM='stdout') of COMMAND with
```
pkubatrh: Should be `... stderr (STREAM='stderr')`

pkubatrh:
Might be a good idea to mention in the docstring that when the 4th (`STREAM`) argument is not provided `parse_output` defaults to stdout, otherwise lgtm.

praiskup:
I'll mention that and merge, thanks for having a look!

PR #23, opened 2017-08-21
tests: execute remote container tests

praiskup:
This is mainly for jenkins CI and it is only the first step
(turning on PostgreSQL tests for now)!  In future we should have
some local-only (and less expensive) tests, too.

Jenkins is only responsible for setting 'OS' variable (defaults to
centos7).

When some changes are needed in remote git repositories (say
PostgreSQL container) one have to first submit a pull request
against that git repository where one of the commits needs to
contain commit message with line of this format:

    Required-by: postgresql-container#111

.. where 'postgresql-container' is the name of project within
sclorg group, and '111' string is the id of already submitted pull
request.  Local testing against such pull-requests works too
(unless the actual branch is not yet merged into 'origin/master'
branch).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,17 @@ Common build helpers for sclorg containers
 This repository is aimed to be added as git submodule into particular
 containers' source repositories.  By default, the path to submodule should be
 named 'common'.
+
+Regression tests
+----------------
+
+Just run `make check`
+
+Dependencies for testsuite:
+
+- docker
+- docker-squash (version 1.0.5)
+- git
+- go-md2man
+- make
+- source-to-image
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,12 @@
+all:
+	@echo >&2 "Only 'make check' allowed"
+
+
+TESTED_IMAGES = \
+	postgresql-container \
+	s2i-python-container
+
+.PHONY: check test all
+test: check
+check:
+	TESTED_IMAGES="$(TESTED_IMAGES)" tests/remote-containers.sh
--- a/build.sh
+++ b/build.sh
@@ -13,6 +13,8 @@ VERSION=${2-$VERSION}
 
 DOCKERFILE_PATH=""
 
+error() { echo "ERROR: $*" ; false ; }
+
 # parse_output COMMAND FILTER_COMMAND OUTVAR [STREAM={stderr|stdout}]
 # -------------------------------------------------------------------
 # Parse standard (error) output of COMMAND with FILTER_COMMAND and store the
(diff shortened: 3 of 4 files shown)


praiskup:
@omron93 can you please have a look at the API (whether it is OK for jenkins)?  @pkubatrh, @hhorak any comment is welcome, too

omron93:
Thanks. Nice idea to include reference of project+PR which should be tested in git log.

@praiskup What about adding some "default" value for "Required-by"? - to be able to easily test small changes in this repo (which does not need any changes in image repositories). For example to test master branch of postgresql-container and s2i-python-container to make sure that even small changes to this repository does not break building and without forcing to add to each commit same "Required-by" values.

I'll try to setup CI for this repo...

praiskup:
On Monday, August 21, 2017 2:28:14 PM CEST Marek Skalický wrote:
> @praiskup What about adding some "default" value for "Required-by"? - to be able to easily test small changes in this repo (which does not need any changes in image repositories).

There's by default mentioned 'postgresql-container#master' (I had a bug in Makefile though, so I fixed it now);  and we can add more with time.  Or better fine-tune the list of tests etc.  Take this as a first step please :)

> For example to test master branch of postgresql-container and s2i-python-container to make sure that even small changes to this repository does not break building and without forcing to add to each commit same "Required-by" values.

Ok, I've added s2i-python-container too.

> I'll try to setup CI for this repo...

Thanks!

hhorak:
It looks good to me, I think even running just two containers every-time should reveal some unexpected incompatibilities, and the rest will be caught when updating submodules.. Anyway, the ability to pick some particular PR seems like power enough extension, which might help as well.

hhorak:
So, any other suggestions or can we merge? and try to deploy into jenkins?

praiskup:
@omron93 promised to have a look at incorporating this into jenkins soon :)

omron93:
[test]

omron93:
[test]

praiskup:
Sounds like CI is doing something!  Thanks @omron93.  Let me rebase this to have 0f93964ea15ba4688233 in.

omron93:
>  Let me rebase this to have 0f93964 in.

You are right.

CI should now be working for this repo in same way as for other container repositories.

For commits from members of `sclorg` organization tests are triggered automatically. Also these members can triggers them by adding `[test]` into commit message.

praiskup:
But 0f93964 is in.  Anybody knows the reason for this message to show up?
`error: Setup script exited with error in docker-py setup command: Invalid environment marker: python_version < "3.5"`

praiskup:
0f93964 is already in, so it means that the commit was useless and the docker-squash still fails to install on RHEL7/CentOS7

omron93:
Now using python27-python-virtualenv for installing docker-squash and I hope it helped.

[test]

README.md
```diff
@@ -4,3 +4,15 @@ Common build helpers for sclorg containers
...
+----------------
+
+Just run `make check`
+
+Dependencies for testsuite:
+
+- docker
+- docker-squash (version 1.0.5)
+- git
+- make
```
omron93: Also `source-to-image` and `go-md2man` is needed for testsuites of tested images)

omron93:
@praiskup One question: is the default list of tested images be cleaned when some explicit "Required by:" is present? I think it should be...

README.md
praiskup: I'll add those, thanks.

praiskup:
@omron93  wrote:
> is the default list of tested images be cleaned

It is not (intentionally).  If one of the tested containers is broken by "commons PR", it should be mandatory for the guy who submits the PR to also submit pull-request against the broken container.

So there's no way to actually avoid testing one of the already listed images.  I don't thihk this is anyhow important implementation detail;  we can discuss peculiarities later.

omron93:
> It is not (intentionally). If one of the tested containers is broken by "commons PR", it should be mandatory for the guy who submits the PR to also submit pull-request against the broken container.

It makes sense. Two additional PRs isn't so much.

LGTM from me.

praiskup:
Thanks @omron93!  Merging.

Issue #22, opened 2017-08-18
RHEL7 build issue: installing docker_py=1.10.6 fails


praiskup:
Seems like this is result of some work-around from 5fe4cbbea27b2299e61e889f78414a9ec65ee70a commit.   But explicitly installing v1.10.6 fails for me with:

```
$ easy_install -q --user docker_py==1.10.6 docker-squash==1.0.5
error: Setup script exited with error in docker-py setup command: Invalid environment marker: python_version < "3.5"
```

OTOH, `easy_install -q --user docker-squash==1.0.5` works just fine. Do we still need explicit version of docker_py?

omron93:
I don't see reason for this workaround - it is quite old so I guess a problem could be fixed now.

PR #24, opened 2017-08-21
build: docker_py==1.10.6 not easily installable on RHEL7

praiskup:
This pipy package fails to install because RHEL7 doesn't have
python 3 available by default:

  $ easy_install -q --user docker_py==1.10.6 docker-squash==1.0.5
  error: Setup script exited with error in docker-py setup
    command: Invalid environment marker: python_version < "3.5

But turns out we don't have to be that explicit about the
docker_py version nowadays.

Reverts: 5fe4cbbea27b2299e61e889f78414a9ec65ee70a
Fixes: #22

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -88,7 +88,7 @@ function docker_build_with_version {
 function squash {
   # FIXME: We have to use the exact versions here to avoid Docker client
   #        compatibility issues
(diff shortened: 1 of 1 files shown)


pkubatrh:
lgtm, merging.

PR #30, opened 2017-09-08
Allow to change the docker build context

hhorak:
In order to allow sharing files in the parent directory, instead of having copies of the same files for every version, we need to be able to set the context of the docker build command for every image specifically. This change defines a new variables DOCKER_BUILD_CONTEXT that may be defined in the Makefile, typically to '..', so docker build will be run with the context of the parent directory.

@praiskup @pkubatrh Does this seem sane?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -61,7 +61,7 @@ function docker_build_with_version {
     BUILD_OPTIONS+=" --pull=true"
   fi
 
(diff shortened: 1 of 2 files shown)


praiskup:
LGTM, let's fix issues later, if any.

PR #31, opened 2017-09-11
tests: git show --no-patch doesn't work on el7

praiskup:


praiskup:
Does this look good?  If yes, I'll fix the "summary" in the patch  and push (no need to respin the CI again).

tests/remote-containers.sh
```diff
@@ -28,23 +28,20 @@ analyse_commits ()
...
+    while read line; do
+        case $line in
+            Required-by:\ *)
+                set -- $line
+                old_IFS=$IFS
+                IFS=\#
+                set -- $2
+                IFS=$old_IFS
+                set -- $1 $2
+                info "Commit $commit sets $1 to PR $2"
```
pkubatrh: `$commit` is no longer set in this version

praiskup: Ouch, good catch, I'll apply this too.
```diff
-                info "Commit $commit sets $1 to PR $2"
+                info "Testing $1 from PR $2"
```

PR #33, opened 2017-09-17
Add test-lib-openshift.sh for testing in OpenShift

hhorak:
Add test-lib-openshift.sh with set of functions for testing docker images in OpenShift using 'oc'


praiskup:
Thanks!

Issue #3, opened 2017-04-06
Ensure that `umask` doesn't affect the container build


praiskup:
* git doesn't track file permissions, thus
* git clone initiates the permissions according the user's umask value
* docker build just copies the files as-is into container
* especial problems are with `root/usr` directory created by `ADD root /` command, this affects the whole image (`/usr` directory might be unreadable)

In RPMs' specfiles, we explicitly set the permissions in %install phase (or by %attr in %files), though there's no such way in `Dockerfile`.

omron93:
Related to https://github.com/sclorg/mariadb-container/issues/29

PR #38, opened 2017-11-21
Distgen - support for image generation

dhodovsk:
This PR was created as reaction to suggestions in https://github.com/sclorg/postgresql-container/pull/203. It should be an initial effort in using distgen in image repositories.

**Notes:**
- The image repository that uses make distgen should contain:
  - specs/multispec.yml file
  - manifest.sh - lists of files to generate, files to copy (including desired mode) and list of symlinks
 - Results are stored in current directory when not specified in DESTDIR in manifest

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -13,7 +13,6 @@ build = $(SHELL) $(common_dir)/build.sh
 test =  $(SHELL) $(common_dir)/test.sh
 tag =   $(SHELL) $(common_dir)/tag.sh
 clean = $(SHELL) $(common_dir)/clean.sh
-generator = $(SHELL) $(common_dir)/generate.sh
 
 ifeq ($(TARGET),rhel7)
 	SKIP_SQUASH ?= 0
@@ -30,13 +29,11 @@ else ifeq ($(TARGET),centos6)
 else
 	OS := centos7
 	DOCKERFILE ?= Dockerfile
-	DG_CONF ?= centos-7-x86_64.yaml
+	DG_CONF := centos-7-x86_64.yaml
 endif
 
 SKIP_SQUASH ?= 1
 DOCKER_BUILD_CONTEXT ?= .
-DISTGEN_BIN ?= /usr/bin/dg
-MANIFEST_FILE ?= manifest.sh
 
 script_env = \
 	SKIP_SQUASH=$(SKIP_SQUASH)                      \
@@ -46,11 +43,6 @@ script_env = \
 	DOCKER_BUILD_CONTEXT=$(DOCKER_BUILD_CONTEXT)    \
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"  \
 	CUSTOM_REPO="$(CUSTOM_REPO)"
-
-generation_env = \
-	DG_CONF=$(DG_CONF) \
-	DG=$(DISTGEN_BIN) \
-	MANIFEST_FILE=$(MANIFEST_FILE)
 
 # TODO: switch to 'build: build-all' once parallel builds are relatively safe
 .PHONY: build build-serial build-all
@@ -99,14 +91,7 @@ clean:
 
 .PHONY: generate
 generate:
-	@$(MAKE) auto_targets.mk
-	@$(MAKE) exec-gen-rules
-	rm auto_targets.mk
+	$(MAKE) VERSIONS="$(VERSIONS)" DG_CONF=$(DG_CONF) -f $(common_dir)/gen.mk gen
 
-auto_targets.mk: $(generator) $(MANIFEST_FILE)
-	VERSIONS="$(VERSIONS)" $(generation_env) $(generator)
-
-include auto_targets.mk
-
-.PHONY: exec-gen-rules
-exec-gen-rules: $(DISTGEN_TARGETS) $(COPY_TARGETS) $(SYMLINK_TARGETS)
+generate-all:
+	$(MAKE) VERSIONS="$(VERSIONS)" DG_CONF="rhel-7-x86_64.yaml fedora-27-x86_64.yaml centos-7-x86_64.yaml" -f $(common_dir)/gen.mk gen
--- a/common.mk
+++ b/common.mk
@@ -13,24 +13,30 @@ build = $(SHELL) $(common_dir)/build.sh
 test =  $(SHELL) $(common_dir)/test.sh
 tag =   $(SHELL) $(common_dir)/tag.sh
 clean = $(SHELL) $(common_dir)/clean.sh
+generator = $(SHELL) $(common_dir)/generate.sh
 
 ifeq ($(TARGET),rhel7)
 	SKIP_SQUASH ?= 0
 	OS := rhel7
 	DOCKERFILE ?= Dockerfile.rhel7
+	DG_CONF := rhel-7-x86_64.yaml
 else ifeq ($(TARGET),fedora)
 	OS := fedora
 	DOCKERFILE ?= Dockerfile.fedora
+	DG_CONF := fedora-27-x86_64.yaml
 else ifeq ($(TARGET),centos6)
 	OS := centos6
 	DOCKERFILE ?= Dockerfile.centos6
 else
 	OS := centos7
 	DOCKERFILE ?= Dockerfile
+	DG_CONF ?= centos-7-x86_64.yaml
 endif
 
 SKIP_SQUASH ?= 1
 DOCKER_BUILD_CONTEXT ?= .
+DISTGEN_BIN ?= /usr/bin/dg
+MANIFEST_FILE ?= manifest.sh
 
 script_env = \
 	SKIP_SQUASH=$(SKIP_SQUASH)                      \
@@ -40,6 +46,11 @@ script_env = \
 	DOCKER_BUILD_CONTEXT=$(DOCKER_BUILD_CONTEXT)    \
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"  \
 	CUSTOM_REPO="$(CUSTOM_REPO)"
+
+generation_env = \
+	DG_CONF=$(DG_CONF) \
+	DG=$(DISTGEN_BIN) \
+	MANIFEST_FILE=$(MANIFEST_FILE)
 
 # TODO: switch to 'build: build-all' once parallel builds are relatively safe
 .PHONY: build build-serial build-all
@@ -85,3 +96,17 @@ clean:
 	mkdir -p $(@D)
 	go-md2man -in "$^" -out "$@"
 	chmod a+r "$@"
+
+.PHONY: generate
+generate:
+	@$(MAKE) auto_targets.mk
+	@$(MAKE) exec-gen-rules
+	rm auto_targets.mk
+
+auto_targets.mk: $(generator) $(MANIFEST_FILE)
+	VERSIONS="$(VERSIONS)" $(generation_env) $(generator)
+
+include auto_targets.mk
+
+.PHONY: exec-gen-rules
+exec-gen-rules: $(DISTGEN_TARGETS) $(COPY_TARGETS) $(SYMLINK_TARGETS)
--- a/gen.mk
+++ b/gen.mk
@@ -1,0 +1,27 @@
+# Helper for generating image repository files
+
+ifndef common_dir
+    common_dir = common
+endif
+
+generator = $(common_dir)/generate.sh
+# default path for dg binary from package distgen
+DISTGEN_BIN ?= /usr/bin/dg
+MANIFEST_FILE ?= manifest.sh
+
+generation_env = \
+	DG=$(DISTGEN_BIN) \
+	MANIFEST_FILE=$(MANIFEST_FILE)
+
+.PHONY: gen
+gen: auto_targets.mk exec-gen-rules
+	rm auto_targets.mk
+
+
+auto_targets.mk: $(generator) $(MANIFEST_FILE)
+	VERSIONS="$(VERSIONS)" DG_CONF="$(DG_CONF)" $(generation_env) $(generator)
+
+include auto_targets.mk
+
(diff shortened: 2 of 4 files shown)


distgen.sh
```diff
@@ -0,0 +1,75 @@
...
+fi
+
+WORKDIR=${PWD}
+DG=${DG:-/usr/bin/dg}
+DG_CONF=${DG_CONF:-centos-7-x86_64.yaml}
+
+cd templates
+OUTPUT_DIR=$(mktemp -d ${WORKDIR}/${OS}-${VERSION}-XXXX)
+echo "Copying files to ${OUTPUT_DIR}:"
+for file in ${FILES_TO_COPY}; do
```
praiskup: I would really prefer having this solved properly through Makefile targets, so
everything is fully parallelized (even this cp && chmod thingy, but especially
the $FILES_TO_GEN bellow).

Thanks for the chmod call btw.!  Could we made that "optional" with some sane
default (say 644)?  Btw., this could solve #3 pretty nicely :-).

distgen.sh
```diff
@@ -0,0 +1,75 @@
...
+    echo "Templates for distgen not found"
+    exit
+fi
+
+# expecting file dg_manifest that contains variables:
+#   FILES_TO_COPY in form "path1:mode1 path2:mode2..."
+#   FILES_TO_GEN in form similar to FILES_TO_COPY
+#   INTERNAL_SYMLINKS in form "name1->target1 name1->target1..."
+if [ -f "./dg_manifest" ]; then
+    source ./dg_manifest
```
praiskup: Since manifest file is shell script, could we rename it to `dg_manifest.sh` (or rather remove the `dg_` prefix, as that's not related to distgen) and generate some Makefile snippet from this automatically?   The snippet could be then included.

I sort of miss the possibility to relocate the file (e.g you can not copy/symlink one **source** file onto two places), note that the quite old proposal here https://github.com/devexp-db/cont-postgresql/blob/master/cl-manifest allows this to be specified.

distgen.sh
```diff
@@ -0,0 +1,75 @@
...
+    echo " ${path} with mode - ${mode:-default}"
+    cp --parents ${path} ${OUTPUT_DIR}
+    if [ ! -z "${mode}" ]; then
+        chmod ${mode} ${OUTPUT_DIR}/${path}
+    fi
+done
+
+echo "Generating files:"
+for file in ${FILES_TO_GEN}; do
+    split=(${file//:/ })
```
praiskup: This is not safe for spaces, better would be "old_IFS=$IFS ; IFS=: ; set -- $file; ... use $1/$2 ... IFS=$old_IFS.

distgen.sh
```diff
@@ -0,0 +1,75 @@
...
+    path=${split[0]}
+    mode=${split[1]}
+
+    if [ ! -d $(dirname ${OUTPUT_DIR}/$path) ]; then
+        mkdir -p dirname ${OUTPUT_DIR}/$path
```
praiskup: shell quotes missing

distgen.sh
```diff
@@ -0,0 +1,75 @@
...
+    fi
+
+    # With max-passes > 1 is this directive non-effective
+    if grep "{%- raw %}" $path > /dev/null; then
+        max_passes=1
```
praiskup: Can you elaborate on this :-), I don't understand much what you try to achieve here....  The grep eats another fork, and is incomplete (one could use `{% raw %}` or `{%- raw -%}`).

dhodovsk: Problem description:
- some template files could contain some chars that are in fact jinja directives - e.g bash variable length vs. jinja comment - `${#myvar}`
- the directive {% raw %}/{% endraw %} fixes this problem, but the file can't be processed more than once
- while using `make distgen` we don't know whether template needs to be processed multiple times or just once.

Possible solutions
1,Add max-passes to manifest in form
```bash
# path:max-passes:mode
./Dockerfile:2:644
./script.sh::755
```
- with 1 as default - this could add complexity to UX

2, Expect jinja2.exceptions.TemplateSyntaxError - could be problem when there is an actual error

3, ... (please add your suggestion)

@praiskup @bkabrda what do you think?

bkabrda: I actually already changed the mechanism of rerendering to not rerender the whole template, but just the spec values [1], which means this shouldn't be an issue going forward. I can do the release today and that would effectively make this problem go away, IIUC. @dhodovsk WDYT?

[1] https://github.com/devexp-db/distgen/pull/63

dhodovsk: @bkabrda I think it is great, would solve this issue nicely and also speed up the generation.

bkabrda: I released distgen 0.20 which should fix this.

praiskup:
[test]

dhodovsk:
@praiskup After our discussion I changed a bit the desired look of manifest file, so it is more readable. Please, let me know if you think it is ok.
Example can be found here: https://github.com/dhodovsk/postgresql-container/blob/ecd036610006bf9af4b0878e145273bfee51a15d/manifest.sh

hhorak:
@praiskup I don't see any other issues with the current status, but I never paid big attention into this distgen, so I'll keep this to you.. WDYT?

praiskup:
IMO we got much more talkative version of the manifest file than originally proposed variant https://github.com/devexp-db/cont-postgresql/blob/master/cl-manifest, but I don't mind in particular - hopefully that manifest format can be attractive enough, and we don't know until we try that..

So more importantly - as this particular PR is just implementation detail - could we vote in sclorg/postgresql-container#208 whether the format is generally OK?

common.mk
```diff
@@ -11,21 +11,27 @@ build = $(common_dir)/build.sh
...
 
 ifeq ($(TARGET),rhel7)
 	SKIP_SQUASH ?= 0
 	OS := rhel7
 	DOCKERFILE ?= Dockerfile.rhel7
+	DG_CONF := rhel-7-x86_64.yaml
 else ifeq ($(TARGET),fedora)
 	OS := fedora
 	DOCKERFILE ?= Dockerfile.fedora
+	DG_CONF := fedora-26-x86_64.yaml
```
TomasTomecek: nit: Fedora 27 would be lovely as a start

hhorak:
I've taken a look again and found nothing suspicious. So, I'd say let's give it a try and see how it flies.

dhodovsk:
The implementation generates Dockerfile for desired TARGET without postfix. There has been requirement to generate Dockerfiles for all targets (Dockerfile, Dockerfile.rhel, Dockerfile.fedora) no matter what is in TARGET. That would also fix incorrect behaviour with two subsequent generations and change of target:
```
$ make generate TARGET=centos7 -> correctly generates centos7 Dockerfile
$ make generate TARGET=rhel7 -> doesn't do anything due to Dockerfile created in previous run (and no change in manifest or source files)
```
I postponed working on this due to change of priorities. In case of need I can change it back.

praiskup:
[test]

pkubatrh:
>There has been requirement to generate Dockerfiles for all targets (Dockerfile, Dockerfile.rhel, Dockerfile.fedora) no matter what is in TARGET.

I say keep the the functionality of generating only a single version of Dockerfile when TARGET is set. Adding a new makefile target (generate-all?) that creates all versions (TARGETs) available would be nice.

Both use cases still need the postfixed Dockerfiles though.

generate.sh
```diff
@@ -0,0 +1,119 @@
...
+
+EOF
+    done
+    IFS=$OLD_IFS
+}
+
+for version in ${VERSIONS}; do
+    # copy targets
+    rules="$COPY_RULES"
+    core="cp \$< \$@ ; \\"
```
omron93: So it won't be possible to copy directories?

dhodovsk: @omron93 It is not possible now, one has to specify all files that are being generated.

omron93:
> Can we expect that all files in directories except of Dockerfile are not affected by distribution change? When running generate-all for one version, one directory with all Dockerfiles would be present?

README ?

dhodovsk:
@omron93 I am asking because AFAIK e.g for posgres there is one set of files for every version and they are os agnostic. Based on your reaction there are distro specific files. How do you handle these kinds of files, do you have specific directories for version and distro, e.g rhel-nginx-1.2?

omron93:
>  Based on your reaction there are distro specific files. How do you handle these kinds of files, do you have specific directories for version and distro, e.g rhel-nginx-1.2?

Name of the image is distro specific. So for examples in README could be distro specific...

Currently we don't do this. So rhel based names are used in examples... But it would be useful to be able to generate right names. But it is nothing which have to be implemented now.

dhodovsk:
So now the desired state is for example in postgres:
 - `make generate VERSIONS=9.6 TARGET=rhel7` -> generates one directory: 9.6 with one Dockerfile, called Dockerfile.rhel (+ all other files)
 - `make generate` -> generates 3 directories and in each of them is one Dockerfile without postfix - since the target is default - centos
 - `make generate-all VERSIONS=9.6` -> generates one directory containing Dockerfile, Dockerfile.rhel and Dockerfile.fedora

Is that correct, @omron93, @pkubatrh ?

pkubatrh:
Also `make generate-all` should generate directories for all versions containing all Dockerfiles.

It should be easy to configure which suffix each of the Dockerfiles will have. For example right now we are using `rhel7` for our rhel images (since they are based on rhel7) but we might want to have a different extension for future rhel images.

dhodovsk:
Requested changes are now applied. What has been changed for image repository maintainer is that desired distro specific Dockerfile dest needs to be added to manifest.

Before:
```
...
DISTGEN_RULES="
    src=src/Dockerfile
    dest=Dockerfile;

    src=src/cccp.yml
    ...
```
After:
```
...
DISTGEN_RULES="
    src=src/Dockerfile
    dest=Dockerfile;

    src=src/Dockerfile
    dest=Dockerfile.rhel7;

    src=src/cccp.yml
    ...
```

The Dockerfile dest will be generated only when related TARGET is provided or by using generate-all.

praiskup:
[test]

phracek:
thx, please merge it.

common.mk
```diff
@@ -41,6 +44,7 @@ script_env = \
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"  \
 	CUSTOM_REPO="$(CUSTOM_REPO)"
 
+
```
pvalena: Formatting change, just saying....

gen.mk
```diff
@@ -0,0 +1,26 @@
...
+generation_env = \
+	DG=$(DISTGEN_BIN) \
+	MANIFEST_FILE=$(MANIFEST_FILE)
+
+.PHONY: gen
+gen: auto_targets.mk exec-gen-rules
+	rm auto_targets.mk
+
+
+auto_targets.mk: $(generator) $(MANIFEST_FILE)
```
pvalena: Why this one has `.mk` at the end? Please forgive my ignorance...

pkubatrh:
Would it be possible to remove the `auto_targets.mk` file in case of a failure? Or maybe just clean it up inside `make clean`.

dhodovsk:
@pkubatrh ofc, on it :)

gen.mk
dhodovsk: Np, auto_targets.mk is generated makefile containing rules for all targets to be created.

gen.mk
```diff
@@ -0,0 +1,26 @@
+# Helper for generating image repository files
+
+ifndef common_dir
+    common_dir = common
+endif
+
+generator = $(common_dir)/generate.sh
+DISTGEN_BIN ?= /usr/bin/dg
```
pvalena: Could you somewhere fail, or warn at least if user does not have this binary? Where does one install it from?

common.mk
```diff
@@ -18,15 +18,18 @@ ifeq ($(TARGET),rhel7)
 	SKIP_SQUASH ?= 0
 	OS := rhel7
 	DOCKERFILE ?= Dockerfile.rhel7
+	DG_CONF := rhel-7-x86_64.yaml
 else ifeq ($(TARGET),fedora)
 	OS := fedora
 	DOCKERFILE ?= Dockerfile.fedora
+	DG_CONF := fedora-27-x86_64.yaml
```
pvalena: Whe is there `:=` when in other cases we use `?=` ?
Same for rhel7.

pvalena:
Otherwise LGTM :). Good job.

gen.mk
pkubatrh: It is. `gen` depends on `auto_targets.mk` which in turn depends on `generator.sh` in which the actual file is generated.

gen.mk
TomasTomecek: dnf install distgen :D

common.mk
dhodovsk: You are right, this should be consistent.

dhodovsk:
[test]

TomasTomecek:
Hm, tests are failing :/ is the failure related to this PR?

pkubatrh:
>Hm, tests are failing :/ is the failure related to this PR?

Looks like the fail happens during checking for an expected failure, so does not seem related.

pkubatrh:
Actually, it does not fail on the expected failure but on `make clean` when it tests for the existence of the `auto_targets.mk` file ...

common.mk
```diff
@@ -79,9 +82,17 @@ tag: build
 
 .PHONY: clean
 clean:
+	test -f auto_targets.mk && rm auto_targets.mk
```
pkubatrh: @dhodovsk Moving this line directly into the `clean.sh` script should get rid of the test failures.

dhodovsk: yeas, sorry about that :)

dhodovsk:
[test]

dhodovsk:
[test]

omron93:
LGTM. Thanks for this PR.

praiskup:
Thanks!

TomasTomecek:
Party!

PR #40, opened 2017-11-29
Fix OpenShift test library - oc cluster up fails

omron93:
- Don't overwrite system default docker options - during adding OS insecure registries.
- Flush iptables after disabling firewalld

Second point is needed to be able to run `oc cluster up `. Otherwise it fails with `Error: Get https://127.0.0.1:8443/api/v1/nodes: dial tcp 127.0.0.1:8443: getsockopt: connection refused`.

@praiskup @pkubatrh Please take a look and merge. To be able to test https://github.com/sclorg/postgresql-container/pull/209


pkubatrh:
lgtm

PR #43, opened 2017-12-01
Wait in ct_os_cluster_up till OpenShift becomes ready

omron93:
Wait in `ct_os_cluster_up` untill OpenShift registry and router become ready.

Add function `ct_os_wait_rc_ready` which waits till replication controller have desired number of pods.

@pkubatrh or @praiskup Please take a look. This change is required to fix 'rhel7 openshift' CI in  https://github.com/sclorg/postgresql-container/pull/209 (registry is not ready in 30s)

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -94,6 +94,27 @@ function ct_os_wait_pod_ready() {
   while ! ct_os_check_pod_readiness "${pod_prefix}" "true" ; do
     echo -n "."
     [ ${SECONDS} -gt ${timeout} ] && echo " FAIL" && return 1
+    SECONDS=$((SECONDS+3))
+    sleep 3
+  done
+  echo " DONE"
+}
+
+# ct_os_wait_rc_ready [pod_prefix, timeout]
+# --------------------
+# Wait maximum [timeout] for the rc having desired number of replicas ready.
+# Arguments: pod_prefix - prefix of the replication controller
+# Arguments: timeout - how many seconds to wait seconds
+function ct_os_wait_rc_ready() {
+  local pod_prefix="${1}" ; shift
+  local timeout="${1}" ; shift
+  SECONDS=0
+  echo -n "Waiting for ${pod_prefix} pod becoming ready ..."
+  while ! test "$(oc get rc -o custom-columns=NAME:.metadata.name,DESIRED:status.replicas,READY:status.readyReplicas 2>/dev/null \
+                 | grep "^${pod_prefix}" | awk '$2==$3 {print "ready"}')" == "ready" ; do
+    echo -n "."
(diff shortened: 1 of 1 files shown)


test-lib-openshift.sh
```diff
@@ -94,6 +94,27 @@ function ct_os_wait_pod_ready() {
...
     echo -n "."
     [ ${SECONDS} -gt ${timeout} ] && echo " FAIL" && return 1
+    SECONDS=$((SECONDS+3))
+    sleep 3
+  done
+  echo " DONE"
+}
+
+# ct_os_wait_rc_ready [pod_prefix, timeout]
+# --------------------
```
praiskup: Nit, with square brackets it usually means the arguement is optional. This header is more readable:
```
# ct_os_wait_rc_ready POD_PREFIX TIMEOUT
# --------------------------------------
```

test-lib-openshift.sh
```diff
@@ -94,6 +94,27 @@ function ct_os_wait_pod_ready() {
...
+# Wait maximum [timeout] for the rc having desired number of replicas ready.
+# Arguments: pod_prefix - prefix of the replication controller
+# Arguments: timeout - how many seconds to wait seconds
+function ct_os_wait_rc_ready() {
+  local pod_prefix="${1}" ; shift
+  local timeout="${1}" ; shift
+  SECONDS=0
+  echo -n "Waiting for ${pod_prefix} pod becoming ready ..."
+  while ! test "$(oc get rc -o custom-columns=NAME:.metadata.name,DESIRED:status.replicas,READY:status.readyReplicas 2>/dev/null \
+                 | grep ^${pod_prefix} | awk '$2==$3 {print "ready"}')" == "ready" ; do
```
praiskup: I prefer all shell vars quoted (except for in `case $var in` and assignments `foo=$bar`)...  But I think we don't have to block this PR with this ... let's go with shell-style fixes in different PR..

praiskup: Also this awk could be used:
```awk
BEGIN { exit_status=1; }                                                                                              
/^<pod_prefix>.*/ && $2 == $3 { exit_status=0 ; }                                                                                
END { exit exit_status; }     
```

test-lib-openshift.sh
omron93: Makes sense. Other function docs updated as well.

test-lib-openshift.sh
omron93: Double quotes added. If you prefer  awk solution I can change it.

praiskup:
Thank you.

praiskup:
Btw., @omron93, the CI doesn't seem to work now.

praiskup:
Ah, it's just not automatically started... [test]

omron93:
There is Jenkins upgrade by Central CI admins. But problems with plugins occurred so we have to wait. That is why CI is not working.

PR #45, opened 2017-12-07
test-libs: restart docker after firewalld stop and iptables flush

pkubatrh:
Related: https://github.com/sclorg/mariadb-container/pull/57


omron93:
Makes sense to me. Thanks.

PR #46, opened 2017-12-10
Add ct_test_response for generic http response testing

hhorak:
SSIA

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -299,13 +299,22 @@ function ct_os_cluster_running() {
 # Arguments: app - url or local path to git repo with the application sources.
 # Arguments: context_dir - sub-directory inside the repository with the application sources.
 # Arguments: expected_output - PCRE regular expression that must match the response body.
-# Arguments: env_params - environment variables parameters for the images.
+# Arguments: port - which port to use (default: 8080)
+# Arguments: protocol - which protocol to use (default: http)
+# Arguments: response_code - what http response code to expect (default: 200)
+# Arguments: readiness_timeout - how many seconds to wait till the s2i image is deployed (default: 300)
+# Arguments: env_params..* - all other arguments are used as additional parameters for the `oc new-app`
+#            command, typically environment variables
 function ct_os_test_s2i_app() {
   local image_name=${1}
   local app=${2}
   local context_dir=${3}
   local expected_output=${4}
-  shift 4
+  local port=${5:-8080}
+  local protocol=${6:-http}
+  local response_code=${7:-200}
+  local readiness_timeout=${8:-300}
+
   local image_name_no_namespace=${image_name##*/}
   local service_name="${image_name_no_namespace}-testing"
 
@@ -317,9 +326,9 @@ function ct_os_test_s2i_app() {
                           --context-dir="${context_dir}" \
                           --name "${service_name}" $@
 
-  ct_os_wait_pod_ready "${service_name}" 300
+  ct_os_wait_pod_ready "${service_name}" ${readiness_timeout}
   local ip=$(ct_os_get_service_ip "${service_name}")
-  ct_test_response "http://${ip}:8080" 200 "${expected_output}"
+  ct_test_response "${protocol}://${ip}:${port}" ${response_code} "${expected_output}"
 
   ct_os_delete_project
 }
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -328,7 +328,7 @@ function ct_os_test_s2i_app_func() {
   local ip=$(ct_os_get_service_ip "${service_name}")
   local check_command_exp=$(echo "$command" | sed -e "s/<IP>/$ip/g")
 
-  ${check_command_exp}
+  eval "$check_command_exp"
 
   ct_os_delete_project
 }
@@ -360,6 +360,6 @@ function ct_os_test_s2i_app() {
                           "${app}" \
                           "${context_dir}" \
                           "ct_test_response ${protocol}://<IP>:${port} ${response_code} ${expected_output}" \
-                          ${oc_args}
+                          "${oc_args}"
 }
 
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -146,6 +146,7 @@ function ct_os_deploy_s2i_image() {
   local app="${1}" ; shift
   # ignore error exit code, because oc new-app returns error when image exists
   oc new-app ${image}~${app} $@ || :
+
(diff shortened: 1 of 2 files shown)


test-lib-openshift.sh
```diff
@@ -291,4 +291,36 @@ function ct_os_cluster_running() {
...
+  # Create a specific imagestream tag for the image so that oc cannot use anything else
+  ct_os_upload_image "${image_name}" "${service_name}"
+
+  ct_os_deploy_s2i_image "${service_name}" "${app}" \
+                          --context-dir="${context_dir}" \
+                          --name "${service_name}" $@
+
+  ct_os_wait_pod_ready "${service_name}" 300
+  local ip=$(ct_os_get_service_ip "${service_name}")
+  ct_test_response "http://${ip}:8080" 200 "${expected_output}"
```
pkubatrh: Could we pass a custom response check function instead? Or in addition to this?

pkubatrh: imo it would make more sense to override this since some images do not provide any content available via http so the failure would have to be ignored.

hhorak: At this point I only added few more things to be configurable (port, protocol, ...). Would that address the use case you have in your mind?
For the other use cases, my thinking was more like to let users write own similar function. This one was meant for images that we might check using `curl` request.

pkubatrh: My idea was to use this for all of our s2i enabled images, so for example to be able to run `pg_isready` instead of `ct_test_response`.

In any case if this would be too much trouble to accomplish I am ok with providing a component specific test case for s2i builds.

omron93:
LGMT, thanks.

test-lib-openshift.sh
hhorak: @pkubatrh there is now `ct_os_test_s2i_app_func()` function that accepts a check function as argument. Your case with `pg_isready` might then be implemented like this:
```
test_s2i_pg() {
  ct_os_test_s2i_app_func "${IMAGE_NAME}" pg_test . "pg_isready -h <IP>" -e "POSTGRESQL_ADMIN_USER=standa"
}
```
What do you think?

pkubatrh: That is what I had in mind, thanks!
Although I am not sure the `ct_os_test_s2i_app` function is necessary. I feel like it would make more sense to use the `ct_os_test_s2i_app_func` instead (with the ct_test_response as the function argument).

test-lib-openshift.sh
```diff
@@ -291,4 +292,74 @@ function ct_os_cluster_running() {
...
+#            command, typically environment variables (optional)
+function ct_os_test_s2i_app() {
+  local image_name=${1}
+  local app=${2}
+  local context_dir=${3}
+  local expected_output=${4}
+  local port=${5:-8080}
+  local protocol=${6:-http}
+  local response_code=${7:-200}
+  local oc_args=${8:-}
```
pkubatrh: Should this not eat up rest of the arguments?

test-lib-openshift.sh
```diff
@@ -291,4 +292,74 @@ function ct_os_cluster_running() {
...
+# Arguments: context_dir - sub-directory inside the repository with the application sources (compulsory)
+# Arguments: check_command - CMD line that checks whether the container works (compulsory; '<IP>' will be replaced with actual IP)
+# Arguments: oc_args - all other arguments are used as additional parameters for the `oc new-app`
+#            command, typically environment variables (optional)
+function ct_os_test_s2i_app_func() {
+  local image_name=${1}
+  local app=${2}
+  local context_dir=${3}
+  local check_command=${4}
+  local oc_args=${5:-}
```
pkubatrh: Should this not eat up rest of the arguments?

test-lib-openshift.sh
pkubatrh: One reasoning for this is that `ct_os_test_s2i_app` requires to source `test-lib.sh` to work. But as nothing inside the actual test script might directly use the library it might be not immediately clear what the library is sourced for.

test-lib-openshift.sh
hhorak: @pkubatrh I was thinking about the approach that you suggest as well, but it would prevent us adding new options in a compatible way in the future. I'm not sure whether that is the correct priority here..

test-lib-openshift.sh
```diff
@@ -291,4 +292,74 @@ function ct_os_cluster_running() {
...
+                          --context-dir="${context_dir}" \
+                          --name "${service_name}" \
+                          ${oc_args}
+
+  ct_os_wait_pod_ready "${service_name}" 300
+
+  local ip=$(ct_os_get_service_ip "${service_name}")
+  local check_command_exp=$(echo "$command" | sed -e "s/<IP>/$ip/g")
+
+  ${check_command_exp}
```
praiskup: Just a small suggestion, in such situations it is preferred to have quoted command, and:
```patch
- ${check_command_exp}
+ eval "$check_command_exp"
```

Usually we don't have that huge flexibility to be able to put `<IP>` on any position in the
evaluated command.  So one can just run the command + <IP>.  But it is possible to do
```bash
eval set -- "$check_command_exp"
local command=()
for arg; do
  if test "$arg" != "<IP>"; then
    command+=( "$arg" )
  else
    command+=( "$ip" )
  fi
done
"${command[@]}"
```

But that's not an issue;  I wouldn't be against merging this "as-is"; that's a new so please don't feel blocked and (from my POV) feel free to just push to master :-).

test-lib-openshift.sh
pkubatrh: I see. Is this reasoning the same for `ct_os_test_s2i_app_func`? If so then `oc_args` needs to be quoted when the function gets called below.

hhorak: Good point, I'll fix this.

test-lib-openshift.sh
hhorak: @praiskup I'll add the `eval` as suggested, although I did not understand the point about `<IP>` -- with the proposed changes, it looks like it wouldn't be possible to use `<IP>` as part of the string, like in a very simple scenario one might provide a check command just like this:
```
curl http://<IP>/api/check | grep true
```

praiskup: Right, I misunderstood the semantics.  Thanks.

hhorak:
@praiskup @pkubatrh I believe the latest fix should address the issues you mentioned, so if you're not against, I'll merge.. we can tune it more later.

pkubatrh:
looks ok

hhorak:
Thanks for all the suggestions.

PR #47, opened 2017-12-14
test-libs: ignore build/deploy pods in ct_os_get_pod_status

pkubatrh:
Since `ct_os_get_pod_status` only returns the status of the first pod it finds it might return the status of a build/deploy pod instead of the actual pod being deployed (since they all share the same prefix).

@omron93 @hhorak ptal


hhorak:
LGTM.

PR #48, opened 2017-12-14
os: add new functions for working with pods

pkubatrh:


test-lib-openshift.sh
```diff
@@ -60,6 +62,13 @@ function ct_os_get_all_pods_status() {
   oc get pods -o custom-columns=NAME:.metadata.name,Ready:status.containerStatuses[0].ready
 }
 
+# ct_os_get_all_pods_name
+# --------------------
+# Returns the full name of all pods.
+function ct_os_get_all_pods_name() {
```
omron93: Will this function be used separately?

IMPOV, more functions this file has, more hard to read it is...

pkubatrh: I can see it being used for when you want to get names of build/deploy pods, since those are by default ignored by `ct_os_get_pod_name`. But might be too much of a corner case. I am ok with removing the function.

omron93: It makes sense. Thanks for explanation.

test-lib-openshift.sh
```diff
@@ -33,13 +33,15 @@ function ct_get_public_ip() {
...
-# ct_os_run_in_pod POD_PREFIX CMD
+# ct_os_run_in_pod POD_NAME CMD
 # --------------------
 # Runs [cmd] in the pod specified by prefix [pod_prefix].
-# Arguments: pod_prefix - prefix or whole ID of the pod to run the cmd in
+# Arguments: pod_name - full name of the pod
 # Arguments: cmd - command to be run in the pod
 function ct_os_run_in_pod() {
-  : # TODO
+  local pod_name="$1"
```
omron93: Missing `shift`?

hhorak:
LGTM.

PR #50, opened 2017-12-15
os: do not manually increment SECONDS

pkubatrh:
As SECONDS is an internal variable of the shell this is not needed.


hhorak:
LGTM. @omron93 FYI, try `echo $SECONDS` in your terminal, it's kinda magic ;)

omron93:
Thanks:-)

PR #51, opened 2018-01-15
Function for printing logs on ERR

omron93:
Add function for printing logs on ERR

Also in waiting for RC include also statefulsets (it is not the same, but with the same function/purpose - so IMPOV no making two separate function would be less usable. But I can split this function)

@pkubatrh Please take a look.


omron93:
Also use temporary dir for persistent volumes -> without it pod can obtain PV with data from previous run (~ testing lower version)

pkubatrh:
lgtm thanks!

Issue #54, opened 2018-01-25
issue with make syntax error


bobbydeveaux:
##Description

When running make, I get an syntax error in common/build.sh

## Versions
Using OSX 10.13.2
Tried both ZSH & Bash
GNU bash, version 4.4.12(1)-release (x86_64-apple-darwin17.0.0)
GNU Make 4.2.1 Built for x86_64-apple-darwin17.3.0

`$ make build TARGET=rhel7 VERSIONS=7.1`
```make[1]: Entering directory '/Users/bobby/code/s2i-php-container'
VERSION="7.1" SKIP_SQUASH=0 UPDATE_BASE= OS=rhel7 CLEAN_AFTER= DOCKER_BUILD_CONTEXT=. OPENSHIFT_NAMESPACES="5.5" common/build.sh
common/build.sh: line 55: syntax error near unexpected token `{stdout_fd}'
make[1]: *** [common/common.mk:53: 7.1] Error 2
make[1]: Leaving directory '/Users/bobby/code/s2i-php-container'
make: *** [common/common.mk:42: build-serial] Error 2```

`$ make check`
```make[1]: Entering directory '/Users/bobby/code/s2i-php-container'
VERSION="5.6" SKIP_SQUASH=1 UPDATE_BASE= OS=centos7 CLEAN_AFTER= DOCKER_BUILD_CONTEXT=. OPENSHIFT_NAMESPACES="5.5" common/build.sh
common/build.sh: line 55: syntax error near unexpected token `{stdout_fd}'
make[1]: *** [common/common.mk:53: 5.6] Error 2
make[1]: Leaving directory '/Users/bobby/code/s2i-php-container'
make: *** [common/common.mk:42: build-serial] Error 2```

## Hmm

I've tried every syntax checker there is, and it all looks fine - as it probably is given others are using it. But don't know why this is happening?
```
parse_output ()
{
  local command=$1 filter=$2 var=$3 stream=$4
  local raw_output= rc=0
  {
      raw_output=$(_parse_output_inner)
  } {stdout_fd}>&1 {stderr_fd}>&2
  rc=$?
  eval "$var=\$raw_output"
  (exit $rc)
}```

Any clues? :/

bobbydeveaux:
@hhorak - dare suppose you have any idea about this?

praiskup:
> Any clues? :/

Thanks for the report.  I would bet the `build.sh` runs the default `/bin/bash` (shebang), which is the old shell version 3, right?

[...]

PR #55, opened 2018-02-02
makefile: use $SHELL defaulting to bash

praiskup:
Fixes: #54


praiskup:
[test]

praiskup:
Feedback from mac users is insufficient.  @pkubatrh IMO this makes sense to merge anyway, wdyt?

pkubatrh:
Yeah looks ok, thanks

PR #56, opened 2018-02-03
Rewrite the help.1 format test so that it doesn't require file utility

hhorak:
That way we should avoid tests not working on a machine that misses `fileutils` package.


hhorak:
[test]

praiskup:
looks fine

Issue #25, opened 2017-08-24
provide a way to build RHEL images on Fedora box


praiskup:
.. and also on not subscribed RHEL boxes (so far, docker build for RHEL images works
only on successfully subscribed RHEL7 box).

Per @hhorak report, it is possible to (optionally) provide `-v /host.repo:/etc/yum.repos.d/host.repo` option for `docker build .` command.

pkubatrh:
>Per @hhorak report, it is possible to (optionally) provide -v /host.repo:/etc/yum.repos.d/host.repo option for docker build . command.

Unfortunately this does not work as `docker build` makes sure every file is mounted as read-only and the first thing we are doing in every image is disabling all repositories, making the build fail.

praiskup:
> the first thing we are doing in every image is disabling all repositories

Do we still have to do this?

pkubatrh:
>Do we still have to do this?

I think so. Afaik this is present so that we do not install any packages from other repositories that might be enabled on the host.

pkubatrh:
The `yum-config-manager --enable ...` commands would have to be removed as well, since they also try to save the .repo file changes.

pkubatrh:
Looking into this again since we are now able to use `prepare-yum-repositories` script in images built from s2i-core.

@hhorak Is there an easy way to set `SKIP_REPOS_ENABLE` on build? I have only found `docker build --build-args ...` so far but it seems that an additional ARG instruction would need to be added to every Dockerfile to enable that functionality...

pkubatrh:
The other way to use `prepare-yum-repositories` with `SKIP_REPOS_ENABLE` set is to modify the actual Dockerfile before the build:

https://github.com/pkubatrh/container-common-scripts/blob/custom_repo/build.sh#L74

This makes sure that the custom repo file is mounted into the build environment (as read-only) and not touched (no enabling/disabling) for the whole duration of the build.

hhorak:
@pkubatrh So, the `ARG` is not inherited?

pkubatrh:
>@pkubatrh So, the ARG is not inherited?

It does not seem to be. Which is imo the expected behaviour as the ARG instruction (and the --build-args argument) is tied to the build of the Dockerfile.

omron93:
> Is there an easy way to set SKIP_REPOS_ENABLE on build? I have only found docker build --build-args ... so far but it seems that an additional ARG instruction would need to be added to every Dockerfile to enable that functionality...

[...]

PR #57, opened 2018-02-06
build: Make it possible to add custom repo files at build

pkubatrh:
This PR adds a new (optional) Makefile argument `CUSTOM_REPO` through which a path to a rpm .repo file (or a directory containing multiple files) can be provided. The content of `CUSTOM_REPO` will then be mounted into `/etc/yum.repos.d/` during build.

Related: #25


praiskup:
Looks like this should do it, thanks!

Issue #58, opened 2018-02-08
Support centos6 builds


rockdreamer:
IIUC In order to maintain compatibility of c++ software with centos6, it is necessary to build it on centos6.

I've got patches for creating devtoolset images for centos6 but they require having support for that on this project.

Would it be possible to add this support so that adding a Dockerfile.centos6 file creates centos6 versions?

At your disposal for any suggestions/critiques/ideas

rockdreamer:
Thanks for merging! 👍

PR #59, opened 2018-02-08
Allow setting centos6 as target

rockdreamer:
Pull request relative to #58 
This would enable the use case in https://github.com/rockdreamer/devtoolset-container/commit/57e613e573377cb6d3cfbb1b9d434e5757e5433d

Thanks!


praiskup:
Looks fine, just a bit of research - what container is this going to be used with?

PR #60, opened 2018-02-13
Always set 'set -e' explicitly

praiskup:
Using '-e' in shebang is not secure since executing the script
like 'bash /the-script' isn't taking the additional options into
account.

Fixes regression caused by 44fb885e2e6b03


praiskup:
[test]

praiskup:
Meh, sorry about this.  We should fix this rather sooner than later till not many
images depend on this.  Basically failed test case could be overlooked.

praiskup:
[test]

praiskup:
[test]

praiskup:
Merged early without review, to not risk that others will depend on broken
commons.

pkubatrh:
Makes sense, thanks!

PR #64, opened 2018-02-16
Add usage for the shared scripts

pkubatrh:


README.md
```diff
@@ -5,6 +5,65 @@ This repository is aimed to be added as git submodule into particular
...
+
+`OS`  
+OS version you want to build the images for. Currently the scripts are able to build for
+centos (default), centos6, rhel7 and fedora.
+
+`SKIP_SQUASH`  
+When set to 1 the build script will skip the squash phase of the build.
+
+`CUSTOM_REPO`  
+Set this variable to the path to your local .repo files you want top have available inside
```
omron93: to have

README.md
```diff
@@ -5,6 +5,65 @@ This repository is aimed to be added as git submodule into particular
...
+This section explains the usage of the shared scripts in this repository when
+it is used as a submodule in a container source repository.
+
+Once you have the repository set as a submodule include `common.mk` into your
+root Makefile in order to access the default rules used to call shared scripts.
+
+**Default rules:**
+
+`make` or `make build`  
+This rule will build an image without tagging it with any tags after it is built.
```
omron93: Add info about squashing and need to have installed `docker-squash` (https://github.com/goldmann/docker-squash) tool?

praiskup: IMO this is implementation detail ... so I wouldn't insist on that.  RHEL images will get
squashed nad centos not.  Documentation would be fine, but I don't think it belongs to
this paragraph.

README.md
```diff
@@ -5,6 +5,65 @@ This repository is aimed to be added as git submodule into particular
...
+
+`make tag`  
+Use this rule if you want to tag an image after it is built. It will be tagged with
+two tags - name:latest and name:version.
+Depends on `build`
+
+`make test` or `make check`  
+This rule will run the testsuite scripts contained in the container source repositories.
+It expects the test to be available at `$gitroot/$version/test/run`
+Depends on `tag` as some tests might need to have the images tagged (s2i).
```
praiskup: I pretty much dislike this api ... I would much rather allow users to not have the $root/$version/test directory, and allow them to use only $root/test(s) dir.., wdyt?

README.md
```diff
@@ -5,6 +5,65 @@ This repository is aimed to be added as git submodule into particular
...
+`OS`  
+OS version you want to build the images for. Currently the scripts are able to build for
+centos (default), centos6, rhel7 and fedora.
+
+`SKIP_SQUASH`  
+When set to 1 the build script will skip the squash phase of the build.
+
+`CUSTOM_REPO`  
+Set this variable to the path to your local .repo files you want top have available inside
+the image while building. Useful for building rhel-based images on an unsubscribed box.
```
praiskup: Dunno, this might or might not be implemented in the image itself...  Something like
best practice, but it might not work without additional work in the depending project.

praiskup:
Nice work, thanks.

README.md
pkubatrh: No strong opinion about this. We will still need to have the tests inside the git downstream but I can work around this in the tooling.

README.md
pkubatrh: This actually does not need any special modifications in the image, apart from not touching the mounted .repo files (which was the change needed in s2i-core). But I will make a note of it in the text

README.md
pkubatrh: Right, will add a note, thanks

praiskup:
Docs only, merging.  Thank you!

PR #67, opened 2018-02-20
add a way to run multiple test suites

TomasTomecek:
Related: https://github.com/sclorg/s2i-ruby-container/pull/167

Initial version. I'm open to change the workflow & names. Please, advise.

TODO:

 * [x] remove docker invocation
 * [x] add concept of `TEST_SUITES := "test1 test2"` and then `for x in ${TEST_SUITES}; ...`
    * [x] make sure all tests are being run and exit 1 if one of them failed
 * [x] TEST env should support: container, local

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -87,6 +87,11 @@ test: script_env += TEST_MODE=true
 test: tag
 	VERSIONS="$(VERSIONS)" $(script_env) $(test)
 
+.PHONY: test-with-conu
+test-with-conu: script_env += TEST_CONU_MODE=true
+test-with-conu: tag
+	VERSIONS="$(VERSIONS)" $(script_env) $(test)
+
 .PHONY: test-openshift
 test-openshift: script_env += TEST_OPENSHIFT_MODE=true
 test-openshift: tag
--- a/test.sh
+++ b/test.sh
@@ -19,6 +19,31 @@ for dir in ${VERSIONS}; do
     VERSION=$dir test/run
   fi
 
+  if [ -n "${TEST_CONU_MODE}" ]; then
+    if [[ -x test/run-conu ]]; then
+      if [ -n "${CONU_IMAGE}" ]; then
+        echo "-> Running conu tests in a container"
+        docker run \
+          --net=host \
+          -e VERSION="${dir}" \
+          -e IMAGE_NAME \
+          --rm \
+          --security-opt label=disable \
+          -ti \
+          -v /var/run/docker.sock:/var/run/docker.sock \
+          -v ${PWD}/../:/src \
+          -w "/src/${dir}/" \
+          -ti \
+          "${CONU_IMAGE}" \
+          ./test/run-conu
+      else
+        VERSION="${dir}" ./test/run-conu
+      fi
+    else
+      echo "-> conu tests are not present, skipping"
+    fi
+  fi
+
   if [ -n "${TEST_OPENSHIFT_MODE}" ]; then
     if [[ -x test/run-openshift ]]; then
       VERSION=$dir test/run-openshift
--- a/README.md
+++ b/README.md
@@ -38,6 +38,15 @@ Depends on `tag` as some tests might need to have the images tagged (s2i).
 Similar to `make test` but runs testsuite for Openshift, expected to be found at
 `$gitroot/$version/test/run-openshift`
 
+`make test-with-conu`  
+The rule is similar to `make test`. It runs a test suite written using [conu
+library](https://github.com/user-cont/conu). The path to the test script is
+meant to be at `$gitroot/$version/test/run-conu`. By default the test suite is
+being run in the current environment. You can also run the tests in a container
+by defining variable `CONU_IMAGE`. Container images with conu are available in
+[this docker hub repository](docker.io/usercont/conu:0.6.2), a good value for
+the variable is `docker.io/usercont/conu:0.6.2`.
+
 `make clean`  
 Runs scripts that clean-up the working dir. Depends on the `clean-images` rule by default
 and additional clean rules can be provided through the `clean-hook` variable.


omron93:
Would it be possible to add `Dockerfile.tests` also in this repo? To be able to share it across containers...

TomasTomecek:
Found some spare time to work on this. Waiting for tests to pass locally.

praiskup:
@TomasTomecek if you specified `Required-by: s2i-ruby-container#167` tag in commit message (see a068a12e2b736 e.g.) we could test directly against s2i-ruby-container.

[test]

TomasTomecek:
```
13:55:52 + false this must fail the whole testsuite
```
Is it my fault?

(rebased and updated the commit message)

pkubatrh:
That was a red herring. The actual error was this:

```
13:55:52  # FATAL: Can not be --ff merged into origin/master
13:55:52 make: *** [check] Error 1
```

So should be fixed by the rebase.

[test]

pkubatrh:
The ruby PR needs to be rebased as well @TomasTomecek

TomasTomecek:
rebased

pkubatrh:
Thanks!

[test]

pkubatrh:
Tests are failing now because conu tests are being run for all the image versions, but the ruby PR only provides `run-conu` for one version (2.4)

@TomasTomecek Can you add the conu tests for other versions as well?

TomasTomecek:
Updated.

[test]

test.sh
```diff
@@ -15,10 +15,15 @@ for dir in ${VERSIONS}; do
   # Kept also IMAGE_NAME as some tests might still use that.
   export IMAGE_NAME=$(docker inspect -f "{{.Config.Labels.name}}" $IMAGE_ID)
 
-  if [ -n "${TEST_MODE}" ]; then
-    VERSION=$dir test/run
-  fi
+  for test_script in ${TEST_SUITES}; do
+    if [ "${TEST_MODE}" == "container" ]; then
+      docker run --net=host -e VERSION=$dir -e IMAGE_NAME --rm -v /dev:/dev:ro -v /var/lib/docker:/var/lib/docker:ro --security-opt label=disable --cap-add SYS_ADMIN -ti -v /var/run/docker.sock:/var/run/docker.sock -v ${PWD}:/src -w /src docker.io/modularitycontainers/conu $test_script
```
pvalena: Would you mind to split this into multiple lines, so it can be more human-readble?

test.sh
```diff
@@ -15,10 +15,15 @@ for dir in ${VERSIONS}; do
...
+    elif [ "${TEST_MODE}" == "local" ]; then
+      VERSION=$dir $test_script
+    fi
+  done
 
+  # FIXME: deprecate TEST_OPENSHIFT_MODE and use TEST_SUITES instead?
```
pvalena: Yes, this seems like the way to go.

common.mk
```diff
@@ -64,7 +66,7 @@ $(VERSIONS): % : %/root/help.1
 .PHONY: test check
 check: test
 
-test: script_env += TEST_MODE=true
+test: script_env += TEST_MODE=local
```
pvalena: The strange thing is, that this variable now has different meaning...

Previously it was used to switch the test suite `on`, not to define test `variant` (or environment). I'd say the variable chosen before was somehow unfortunate.

TomasTomecek: do I invent a new one then?

test.sh
TomasTomecek: I would do that in dedicated PR since I have no idea how much it has to be done in other repos.

test.sh
TomasTomecek: updated

common.mk
pkubatrh: There is some meaning change behind the variable usage but I do not think it is necessary to introduce a different one just because of that. The variable should only be used in common scripts anyway.

pkubatrh: The issue I see with the way it is used now is that there is no way to set this variable to any other value.

test.sh
```diff
@@ -15,10 +15,28 @@ for dir in ${VERSIONS}; do
...
+        -e IMAGE_NAME \
+        --rm -v /dev:/dev:ro \
+        -v /var/lib/docker:/var/lib/docker:ro \
+        --security-opt label=disable \
+        --cap-add SYS_ADMIN \
+        -ti \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v ${PWD}:/src \
+        -w /src \
+        docker.io/modularitycontainers/conu \
```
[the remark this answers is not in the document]
pkubatrh: The image does not seem to have `pytest-3` though ...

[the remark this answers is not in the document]
TomasTomecek: We moved conu to our github org and I can't figure out how to configure automated build there. Seems github changed how apps/services work.

TomasTomecek:
Unfortunately, I don't have time to finish this one. What would I need to do to land this one?

test.sh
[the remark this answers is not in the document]
pvalena: Lastest image I've found is `dev`, which is 5 months old:
  https://hub.docker.com/r/modularitycontainers/conu/tags/

pvalena:
I think it just needs to pass the tests- the code looks good.

TomasTomecek:
Finally found some spare time to finish this. Please see my changes.

test.sh
```diff
@@ -19,6 +19,28 @@ for dir in ${VERSIONS}; do
...
+        --rm -v /dev:/dev:ro \
+        -v /var/lib/docker:/var/lib/docker:ro \
+        --security-opt label=disable \
+        --cap-add SYS_ADMIN \
+        -ti \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v ${PWD}/../:/src \
+        -w /src/${dir}/ \
+        -ti \
+        docker.io/usercont/conu:0.6.2 \
```
praiskup: Don't you want to have some floating docker tag here instead of 0.6.2?

test.sh
```diff
@@ -19,6 +19,28 @@ for dir in ${VERSIONS}; do
...
+      docker run \
+        --net=host \
+        -e VERSION=$dir \
+        -e IMAGE_NAME \
+        --rm -v /dev:/dev:ro \
+        -v /var/lib/docker:/var/lib/docker:ro \
+        --security-opt label=disable \
+        --cap-add SYS_ADMIN \
+        -ti \
+        -v /var/run/docker.sock:/var/run/docker.sock \
```
praiskup: Hmpfs, this means we give _a lot_ of trust into that container.

praiskup:
The code basically LGTM if that was optional.  But ideally, we should give the end-user chance to install conu library on the host.  If installing conu isn't possible, user could opt-in running privileged container (e.g. by `USE_CONU_CONTAINER=conu:0.6.2`?).

test.sh
TomasTomecek: Actually not: in case we accidentally create a breaking change in a new version, we don't want all your conu tests to start failing. I think it's a better approach to systematically update to newer versions once you're sure they work as you expect. Using latest greatest might be pretty adventurous, especially when the container ecosystem is still changing rapidly.

But your call, really.

test.sh
TomasTomecek: what's the difference in running the tests directly on host?

TomasTomecek:
The library is packaged in Fedora so it's not a problem installing it. One thing to note is that we are planning to remove python2 support (no one ever used it actually): I'm not sure if you are planning to use it on CentOS 7 hosts.

> If installing conu isn't possible, user could opt-in running privileged container

Please advise me the precise steps, maybe something like this?
```
if `python3 -c "import conu"`; then
  ./test/run-conu
else
  $start_that_container
fi
```

test.sh
praiskup: > Using latest greatest might be pretty adventurous

By floating tag I did not mean "latest greatest", but rather "latest stable".  Whether we'll update the version here all the time by PR, or you'll tag it into "conu:stable" tag doesn't make much practical difference - it will hurt us and  PR is only additional work.  Btw.,  conu should simplify our lives so we don't have to take care of ever-changing container API, not that we'll have to take care ever-changing API in conu.

test.sh
praiskup: I hope we didn't run privileged container, nor UID=0 container in testsuite so far in our suite -- except for build -- but I don't know, I thought `docker build` isn't mounting insecure things into the container.  That said, I thought it is banned policy to give a random container write access to a docker socket.  But I'm probably too old school;  /me just dropped the UID from docker group, and I'll run the suite in VM.

omron93:
> as long as we have documented this to be a prerequisite (in README.md, with others)

There are other prerequisites, so this shouldn't be a problem as long as rpm is easily installable in Fedora/RHEL/CentOS! Python3 RHSCL is fine, but it's harder to use it as rpm dependency. Using conu in container might have some problems, but user gets running tests out of the box.

How to install conu in RHEL7/Centos7?

TomasTomecek:
It's packaged in epel7. I can see we haven't updated it in a while: https://src.fedoraproject.org/rpms/conu/blob/epel7/f/conu.spec I don't think it's such a big deal to keep packaging it for epel.

On the other hand, I just realized that python-kubernetes is not available in epel:
https://github.com/user-cont/conu/issues/323

> I simply don't like the unannounced execution of privileged container
Yes, that's a fair point: the container has more permissions than your unpriv user.

I updated the code with your suggestions.

The ruby test suite keeps passing:
```
test/test_s2i.py::TestSuite::test_s2i_apps[/home/tt/g/sclorg/s2i-ruby-container/2.5/test/puma-test-app] PASSED                                                                                              [ 25%]
test/test_s2i.py::TestSuite::test_s2i_apps[/home/tt/g/sclorg/s2i-ruby-container/2.5/test/rack-test-app] PASSED                                                                                              [ 50%]
test/test_s2i.py::TestSuite::test_invoking_container PASSED                                                                                                                                                 [ 75%]
test/test_s2i.py::TestSuite::test_usage PASSED                                                                                                                                                              [100%]
```

praiskup:
Thanks!

test.sh
```diff
@@ -19,6 +19,32 @@ for dir in ${VERSIONS}; do
...
+          --cap-add SYS_ADMIN \
+          -ti \
+          -v /var/run/docker.sock:/var/run/docker.sock \
+          -v ${PWD}/../:/src \
+          -w /src/${dir}/ \
+          -ti \
+          $CONU_IMAGE \
+          ./test/run-conu
+      else
+        VERSION=$dir ./test/run-conu
```
praiskup: One pedantic note, can you please double-quote the shell vars above?

TomasTomecek: Yes, I can.

TomasTomecek:
I just realized that the container had too many privileges for no reason, I stripped that down. PTAL

praiskup:
[test] let's test the Required-by

praiskup:
Pushed as f1539e9e29b5cf1c397d6d8a94dc2e5ffa213279 (fixed $PWD escape).

TomasTomecek:
Thank you!

I'll have a beer to celebrate this.

PR #68, opened 2018-02-26
Allow users passing spaces in ct_create_container as $CONTAINER_ARGS …

ficap:
…(testing passwords)
As we often test whether container does not accept some crazy passwords, we use some strange symbols as parameters (for example a space). This PR enables users to use spaces during testing.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -139,7 +139,7 @@ function ct_assert_container_creation_fails() {
 function ct_create_container() {
   local cid_file="$CID_FILE_DIR/$1" ; shift
   # create container with a cidfile in a directory for cleanup
-  docker run --cidfile="$cid_file" -d ${CONTAINER_ARGS:-} $IMAGE_NAME "$@"
+  docker run --cidfile="$cid_file" -d "${CONTAINER_ARGS:-}" $IMAGE_NAME "$@"
   ct_wait_for_cid $cid_file || return 1
   : "Created container $(cat $cid_file)"
 }


omron93:
Is this working for you?

After this change whole CONTAINER_ARGS is passed as one string to docker run and docker can't parse it. (maybe I've tested it wrong)

test-lib.sh
```diff
@@ -139,7 +139,7 @@ function ct_assert_container_creation_fails() {
 function ct_create_container() {
   local cid_file="$CID_FILE_DIR/$1" ; shift
   # create container with a cidfile in a directory for cleanup
-  docker run --cidfile="$cid_file" -d ${CONTAINER_ARGS:-} $IMAGE_NAME "$@"
+  docker run --cidfile="$cid_file" -d "${CONTAINER_ARGS:-}" $IMAGE_NAME "$@"
```
praiskup: This doesn't feel right.  Spaces in CONTAINER_ARGS are to separate options.

This is a "known" weak point though, we should have there properly shell-escaped options so the "docker run" can be executed through eval.  That way you can have multiple args and some of them can container spaces.

ficap:
Yes, you're right, my fault. It really does not work. I tested it wrong way. I'll investigate the way which @praiskup proposed.

praiskup:
Closing this, there's quite some work needed to get this done - without any progress so far. Filip is not in the team anymore, so it is waiting for someone else.

PR #70, opened 2018-03-05
test-lib: add ct_random_string function

praiskup:

PR #71, opened 2018-03-05
Add ct_get_random_str function

ficap:
For example for generating random id suffixes.
I think there's no need for passing argument length for adjusting string length?


pkubatrh:
@praiskup is already working on a similar PR. Also trying to fix some issues with the current (cat based) implementation. So closing this one, sorry.

https://github.com/sclorg/container-common-scripts/pull/70

PR #73, opened 2018-03-19
openshift: pull always the :latest tag

praiskup:
We tag each new release of image under :latest, so it's better
to not rely on what _local docker thinks_ is the "latest" tag.

Related to sclorg/postgresql-container/255.


pkubatrh:
Makes sense, thanks. Merging

PR #77, opened 2018-04-04
Add option to specify cluster version

hhorak:
In order to allow parametrize tests, we need to allow to choose version of the cluster (server) to be used, no matter what version the `oc` tool is. That is possible by setting a variable `OC_CLUSTER_VERSION` (parameter of the function `ct_os_cluster_up` may be also used), which results in parameter `--version=vX.Y` of the `oc cluster up` call.

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -287,18 +287,23 @@ function ct_os_install_in_centos() {
                  bash-completion origin-clients docker origin-clients
 }
 
-# ct_os_cluster_up [DIR, IS_PUBLIC]
+# ct_os_cluster_up [DIR, IS_PUBLIC, CLUSTER_VERSION]
 # --------------------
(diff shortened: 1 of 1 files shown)


test-lib-openshift.sh
```diff
@@ -294,11 +294,16 @@ function ct_os_install_in_centos() {
 # Arguments: is_public - sets either private or public hostname for web-UI,
 #                        use "true" for allow remote access to the web-UI,
 #                        "false" is default
+# Arguments: cluster_version - version of the OpenShift cluster to use, empty
```
praiskup: The new argument is missing in `ct_os_cluster_up [DIR, IS_PUBLIC]` header.

praiskup:
small nit, otherwise lgtm, feel free to push directly

praiskup:
Well, it is preferred to have the comment also in the git commit message (not only on github page).

hhorak:
Thanks for the feedback, fixed as suggested.

PR #78, opened 2018-04-06
Add script to generate sources into generated branch

pkubatrh:
Created as a result of the discussion in https://github.com/sclorg/postgresql-container/pull/268

### DIFF
--- a/update-generated.sh
+++ b/update-generated.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+set -ex
+
+SOURCE_BRANCH=${1:-master}
+GENERATED_BRANCH=${2:-generated}
+
+git clean -f -d
+
+# checkout source branch
+git checkout "$SOURCE_BRANCH" && git submodule update
+tag=$(git rev-parse HEAD)
+
+# copy the branch content into source/
+rsync -a ./* source
+VERSIONS=$(sed -n 's/^VERSIONS[[:space:]]*=//p' source/Makefile)
+
+git checkout "$GENERATED_BRANCH" && git submodule update
+for i in *; do
+    case "$i" in
+    # Do not remove the update script or common submodule
+    update|common|source)
+        continue
+        ;;
+    *)
+        rm -rf "$i"
+        ;;
+    esac
+done
+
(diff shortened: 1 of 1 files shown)


pkubatrh:
@praiskup @omron93 ptal

pkubatrh:
Made some more minor changes. Ready for review now.

update-generated.sh
```diff
@@ -0,0 +1,55 @@
...
+# checkout source branch
+git checkout "$SOURCE_BRANCH" && git submodule update
+tag=$(git rev-parse HEAD)
+
+# copy the branch content into source/
+rsync -a ./* source
+VERSIONS=$(sed -n 's/^VERSIONS[[:space:]]*=//p' source/Makefile)
+
+git checkout "$GENERATED_BRANCH" && git submodule update
+for i in *; do
```
omron93: What about

```bash
shopt -s extglob
rm -rf !(common|sources)
```
?

update-generated.sh
```diff
@@ -0,0 +1,55 @@
...
+    case "$i" in
+    # Do not remove the update script or common submodule
+    update|common|source)
+        continue
+        ;;
+    *)
+        rm -rf "$i"
```
omron93: I'm thinking... Why is needed to delete files in generated branch? - `rsync` could be used to update code by new code from `sources`.

 Without deleting files, `Add deleted files to the index as well` step isn't required.

pkubatrh: iirc we decided (while discussing source generation in general) that we do not want to keep archives of EOL images which is why everything gets removed first before the new generated sources get added.

update-generated.sh
pkubatrh: There is still and update symlink in root that points to common scripts for sake of convenience.

Using extended globbing for this is a good idea though, thanks!

pkubatrh: >There is still an update symlink in root that points to common scripts for sake of convenience.

That is for the postgresql POC I will base from this PR at least.

pkubatrh:
Thanks, merging

PR #81, opened 2018-04-25
cleanup some mess in 'make generate*'

praiskup:
First, I wanted to allow users to replace DG, either by adding
'DG = /some/other/dg' assignment into project's Makefile, or by
'make DG=...' option.  But to make this all work, I had to remove
the recursive make call through 'make -f gen.mk gen' which dropped
this configuration.  So I also removed gen.mk entirely and moved
the important stuff into common.mk, the separation was useless
anyways.

At the same time, I realized that the DG_CONF variable is totally
useless, except for local use in generate.sh, so I removed it
almost from everywhere.

Also, there was no difference between generate and generate-all.
I guess 2c28a31c1782b093 removed the difference, so I now made
clear those two to be poor aliases.


praiskup:
@mcyprian @pkubatrh review would be welcome

praiskup:
[test]

praiskup:
[test]

praiskup:
[test]

praiskup:
[test]

praiskup:
[test]

generate.sh
```diff
@@ -68,26 +74,30 @@ parse_rules() {
...
 
+        case $version$dest$src in
+            *' '*) die "space not allowed in version, dest, src" ;;
+        esac
+
+        target=$version/$dest
+        targets+="\\$nl	$target"
+
+        cat >> auto_targets.mk << EOF
+$nl$target: $src manifest.sh
```
pkubatrh: This should depend on `$MANIFEST_FILE ` instead of `manifest.sh`

generate.sh
```diff
@@ -96,66 +106,58 @@ EOF
...
-    version=${combination##*=}
-    DG_CONF=$(echo $combination | cut -d' ' -f2)
+    # line looks like: --distro rhel-7-x86_64.yaml --multispec-selector version=9.4
+    eval 'set -- $combination'
+    case $4 in
+        version=*) version=${4##*=} ;;
+        *) die "version not found"
+    esac
+    case $2 in
+        *x86_64*) DG_CONF=$2 ;;
```
pkubatrh: I wonder if it would be worth it to future proof this for additional architectures

pkubatrh:
Otherwise I did not see any issues

generate.sh
praiskup: Very valid point, thanks!

praiskup:
[test]

pkubatrh:
[test]

pkubatrh:
Seems like the failures are coming from https://github.com/sclorg/s2i-python-container/issues/269

pkubatrh:
Looks like the clean failures were present in the previous (successful) run of the CI tests as well, so probably a red herring ...

pkubatrh:
```
07:41:23  # ERROR: Tests for postgresql-container failed
```
Due to
```
07:41:23 Error response from daemon: Syntax error - can't find = in "\\ ". Must be of the form: name=value
```

This looks like its connected to the recent distgen update

pkubatrh:
Probably because its trying to test for https://github.com/sclorg/postgresql-container/pull/266 which does not yet have the distgen fixes. I am going to rebase it for Michal to get it out of the way.

pkubatrh:
[test]

pkubatrh:
Tests are green now and no other complaints, so merging this in together with https://github.com/sclorg/postgresql-container/pull/266

PR #85, opened 2018-05-03
clean-hook: switch from var to target

praiskup:
We need to order the rules so the 'clean-hook' happens after
'clean-images', and so we need separate target which may be
undefined by default (instead of variable holding list of targets,
by default empty).


pkubatrh:
Right, makes sense. I keep forgetting make is able to run multiple rules in parallel.

common.mk
```diff
@@ -93,14 +93,13 @@ test-openshift: tag
 tag: build
 	VERSIONS="$(VERSIONS)" $(script_env) $(tag)
 
-.PHONY: clean
-clean: clean-images $(clean-hook)
+.PHONY: clean clean-images clean-versions
+clean: clean-images
+	$(MAKE) clean-hook
```
pkubatrh: Is the sub-make going to inherit all of the defined variables by default?

praiskup: Yes.  I'll rebase once more to add --no-print-directory

praiskup:
ptal

common.mk
```diff
@@ -93,14 +93,13 @@ test-openshift: tag
 tag: build
 	VERSIONS="$(VERSIONS)" $(script_env) $(tag)
 
-.PHONY: clean
-clean: clean-images $(clean-hook)
+.PHONY: clean clean-images clean-versions
```
pkubatrh: Add `clean-hook` to PHONY as well so `make clean` does not fail when no hook is defined

praiskup: argh, yes... it was there and I dropped that.  Thanks.

praiskup: done

PR #86, opened 2018-05-07
Pick oc version by setting PATH

hhorak:
Picking version of OpenShift is not enough to be done by setting `--version` argument of `oc`. Ideally, there should be `oc` version of the same version installed. This change first tries to pick `oc` from `/usr/local...` where different preinstalled versions of `oc` may already be prepared. Then, if no suitable `oc` is not found, a community released `oc` binary is downloaded into `/tmp` from https://github.com/openshift/origin/.

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -316,7 +316,10 @@ function ct_os_cluster_up() {
   local cluster_ip="127.0.0.1"
   [ "${is_public}" == "true" ] && cluster_ip=$(ct_get_public_ip)
 
-  ct_os_set_path_oc "${cluster_version}"
+  if [ -n "${cluster_version}" ] ; then
+    # if $cluster_version is not set, we simply use oc that is available
+    ct_os_set_path_oc "${cluster_version}"
+  fi
 
   mkdir -p ${dir}/{config,data,pv}
   oc cluster up --host-data-dir=${dir}/data --host-config-dir=${dir}/config \
@@ -395,7 +398,7 @@ function ct_os_set_path_oc() {
 function ct_os_get_latest_ver(){
   local vxy="v$1"
   for vz in {3..0} ; do
-    curl -si "https://github.com/openshift/origin/releases/tag/${vxy}.${vz}" | grep -q 'HTTP/1.1 200 OK' && echo "${vxy}.${vz}" && return 0
+    curl -sif "https://github.com/openshift/origin/releases/tag/${vxy}.${vz}" >/dev/null && echo "${vxy}.${vz}" && return 0
   done
   echo "ERROR: version ${vxy} not found in https://github.com/openshift/origin/tags" >&2
   return 1
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -295,7 +295,7 @@ function ct_os_install_in_centos() {
 #                        use "true" for allow remote access to the web-UI,
 #                        "false" is default
 # Arguments: cluster_version - version of the OpenShift cluster to use, empty
-#                              means default version of `oc`; example value: v3.7.0;
+#                              means default version of `oc`; example value: 3.7;
 #                              also can be specified outside by OC_CLUSTER_VERSION
 function ct_os_cluster_up() {
   ct_os_cluster_running && echo "Cluster already running. Nothing is done." && return 0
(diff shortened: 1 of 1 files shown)


test-lib-openshift.sh
```diff
@@ -316,10 +316,11 @@ function ct_os_cluster_up() {
   local cluster_ip="127.0.0.1"
   [ "${is_public}" == "true" ] && cluster_ip=$(ct_get_public_ip)
 
+  ct_os_set_path_oc "${cluster_version}"
```
omron93: What if the `cluster_version` is not set?

test-lib-openshift.sh
```diff
@@ -344,6 +345,89 @@ function ct_os_cluster_running() {
...
 
+# ct_os_set_path_oc OC_VERSION
+# --------------------
+# This is a trick that helps using correct version of the `oc`:
+# The input is version of the openshift in format v3.6.0 etc.
+# If the currently available version of oc is not of this version,
+# it first takes a look into /usr/local/oc-<ver>/bin directory,
+# and if not found there it downloads the community release from github.
+# In the end the PATH variable is changed, so the other tests can still use just 'oc'.
+# Arguments: oc_ver - X.Y part of the version of OSE (e.g. 3.9)
```
praiskup: s/oc_ver/OC_VERSION/

test-lib-openshift.sh
```diff
@@ -344,6 +345,89 @@ function ct_os_cluster_running() {
...
+
+# ct_os_get_latest_ver VERSION_PART_X
+# --------------------
+# Returns full version (vX.Y.Z) from part of the version (X.Y)
+# Arguments: vxy - X.Y part of the version
+# Returns vX.Y.Z variant of the version
+function ct_os_get_latest_ver(){
+  local vxy="v$1"
+  for vz in {3..0} ; do
+    curl -si "https://github.com/openshift/origin/releases/tag/${vxy}.${vz}" | grep -q 'HTTP/1.1 200 OK' && echo "${vxy}.${vz}" && return 0
```
praiskup: `curl -sif ...` can be used instead of `| grep -q ...`.

praiskup: Is `{3..0}` range large enough (for the future)?  It seems like yes looking at the release numbers released so far.

test-lib-openshift.sh
```diff
@@ -344,6 +345,89 @@ function ct_os_cluster_running() {
...
+
+  mkdir -p "${output_dir}"
+  # using html output instead of https://api.github.com/repos/openshift/origin/releases/tags/${oc_ver},
+  # because API is limited for number of queries if not authenticated
+  tarball=$(curl -si "https://github.com/openshift/origin/releases/tag/${oc_ver}" | grep -o -e "openshift-origin-client-tools-${oc_ver}-[a-f0-9]*-linux-64bit.tar.gz" | head -n 1)
+
+  # download, unpack the binaries and then put them into output directory
+  echo "Downloading https://github.com/openshift/origin/releases/download/${oc_ver}/${tarball} into ${output_dir}/" >&2
+  curl -sL https://github.com/openshift/origin/releases/download/${oc_ver}/${tarball} | tar -C "${output_dir}" -xz
+  mv -f "${output_dir}/${tarball%.tar.gz}"/* ${output_dir}/
```
praiskup: I would quote the $tarball (result of parsing) and rather also the ${output_dir}/.

test-lib-openshift.sh
hhorak: Good point, we need to have some version every-time, but rather than hard-code e.g. latest released, I'd expected we should simply use what we have installed. What do you think?

test-lib-openshift.sh
hhorak: Not sure I understand, you mean quoting all separately?
`mv -f "${output_dir}"/"${tarball%.tar.gz}"/* "${output_dir}/"` ?

test-lib-openshift.sh
hhorak: So far yes, the largest .z part is 3.7.2, and I wanted to keep it small so we don't do too many requests. Anyway, this way of finding out latest version is really creepy, if anybody would have something that would work similarly, I'm welcoming ideas :)

test-lib-openshift.sh
hhorak: I used deliberately `oc_ver`, since that is the local variable that the value is stored in the function, and it's done similarly in other functions. Although I'm not familiar with shell conventions, so maybe we can sync it in all doctexts in the file.

test-lib-openshift.sh
hhorak: `curl -sif ` looks good, will use it.

test-lib-openshift.sh
praiskup: Yes, plus `curl -sL https.../"$tarball"`, content of $tarball is result of parsing so it mistakenly could contain some space..

test-lib-openshift.sh
praiskup: Ok, then s/OC_VERSION/oc_ver/ in header?  It's just inconsistency in function docs.

omron93: Other images in this file use same convention for parameters. Capitals above `# ---------------` and small letters in description.

So we should then fix all descriptions.

praiskup: Well, whether that's oc_ver or OC_VER doesn't matter; more important is `oc_ver != oc_version`.

test-lib-openshift.sh
hhorak: Ok, fixing.

test-lib-openshift.sh
hhorak: ah, right, will use full name oc_version then.

praiskup:
lgtm

PR #89, opened 2018-05-14
update-generated.sh: common/ is not needed in 'generated' branch

praiskup:
Complements: rhscl-container-ci#16


pkubatrh:
This makes it easier to work with the generated branch, thanks!

There is some change in how the script is used (called from source branch by default, instead of from generated) but I think only postgresql and the CI is using it so should be safe to merge.

PR #90, opened 2018-06-04
generate: provide an error message when distgen binary not available

pkubatrh:
Related: https://github.com/sclorg/s2i-python-container/issues/274


praiskup:
thanks

PR #91, opened 2018-06-05
Allow to run with other than local cluster

hhorak:
Do not run local cluster, if already logged into any (remote/local) cluster -- `oc whoami` should work here, to determine whether we're logged in to the cluster already, but maybe there is a better solution.
Also do not login to a docker registry, if the address (now newly configurable) is defined.

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -789,7 +789,7 @@ metadata:
 spec:
   containers:
   - name: command-container
-    image: ${image_name}
+    image: "${image_name}"
     command: ["sleep"]
     args: ["3h"]
   restartPolicy: OnFailure
@@ -798,7 +798,8 @@ EOF
   SECONDS=0
   echo -n "Waiting for command POD ."
   while [ $SECONDS -lt 60 ] ; do
-    ct_os_cmd_image_run "echo 24" &>/dev/null && echo "DONE" && return 0
+    sout="$(ct_os_cmd_image_run 'echo $((11*11))')"
+    grep -q '^121$' <<< "$sout" && echo "DONE" && return 0 || :
     sleep 3
     echo -n "."
   done
@@ -874,7 +875,7 @@ ct_os_test_response_internal() {
 ct_os_get_image_from_pod() {
   local pod_prefix=$1 ; shift
   local pod_name=$(ct_os_get_pod_name $pod_prefix)
-  oc get "po/${pod_name}" -o yaml | grep '^    image: ' | head -n 1 | sed -e 's|^    image: ||'
+  oc get "po/${pod_name}" -o yaml | sed -ne 's/^\s*image:\s*\(.*\)\s*$/\1/ p' | head -1
 }
 
 # ct_os_check_cmd_internal
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -263,9 +263,11 @@ function ct_os_delete_project() {
 # Deletes all objects within the project.
 # Handy when we have one project and want to run more tests.
 function ct_delete_all_objects() {
-  for x in bc builds is dc svc po routes secrets ; do
+  for x in bc builds dc is isimage istag po pv pvc rc routes secrets svc ; do
     oc delete $x --all
   done
+  # for some objects it takes longer to be really deleted, so a dummy sleep
+  # to avoid some races when other test can see not-yet-deleted objects and can fail
   sleep 10
 }
 
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -223,7 +223,7 @@ function _ct_os_get_uniq_project_name() {
   local r
   while true ; do
     r=${RANDOM}
-    mkdir /var/tmp/os-test-${r} &>/dev/null && echo test-${r} && break
+    mkdir /var/tmp/sclorg-test-${r} &>/dev/null && echo sclorg-test-${r} && break
   done
 }
 
@@ -234,6 +234,10 @@ function _ct_os_get_uniq_project_name() {
 # Expects 'os' command that is properly logged in to the OpenShift cluster.
 # Not using mktemp, because we cannot use uppercase characters.
 function ct_os_new_project() {
+  if [ "${CT_SKIP_NEW_PROJECT:-false}" == 'true' ] ; then
+    echo "Creating project skipped."
+    return
+  fi
   local project_name="${1:-$(_ct_os_get_uniq_project_name)}" ; shift || :
   oc new-project ${project_name}
   # let openshift cluster to sync to avoid some race condition errors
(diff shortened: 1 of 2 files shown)


test-lib-openshift.sh
```diff
@@ -249,13 +249,23 @@ function ct_os_delete_project() {
   oc delete project "${project_name}"
 }
 
+# ct_os_docker_logged
+# -------------------
+# Returns 0 if already logged in docker
+# Uses global REGISRTY_ADDRESS environment variable for arbitrary registry address.
+function ct_os_docker_logged() {
+  grep -q -e ${REGISRTY_ADDRESS:-172.30.1.1:5000} ~/.docker/config.json
```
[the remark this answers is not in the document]
hhorak: Sounds fine to me, thanks for the suggestion.

hhorak:
I have some more changes in a queue, still tuning, so, please, do not merge yet.

hhorak:
I believe I finished the changes, so it should be ready for a review. Except additions, there is one incompatible change worth mentioning: since `-p NAME` is not set in all templates, it is not used in `ct_os_test_template_app_func` any more, so templates which need it set, must set it explicitly as a parameter. Typically it means to add `-p NAME=perl-testing` when working with e.g. perl templates, because otherwise the template defines a different name and the functions won't know which pod to work with.

test-lib-openshift.sh
```diff
@@ -245,17 +249,36 @@ function ct_os_new_project() {
...
   local project_name="${1:-$(oc project -q)}" ; shift || :
   oc delete project "${project_name}"
 }
 
+# ct_delete_all_objects
+# -----------------
+# Deletes all objects within the project.
+# Handy when we have one project and want to run more tests.
+function ct_delete_all_objects() {
+  for x in bc builds is dc svc po routes secrets ; do
```
omron93: Only these types? or everything? - see `oc types`

test-lib-openshift.sh
```diff
@@ -460,7 +494,19 @@ function ct_os_test_s2i_app_func() {
 
   ct_os_new_project
   # Create a specific imagestream tag for the image so that oc cannot use anything else
-  ct_os_upload_image "${image_name}" "${image_tagged}"
+  if [ "${CT_SKIP_UPLOAD_IMAGE:-false}" == 'true' ] ; then
```
omron93: Do we have usage for both cases? Wouldn't `oc import-image` fits our current usage in tests too? @pkubatrh Ideas?

test-lib-openshift.sh
```diff
@@ -694,3 +756,156 @@ ct_os_test_image_update() {
...
+  done
+  echo "FAIL"
+  return 1
+}
+
+# ct_os_cmd_image_run CMD [ ARG ... ]
+# --------------------
+# Runs a command CMD inside a special command pod
+# Arguments: cmd - shell command with args to run in a pod
+function ct_os_cmd_image_run() {
```
omron93: What about automatic calling `ct_os_deploy_cmd_image` on first `ct_os_cmd_image_run` invocation?

test-lib-openshift.sh
hhorak: I'm not sure whether import-image works with the local docker -- if it would, then it would make it easier. @pvalena might know maybe, have you tried that?

test-lib-openshift.sh
hhorak: Thanks, will extend this list.

test-lib-openshift.sh
hhorak: `ct_os_deploy_cmd_image` requires an image name to be used, so we would either need to specify image for each `ct_os_cmd_image_run` call, or run a default (python) image for those cases. I don't have a strong opinion about which variant is better, or whether the current solution is enough.. Feel free to express your preference.

test-lib-openshift.sh
pvalena: Current list is:
https://docs.openshift.com/container-platform/3.9/cli_reference/basic_cli_operations.html#object-types

Thanks for the `oc types` tip :).

hhorak: Actually I did not include events, since it produced too many errors that I'm not able to remove them (external cluster). I think it's ok to keep events as a garbage after the test run.

test-lib-openshift.sh
omron93: You are right. Maybe better to require explicit call to deploy image.

praiskup:
I'd skip the deep review, and giving +1.

test-lib-openshift.sh
pvalena: AFAIR It should work, if you already have the image locally. But then you have to setup the connection to `remote registry` in your OpenShift instance and use it for the import.

pvalena: Looking at it more `--from=docker.io/openshift/ruby-20-centos7` should probably work even with the IP of the `host`. But I tried the access to host (via some IP), so it may be blocked internally by some OpenShift routing system.

test-lib-openshift.sh
```diff
@@ -694,3 +758,156 @@ ct_os_test_image_update() {
...
+  echo "command POD not running yet, will start one called command-app"
+  oc create -f - <<EOF
+apiVersion: v1
+kind: Pod
+metadata:
+  name: command-app
+spec:
+  containers:
+  - name: command-container
+    image: ${image_name}
```
pvalena: This should be in quotes if I'm not mistaken.

test-lib-openshift.sh
```diff
@@ -694,3 +758,156 @@ ct_os_test_image_update() {
...
+    command: ["sleep"]
+    args: ["3h"]
+  restartPolicy: OnFailure
+EOF
+
+  SECONDS=0
+  echo -n "Waiting for command POD ."
+  while [ $SECONDS -lt 60 ] ; do
+    ct_os_cmd_image_run "echo 24" &>/dev/null && echo "DONE" && return 0
```
pvalena: How about using capturing the stdout and using `grep -q` to check for an actual command execution.
```
sout="$(bash -c 'echo $((11*11))')";
grep -q '^121$' <<< "$sout" && echo OK
```

test-lib-openshift.sh
```diff
@@ -694,3 +758,156 @@ ct_os_test_image_update() {
...
+}
+
+# ct_os_get_image_from_pod
+# ------------------------
+# Print image identifier from an existing pod to stdout
+# Argument: pod_prefix - prefix or full name of the pod to get image from
+ct_os_get_image_from_pod() {
+  local pod_prefix=$1 ; shift
+  local pod_name=$(ct_os_get_pod_name $pod_prefix)
+  oc get "po/${pod_name}" -o yaml | grep '^    image: ' | head -n 1 | sed -e 's|^    image: ||'
```
pvalena: You can use (IMO) simplified and resilient:

`oc get "po/${pod_name}" -o yaml | sed -ne 's/^\s*image:\s*\(.*\)\s*$/\1/ p' | head -1`

test-lib-openshift.sh
```diff
@@ -694,3 +758,156 @@ ct_os_test_image_update() {
...
+  local check_command_exp=$(echo "$check_command" | sed -e "s/<IP>/$ip/g")
+
+  ct_os_deploy_cmd_image $(ct_os_get_image_from_pod "${util_image_name}" | head -n 1)
+  SECONDS=0
+
+  echo -n "Waiting for ${service_name} service becoming ready ..."
+  while true ; do
+    output=$(ct_os_cmd_image_run "$check_command_exp")
+    ret=$?
+    echo "${output}" | grep -qe "${expected_content_match}" || ret=1
```
pvalena: Checking for correct output of `ct_os_cmd_image_run` for 60 in loop seconds seem quite duplicated (I've counted 3) in this PR- maybe you could use a helper function?

pvalena:
Otherwise LGTM- but I've not tested it.

test-lib-openshift.sh
hhorak: But if you only have the image in the local registry, that is not visible to the remote cluster, this won't work.

test-lib-openshift.sh
hhorak: Thx, will use quotes.

test-lib-openshift.sh
hhorak: Good idea :)

test-lib-openshift.sh
hhorak: Ok, will try.

test-lib-openshift.sh
hhorak: I'd add for some next improvements, it is used differently each time, so I don't consider this a big deal for now.

pvalena: I could probably create a PR for that.

hhorak: Great. If you'd be blocked by this one, feel free to merge it, I didn't want to do it before pto and leave it unfixed in other repos (the `-p NAME=perl-testing` thing mentioned above..

omron93:
@hhorak Please rebase. Otherwise, +1.

hhorak:
Let's do it, rebasing and merging.

PR #92, opened 2018-08-06
Use --base-dir instead of --host-*dir on newer oc tools

hhorak:
Options `--host-*dir` are missing in `oc` version 3.10 and higher.

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -322,8 +322,19 @@ function ct_os_cluster_up() {
   fi
 
   mkdir -p ${dir}/{config,data,pv}
-  oc cluster up --host-data-dir=${dir}/data --host-config-dir=${dir}/config \
-                --host-pv-dir=${dir}/pv --use-existing-config --public-hostname=${cluster_ip}
+  case $(oc version| head -n 1) in
+    "oc v3.1"*)
+      oc cluster up --base-dir=${dir}/data --public-hostname=${cluster_ip}
+      ;;
+    "oc v3."*)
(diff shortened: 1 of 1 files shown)


test-lib-openshift.sh
```diff
@@ -322,8 +322,19 @@ function ct_os_cluster_up() {
   fi
 
   mkdir -p ${dir}/{config,data,pv}
-  oc cluster up --host-data-dir=${dir}/data --host-config-dir=${dir}/config \
-                --host-pv-dir=${dir}/pv --use-existing-config --public-hostname=${cluster_ip}
+  case $(oc version| head -n 1) in
+    "oc v3.1"*)
```
praiskup: This will match `v3.1` as well as `v3.10`.  Is that OK?

test-lib-openshift.sh
```diff
@@ -322,8 +322,19 @@ function ct_os_cluster_up() {
...
+      oc cluster up --base-dir=${dir}/data --public-hostname=${cluster_ip}
```
praiskup: I'd prefer e.g. `--base-dir="$dir/data"` format of quotation, but I guess this is not a new problem, so NP.

praiskup:
lgtm

test-lib-openshift.sh
hhorak: Well, kinda, I did not care about 3.1 since it is dead.. But I guess I can change it to `3.1?.*)` which should match what was supposed to match.

test-lib-openshift.sh
hhorak: good idea, thx

praiskup:
Thanks!

PR #94, opened 2018-09-12
[WIP] Modify common scripts so that they work on both podman and docker

pkubatrh:
Expecting docker -> podman symlink/wrapper

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -106,9 +106,8 @@ function docker_build_with_version {
   fi
 
   parse_output 'docker build $BUILD_OPTIONS -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}"' \
-               "awk '/Successfully built/{print \$NF}'" \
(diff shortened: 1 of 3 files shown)


praiskup:
[test]

pkubatrh:
[test]

omron93:
@pkubatrh About our CI: is podman going to be available in RHEL7 ?

@pkubatrh @praiskup @hhorak and should we use it instead of docker?

praiskup:
@omron93 wrote:
>  should we use it instead of docker?

I think we could test it as soon as it is reasonably possible, but it is not right time to switch yet (my guess is that we'll have to support both in future for some period of time)

praiskup:
LGTM

pkubatrh:
>should we use it instead of docker?

It would be nice to use it for Fedora versions I guess. As soon as our tests are able to work with it.

PR #95, opened 2018-09-13
Tag images into localhost namespace

pkubatrh:
`podman tag` tags images into `localhost/` namespace by default.
We should do this explicitly so that the `IMAGE_NAME` used is the same
even for docker built images.


praiskup:
[test]

hhorak:
LGTM, merging.

PR #98, opened 2018-10-08
generate: do not dereference symlinks in link target

pkubatrh:
Fixes: https://github.com/sclorg/s2i-python-container/issues/272


praiskup:
thanks

PR #106, opened 2019-01-16
Revert "Tag images into localhost namespace"

pkubatrh:
Since it seems that this was a bug instead of a feature and was fixed in podman already (v0.10.1) let's revert the "fix" instead of trying to introduce more workarounds for the docker use case.

podman PR: https://github.com/containers/libpod/pull/1551

This reverts commit 7127096e2af5cb37e8f98fd99045b9b6b76f7545.


praiskup:
thanks

PR #108, opened 2019-02-03
Add support for remote cluster tests

hhorak:
Also share BASE_IMAGE_NAME variable for the test script, so it can be used in tests.


hhorak:
[test]

praiskup:
thanks

praiskup:
Looks like just file removal, and adding one new test which is possible to run against local or remote OpenShift cluster.  If I interpret that correctly, it sounds wise - we should move as many tests as possible into the "common" tests subset...  LGTM.

PR #109, opened 2019-02-25
test-lib: add incremental build feature to s2i build

pkubatrh:
Required-by: s2i-nodejs-container#201

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -456,7 +456,13 @@ ct_s2i_build_as_df()
     local user_id=
     local df_name=
     local tmpdir=
+    local oldstate=
     local incremental=false
+
+    oldstate=$(set +o)
+    # Error out if any part of the build fails
+    set -e
+
     # Use /tmp to not pollute cwd
     tmpdir=$(mktemp -d)
     df_name=$(mktemp -p "$tmpdir" Dockerfile.XXXX)
@@ -535,6 +541,7 @@ EOF
     # Run the build and tag the result
     docker build -f "$df_name" -t "$dst_image" .
     popd
+    eval "$oldstate"
 }
 
 # vim: set tabstop=2:shiftwidth=2:expandtab:
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -456,13 +456,37 @@ ct_s2i_build_as_df()
     local user_id=
     local df_name=
     local tmpdir=
+    local incremental=false
     # Use /tmp to not pollute cwd
     tmpdir=$(mktemp -d)
     df_name=$(mktemp -p "$tmpdir" Dockerfile.XXXX)
     pushd "$tmpdir"
     # Check if the image is available locally and try to pull it if it is not
     docker images "$src_image" &>/dev/null || echo "$s2i_args" | grep -q "pull-policy=never" || docker pull "$src_image"
-    user_id=$(docker inspect -f "{{.Config.User}}" "$src_image")
+    user=$(docker inspect -f "{{.Config.User}}" "$src_image")
+    # Default to root if no user is set by the image
+    user=${user:-0}
+    # run the user through the image in case it is non-numeric or does not exist
+    # NOTE: The '-eq' test is used to check if $user is numeric as it will fail if $user is not an integer
+    if ! [ "$user" -eq "$user" ] 2>/dev/null && ! user_id=$(docker run --rm "$src_image" bash -c "id -u $user 2>/dev/null"); then
+        echo "ERROR: id of user $user not found inside image $src_image."
+        echo "Terminating s2i build."
+        return 1
+    else
+        user_id=${user_id:-$user}
+    fi
+    echo "$s2i_args" | grep -q "\-\-incremental" && incremental=true
+    if $incremental; then
+        inc_tmp=$(mktemp -d --tmpdir incremental.XXXX)
+        setfacl -m "u:$user_id:rwx" "$inc_tmp"
+        # Check if the image exists, build should fail (for testing use case) if it does not
+        docker images "$dst_image" &>/dev/null || (echo "Image $dst_image not found."; false)
+        # Run the original image with a mounted in volume and get the artifacts out of it
+        cmd="if [ -s /usr/libexec/s2i/save-artifacts ]; then /usr/libexec/s2i/save-artifacts > \"$inc_tmp/artifacts.tar\"; else touch \"$inc_tmp/artifacts.tar\"; fi"
+        docker run --rm -ti -v "$inc_tmp:$inc_tmp:Z" "$dst_image" bash -c "$cmd"
+        # Move the created content into the $tmpdir for the build to pick it up
+        mv "$inc_tmp/artifacts.tar" "$tmpdir/"
+    fi
     # Strip file:// from APP_PATH and copy its contents into current context
     mkdir -p "$local_app"
     cp -r "${app_path/file:\/\//}/." "$local_app"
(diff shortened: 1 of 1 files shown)


pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
[test]

test-lib.sh
```diff
@@ -488,7 +501,15 @@ EOF
     fi
     # Filter out env var definitions from $s2i_args and create Dockerfile ENV commands out of them
     echo "$s2i_args" | grep -o -e '\(-e\|--env\)[[:space:]=]\S*=\S*' | sed -e 's/-e /ENV /' -e 's/--env[ =]/ENV /' >>"$df_name"
+    # Add in artifacts if doing an incremental build
+    if [ $incremental == true ]; then
+        echo "COPY artifacts.tar /tmp/" >>"$df_name"
```
omron93: Dockerfile ADD instruction could be used to copy in and extract the .tar archive. But have no idea if podman supports this.

pkubatrh: Thanks for the idea. I expect podman supports this.I tried to stay as close to the multistage Dockerfile from `s2i build --as-dockerfile` [output](https://github.com/sclorg/s2i-nodejs-container/pull/201#issuecomment-462318613) as possible but their use of COPY might have been just because of ADD not being able to work with multistage Dockerfiles.

Let me try it with ADD.

pkubatrh:
[test]

test-lib.sh
```diff
@@ -456,13 +456,26 @@ ct_s2i_build_as_df()
...
     tmpdir=$(mktemp -d)
     df_name=$(mktemp -p "$tmpdir" Dockerfile.XXXX)
     pushd "$tmpdir"
     # Check if the image is available locally and try to pull it if it is not
     docker images "$src_image" &>/dev/null || echo "$s2i_args" | grep -q "pull-policy=never" || docker pull "$src_image"
     user_id=$(docker inspect -f "{{.Config.User}}" "$src_image")
+    echo "$s2i_args" | grep -q "\-\-incremental" && incremental=true
+    if [ $incremental == true ]; then
+        inc_tmp=$(mktemp -d --tmpdir incremental.XXXX)
+        setfacl -m "u:$user_id:rwx" "$inc_tmp"
```
praiskup: I'm just not sure that `$user_id` really contains numerical value, so I'd prefer a check for this.

test-lib.sh
```diff
@@ -456,13 +456,26 @@ ct_s2i_build_as_df()
...
+        # Check if the image exists, build should fail (for testing use case) if it does not
+        docker images "$dst_image" &>/dev/null || (echo "Image $dst_image not found."; false)
+        # Run the original image with a mounted in volume and get the artifacts out of it
+        cmd="if [ -s /usr/libexec/s2i/save-artifacts ]; then /usr/libexec/s2i/save-artifacts > $inc_tmp/artifacts.tar; else touch $inc_tmp/artifacts.tar; fi"
```
praiskup: Shouldn't we rather quote `$inc_tmp`?

test-lib.sh
```diff
@@ -488,6 +501,13 @@ EOF
     fi
     # Filter out env var definitions from $s2i_args and create Dockerfile ENV commands out of them
     echo "$s2i_args" | grep -o -e '\(-e\|--env\)[[:space:]=]\S*=\S*' | sed -e 's/-e /ENV /' -e 's/--env[ =]/ENV /' >>"$df_name"
+    # Add in artifacts if doing an incremental build
+    if [ $incremental == true ]; then
```
praiskup: Since $incremental might only be true/false, much nicer would be `if $incremental; then` variant.

praiskup:
Looks fine in general, thank you.

test-lib.sh
pkubatrh: Will do, thanks!

test-lib.sh
pkubatrh: That is a good point. I tried it out and it turns out the config value is anything you give the `USER ...` command in the Dockerfile which can include username instead of a numerical value. It can also be empty (if no `USER ...` command is used) or a username of a non-existing user.

Will make this more bullet-proof.

pkubatrh:
Thanks for the review, should be fixed now.
[test]

test-lib.sh
```diff
@@ -456,13 +456,35 @@ ct_s2i_build_as_df()
...
     df_name=$(mktemp -p "$tmpdir" Dockerfile.XXXX)
     pushd "$tmpdir"
     # Check if the image is available locally and try to pull it if it is not
     docker images "$src_image" &>/dev/null || echo "$s2i_args" | grep -q "pull-policy=never" || docker pull "$src_image"
-    user_id=$(docker inspect -f "{{.Config.User}}" "$src_image")
+    user=$(docker inspect -f "{{.Config.User}}" "$src_image")
+    # Default to root if no user is set by the image
+    user=${user:-0}
+    # run the user through the image in case it is non-numeric or does not exist
+    user_id=$(docker run --rm "$src_image" bash -c "id -u $user" 2>/dev/null || echo "notfound")
```
praiskup: Except that this happens always, even if that is numeric already (or default == 0).  It is probably OK to do `id -u 26` (gives 26), but we can save one container run if we've "only run through the image  in case it is non-numeric".

praiskup: nit, slightly nicer is `if ! user_id=$(docker run ... 2>/dev/null); then`, you can avoid the "echo notfound" which might be theoretically a real user :-).

kubco2: id -u returns number.

pkubatrh: >nit, slightly nicer is if ! user_id=$(docker run ... 2>/dev/null); then, you can avoid the "echo notfound" which might be theoretically a real user :-).

Would `id -u ...` be able to return `notfound`? But I agree that it looks cleaner so will fix.

>Except that this happens always, even if that is numeric already (or default == 0).

I did not add the check as I thought it might be able to pick up a user that is numeric and not located in the image's `/etc/passwd` but now that I think I am not sure if that can even happen, so will add the check. Thanks!

praiskup: right, `id -u ..` should only return numbers, ... so this is nothing more than a "style"-comment, sorry....

pkubatrh:
[test]

pkubatrh:
Rebased.
[test]

test-lib.sh
```diff
@@ -456,13 +456,34 @@ ct_s2i_build_as_df()
...
     df_name=$(mktemp -p "$tmpdir" Dockerfile.XXXX)
     pushd "$tmpdir"
     # Check if the image is available locally and try to pull it if it is not
     docker images "$src_image" &>/dev/null || echo "$s2i_args" | grep -q "pull-policy=never" || docker pull "$src_image"
-    user_id=$(docker inspect -f "{{.Config.User}}" "$src_image")
+    user=$(docker inspect -f "{{.Config.User}}" "$src_image")
+    # Default to root if no user is set by the image
+    user=${user:-0}
+    # run the user through the image in case it is non-numeric or does not exist
+    if ! [ "$user" -eq "$user" ] 2>/dev/null || ! user_id=$(docker run --rm "$src_image" bash -c "id -u $user 2>/dev/null"); then
```
praiskup: `s/||/&&/` ?  btw. interesting hack with `-eq` :-)

pkubatrh: Ah, yes it should be `&&`. Also missing an `else` for when the condition fails on the `-eq` check to get the already numeric value into `$user_id`. Will also add a note explaining the use of `-eq` in the test.

pkubatrh:
[test]

praiskup:
looks good to me (please consider squash).  I guess the Required-by: s2i-nodejs-container#201 means that this is tested in this PR, right?  What if we added the nodejs permanently to testsuite?

pkubatrh:
Squashed. Im not sure if we want to add nodejs to the testsuite given we will not be supporting the images after nodejs version 8.

[test]

praiskup:
> Im not sure if we want to add nodejs to the testsuite given we will not be supporting the images after nodejs version 8.

ACK, though I'm afraid this is pretty complicated addition - can we have at least an issue for adding a trivial test case for this?  Otherwise feel free to merge if the tests are green.

praiskup:
I'm just not sure this is correct:
```
Testing incremental build
the input device is not a TTY
mv: cannot stat '/tmp/incremental.EPYJ/artifacts.tar': No such file or directory
lstat artifacts.tar: no such file or directory
```

pkubatrh:
>the input device is not a TTY
This was a leftover in the test-lib where `docker run` was called with `--ti`.

I have also enabled `set -e` inside `ct_s2i_build_as_df` in order to fail the whole build if any part of it fails.

>mv: cannot stat '/tmp/incremental.EPYJ/artifacts.tar': No such file or directory

I have no idea how this fails in CI. This looks like the artifacts are not generated inside the image, or the mount is not working properly. Does not happen on my box with either docker or podman

pkubatrh:
Also added more `check_result` calls into the nodejs PR, to check whether each of the two builds failed.

pkubatrh:
[test]

praiskup:
```
npm WARN src No license field.
Testing incremental build
Success!
```
Not very talkative, but lgtm.

praiskup:
thank you!

PR #110, opened 2019-03-07
build.sh: correctly use quoted $BUILD_OPTIONS

praiskup:
Previously we built the (centos only?) image label like this:

  "io.openshift.builder-version": "\"7ab45d0\"",

... which was wrong, but it did not hurt Docker.  Now
buildah/podman complaints:

Error: invalid argument
"io.openshift.builder-version=\"7ab45d0\"" for "--label" flag:
parse error on line 1, column 29: bare " in non-quoted-field

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -108,7 +108,7 @@ function docker_build_with_version {
     fi
   fi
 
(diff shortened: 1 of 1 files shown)


praiskup:
[test]

pkubatrh:
lgtm, thanks!

PR #112, opened 2019-03-25
Add ubi7 & ubi8

sub-mod:
For makefile to work in https://github.com/sclorg/s2i-base-container/pull/184


pkubatrh:
Closing as this is not needed - we will base rhel7 on ubi7 and rhel8 on ubi8.

PR #113, opened 2019-04-03
Add a new tag with date and git hash

torsava:
Because $version almost never changes, we need an extra tag so that we
can keep history of images on Docker hub and elsewhere.

Corresponding PR for the docker CI script: https://github.com/sclorg/rhscl-container-ci/pull/25

### DIFF
--- a/tag.sh
+++ b/tag.sh
@@ -15,10 +15,12 @@ for dir in ${VERSIONS}; do
   IMAGE_ID=$(cat .image-id)
   name=$(docker inspect -f "{{.Config.Labels.name}}" $IMAGE_ID)
   version=$(docker inspect -f "{{.Config.Labels.version}}" $IMAGE_ID)
+  date_and_hash="$(date +%Y%m%d)-$(git rev-parse --short HEAD)"
 
-  echo "-> Tagging image '$IMAGE_ID' as '$name:$version' and '$name:latest'"
+  echo "-> Tagging image '$IMAGE_ID' as '$name:$version' and '$name:latest' and '$name:$date_and_hash'"
   docker tag $IMAGE_ID "$name:$version"
   docker tag $IMAGE_ID "$name:latest"
+  docker tag $IMAGE_ID "$name:$date_and_hash"
 
   for suffix in squashed raw; do
     id_file=.image-id.$suffix


frenzymadness:
Looks very nice! Thanks!

praiskup:
thank you

tag.sh
```diff
@@ -15,10 +15,12 @@ for dir in ${VERSIONS}; do
   IMAGE_ID=$(cat .image-id)
   name=$(docker inspect -f "{{.Config.Labels.name}}" $IMAGE_ID)
   version=$(docker inspect -f "{{.Config.Labels.version}}" $IMAGE_ID)
+  date_and_hash="$(date +%Y%m%d)-$(git rev-parse --short HEAD)"
```
praiskup: I now realized that $(date) might not be what we want here;  shouldn't we rather put there a date from the git commit, so if there are mutiple builds -> all would get the same tag?

pkubatrh: I expect using just the date from the commit could cause some confusion since you can merge a commit that is older than what is in master HEAD.

But I see the issue with the date being there. It does not really affect our automatic pushes to dockerhub (which was the reason for this PR) but for local builds you would end up with multiple dated tags to a single image... Maybe just removing the date altogether and staying with hash only?

praiskup: > It does not really affect our automatic pushes to dockerhub (which was the reason for this PR)

I don't think it is guaranteed that the build is issued only once;  one can hit "rebuild" button in jenkins, and you have two builds of the same image, against the same git hash -- but with two dates.

> Maybe just removing the date altogether and staying with hash only?

I don't know;  just trying to discuss...  I'd use either author date, or commit date tbh.

pkubatrh:
>I don't think it is guaranteed that the build is issued only once; one can hit "rebuild" button in jenkins

True. On the other hand there are also some valid use cases for having the same image tagged under different tags now that I think about it. Since it might not be the same image even if the commit hash did not change - the rpm content inside the image (or any of its base images) could have changed.

tag.sh
pkubatrh: >I don't think it is guaranteed that the build is issued only once; one can hit "rebuild" button in jenkins

True. On the other hand there are also some valid use cases for having the same image tagged under different tags now that I think about it. Since it might not be the same image even if the commit hash did not change - the rpm content inside the image (or any of its base images) could have changed.

praiskup: I assume that it is OK ten.  I noticed that the rhscl-container-ci change for this did not get propagated for some, so I'll try to re-deploy.

torsava: I agree with @praiskup, I think `commit date` would be best, it's always increasing (unlike `author date`). Using the rebuild date would be more confusing. I propose we change it, unless anybody objects?

torsava: https://github.com/sclorg/container-common-scripts/pull/114

PR #114, opened 2019-04-05
Tag: Use commit date instead of current date

torsava:
This will mean the tags will be reproducible, which will be less confusing to the users.

Follows PR #113.

### DIFF
--- a/tag.sh
+++ b/tag.sh
@@ -15,7 +15,8 @@ for dir in ${VERSIONS}; do
   IMAGE_ID=$(cat .image-id)
   name=$(docker inspect -f "{{.Config.Labels.name}}" $IMAGE_ID)
   version=$(docker inspect -f "{{.Config.Labels.version}}" $IMAGE_ID)
-  date_and_hash="$(date +%Y%m%d)-$(git rev-parse --short HEAD)"
+  commit_date=$(git show -s HEAD --format=%cd --date=short | sed 's/-//g')
+  date_and_hash="${commit_date}-$(git rev-parse --short HEAD)"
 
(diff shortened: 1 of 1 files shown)


tag.sh
```diff
@@ -15,7 +15,8 @@ for dir in ${VERSIONS}; do
   IMAGE_ID=$(cat .image-id)
   name=$(docker inspect -f "{{.Config.Labels.name}}" $IMAGE_ID)
   version=$(docker inspect -f "{{.Config.Labels.version}}" $IMAGE_ID)
-  date_and_hash="$(date +%Y%m%d)-$(git rev-parse --short HEAD)"
+  commit_date=$(git show HEAD --format=%cd --date=short | sed 's/-//g')
```
praiskup: to not have the diff in $commit_date, I need to add --quiet (at least git 2.20.1)

praiskup:
I'm +1 personally.

tag.sh
torsava: Ah, I was doing it on a diff-less commit, hence I didn't catch it :)

I can't find a `--quiet` option for `git show`, but it appears `-s` is the appropriate flag. Amending.

praiskup: looks fine now, thanks

torsava:
@praiskup I don't see any opposition, so I'd propose we merge this.

praiskup:
@pkubatrh can we merge this?

pkubatrh:
I am on the fence regarding which approach is better but since you are both in favour I will not stall this. Its probably just a corner case anyway.

Merging, thanks!

PR #115, opened 2019-04-19
Add support for ARCH parameter

zregvart:
Adds support for `ARCH` parameter which is passed to docker build as a `TARGETARCH` argument.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -59,7 +59,8 @@ script_env = \
 	CLEAN_AFTER=$(CLEAN_AFTER)                      \
 	DOCKER_BUILD_CONTEXT=$(DOCKER_BUILD_CONTEXT)    \
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"  \
-	CUSTOM_REPO="$(CUSTOM_REPO)"
+	CUSTOM_REPO="$(CUSTOM_REPO)"                    \
+    ARCH="$(ARCH)"
 
 # TODO: switch to 'build: build-all' once parallel builds are relatively safe
 .PHONY: build build-serial build-all
--- a/build.sh
+++ b/build.sh
@@ -13,6 +13,7 @@ script_dir=$(dirname "$script_name")
 
 OS=${1-$OS}
 VERSION=${2-$VERSION}
+ARCH="${3-$ARCH}"
 
 DOCKERFILE_PATH=""
 
@@ -106,6 +107,9 @@ function docker_build_with_version {
     else
       echo "ERROR: file type not known: $CUSTOM_REPO" >&2
     fi
+  fi
+  if [ ! -z "$ARCH" ]; then
+    BUILD_OPTIONS+=" --build-arg TARGETARCH=$ARCH/"
   fi
 
   parse_output 'docker build '"$BUILD_OPTIONS"' -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}"' \


pkubatrh:
Hi @zregvart , thanks for the PR.

Can you provide more info what this would be useful for? From what I can tell the build argument added by your patch is not used anywhere.

zregvart:
@pkubatrh as it currently stands there is no way to build images for different architectures (say ARM64v8) other than changing the base image, this is passing a parameter so in the next PR like the one we did in https://github.com/Project31/s2i-base-container/pull/1 we can build images for other architectures.

I've written about this to the `sclorg@redhat.com` mailing list, I don't think the e-mail got through there seems to be mandatory moderation setup.

pkubatrh:
Ah right, so the build argument is a means to provide a different namespace for the base image? I do not think this is something that would be useful upstream, sounds more like a change you would use in a fork (like you already have).

@hhorak WDYT? Also ping re: ML moderation issue mentioned above.

praiskup:
[test]

common.mk
```diff
@@ -59,7 +59,8 @@ script_env = \
 	CLEAN_AFTER=$(CLEAN_AFTER)                      \
 	DOCKER_BUILD_CONTEXT=$(DOCKER_BUILD_CONTEXT)    \
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"  \
-	CUSTOM_REPO="$(CUSTOM_REPO)"
+	CUSTOM_REPO="$(CUSTOM_REPO)"                    \
+    ARCH="$(ARCH)"
```
praiskup: Tiny: indent.

hhorak:
> @hhorak WDYT? Also ping re: ML moderation issue mentioned above.

non-member contributions are approved by hand, I've done it now, sorry for delay.

hhorak:
It looks like this PR is motivated by the way how the containers are built -- in our case (OSBS), we can use the same image names and thus even FROM argument for all arguments. I'm not sure how this will work for container built in CentOS Container pipeline, but I agree with @pkubatrh that it does not look very generic (suitable for this upstream).

pkubatrh:
I am wondering... What is the reason you need to use the `$ARCH/` namespace in the `FROM` command? Afaict `docker.io/centos` and `docker.io/arm64v8/centos` are the same image on an aarch64 machine.

zregvart:
@pkubatrh as far as I know there is no way to tell Docker which architecture the image is built for and which image to use from the manifest is dependant of the architecture of the machine the build is running on. So in order to force the build of an image for an architecture different from the architecture of the machine the image is built on we need to point to a specific image.

We can build ARM images on x86 machines [by using QEMU](https://medium.com/@kurt.stam/building-aarch64-arm-containers-on-dockerhub-d2d7c975215c).

pkubatrh:
@zregvart Thanks for the info, did not think about the use case of building the arm images in QEMU. 

I would rather not make any changes to the existing Dockerfiles, especially some that could affect the tooling we are currently using (OSBS, CCCP, CWT, ...). I also vaguely remember there being some issues with build args in regards to the use of build cache last I experimented with them.

How about generating arch-specific Dockerfiles on demand? If you only need the `FROM` image's namespace to change it would be pretty easy to implement and I would be ok with that as long as the default behaviour does not change and the changes are limited to this repository (aside from submodule rebases of course).

Still, I think this is more of a feature that would be better suited for a fork, or maybe for the build tools themselves (docker/podman)?

pkubatrh:
After a quick search a similar use case seems to be already tracked for buildah so you might want to join the discussion:
https://github.com/containers/buildah/issues/1590

zregvart:
Moving this to the image build tooling makes perfect sense to me. This was really a way to get this done at the point when `buildx` wasn't available. I can see how this can complicate maintenance by affecting tooling taking this as input.
I'll close this and watch the space around `buildx` and suport in buildah on the issue you mentioned. Thank you for considering this.

PR #117, opened 2019-07-16
test CI

kubco2:


kubco2:
[test]

pkubatrh:
rhel8 tests failing because of docker_squash use (which should not be used on rhel8). Will have to take a look at the state of https://github.com/containers/python-podman for rhel8 implementation.

For now we could probably just disable squash tests on rhel8?

phracek:
The test fails from the reason
```bash
VERSIONS="1.5" SKIP_SQUASH=1 UPDATE_BASE=1 OS=rhel8 CLEAN_AFTER= DOCKER_BUILD_CONTEXT=. OPENSHIFT_NAMESPACES="" CUSTOM_REPO="" TEST_MODE=true /usr/bin/env bash common/test.sh
common/test.sh: line 19: test/run: No such file or directory
make: *** [common/common.mk:89: test] Error 127
Build step 'Execute shell' marked build as failure
```

In this git repository, there is no file called `test/run` like here https://github.com/sclorg/s2i-python-container/tree/master/test

PR #120, opened 2019-08-16
Add resources overview at the beginning of the cleanup function

hhorak:
We suspect that test failures often happen because of lack of resources, so let's see at the beginning of the cleanup function (that is called at the moment of failure) what resources do we actually have.


test-lib.sh
```diff
@@ -544,4 +545,18 @@ EOF
     eval "$oldstate"
 }
 
+# ct_show_resources
+# ----------------
+# Prints the available resources
+ct_show_resources()
+{
+  echo "Resources info:"
+	echo "Memory:"
```
pkubatrh: Just a nitpick, use spaces instead of tabs here

hhorak: Fixed in the rebased PR.

PR #121, opened 2019-08-22
Make docker build not use the cache

hhorak:
Some tests (like the incremental build in the nodejs image) depend on the fact
that a docker image is built from scratch and examines the build log.
With cache, the build log could contain different values than the build run
without the cache.
In case of the nodejs image, the result was that the incremental build
succeeded only every second time when run on the same machine.
Using no cache for docker build should make the tests more stable.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -539,7 +539,7 @@ EOF
         echo "CMD /usr/libexec/s2i/run" >>"$df_name"
     fi
     # Run the build and tag the result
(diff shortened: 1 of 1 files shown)


pkubatrh:
Thanks for the PR! Makes sense to me to not have cache available specifically for inclemental build testing as we want to have `assemble` logs available (at lest in the case of nodejs).

[test]

pkubatrh:
lgtm, merging

PR #124, opened 2019-09-05
add option to create a registry pullsecret for openshift test

pkubatrh:
this pullsecret can be used to pull images from registries
that need authentication to access (eg. registry.redhat.io)

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -264,6 +264,9 @@ function _ct_os_get_uniq_project_name() {
 # Arguments: project - project name, uses a new random name if omitted
 # Expects 'os' command that is properly logged in to the OpenShift cluster.
 # Not using mktemp, because we cannot use uppercase characters.
+# The OPENSHIFT_CLUSTER_PULLSECRET_PATH environment variable can be set
+# to contain a path to a k8s secret definition which will be used
+# to authenticate to image registries.
 function ct_os_new_project() {
   if [ "${CT_SKIP_NEW_PROJECT:-false}" == 'true' ] ; then
     echo "Creating project skipped."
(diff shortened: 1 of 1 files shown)


pkubatrh:
@hhorak PTAL

pkubatrh:
Actually might still need some work. Let me double check

pkubatrh:
Fixed now.

This is needed for https://github.com/sclorg/redis-container/pull/53 but afaik we do not have any way to trigger openshift tests from common scripts in CI before merging. I verified this locally on a rhel7 VM.

test-lib-openshift.sh
```diff
@@ -273,6 +273,13 @@ function ct_os_new_project() {
   oc new-project ${project_name}
   # let openshift cluster to sync to avoid some race condition errors
   sleep 3
+  if test -n "${OPENSHIFT_CLUSTER_PULLSECRET_PATH:-}" -a -e "${OPENSHIFT_CLUSTER_PULLSECRET_PATH:-}"; then
+    oc create -f "$OPENSHIFT_CLUSTER_PULLSECRET_PATH"
+    # add registry pullsecret to the serviceaccount if provided
+    secret_name=$(grep "name:" "$OPENSHIFT_CLUSTER_PULLSECRET_PATH" | awk '{ print $2 }')
```
hhorak: I'd suggest to use '^\s*name:' to avoid a potential wrong match in the future.

test-lib-openshift.sh
```diff
@@ -273,6 +273,13 @@ function ct_os_new_project() {
   oc new-project ${project_name}
   # let openshift cluster to sync to avoid some race condition errors
   sleep 3
+  if test -n "${OPENSHIFT_CLUSTER_PULLSECRET_PATH:-}" -a -e "${OPENSHIFT_CLUSTER_PULLSECRET_PATH:-}"; then
```
hhorak: I think before this line, there could be a short explanation of what this variable means (as it is not described elsewhere). E.g. `OPENSHIFT_CLUSTER_PULLSECRET_PATH variable contains a path to the k8s secret definition that is later used for pulling images from registries that need authentication to access (eg. registry.redhat.io)`

hhorak:
Two suggestions, otherwise LGTM.

test-lib-openshift.sh
pkubatrh: Makes sense, will fix

test-lib-openshift.sh
pkubatrh: Good point, will add it to the docstring of the function.

pkubatrh:
Force pushed requested changes

hhorak:
Thanks, merging.

PR #125, opened 2019-10-04
test-lib: run ct_s2i_build_as_df code inside a subshell to avoid changing shell options

pkubatrh:
The previous version with 'oldstate=$(set +o)' followed by 'eval "$oldstate"' seems
to have been setting "+e" to the caller of the function.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -456,17 +456,17 @@ ct_s2i_build_as_df()
     local user_id=
     local df_name=
     local tmpdir=
-    local oldstate=
     local incremental=false
 
-    oldstate=$(set +o)
+    # Run the entire thing inside a subshell so that we do not leak shell options outside of the function
+    (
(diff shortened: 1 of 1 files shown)


pkubatrh:
Note, this should actually fail the test case as at least one postgresql test case failure was masked by this.

[test]

test-lib.sh
```diff
@@ -541,7 +541,7 @@ EOF
     # Run the build and tag the result
     docker build -f "$df_name" --no-cache=true -t "$dst_image" .
     popd
```
praiskup: popd is useless now

praiskup: and pushd can be replaced with `cd`

pkubatrh: Fixed now, thanks!

pkubatrh:
[test]

PR #127, opened 2019-10-21
Add internal Red Hat Nexus proxy server

phracek:
This pull request adds to the `test-lib.sh` file internal Red Hat Nexus proxy server to `ct_npm_works` function.
We do not want to connect `npmjs` registry directly, based on the request from npmjs.

Also, s2i-nodejs-container is added to `Makefile` in order to check that `ct_npm_works` properly with the internal Red Hat Nexus proxy server.
This is used ONLY for testing containers.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -221,7 +221,7 @@ function ct_mount_ca_file()
 {
   # mount CA file only if NPM_REGISTRY variable is present.
   local mount_parameter=""
-  if [ -n "$NPM_REGISTRY" -a -f "$(full_ca_file_path)" ]; then
+  if [ -n "$NPM_REGISTRY" ] && [ -f "$(full_ca_file_path)" ]; then
     mount_parameter="-v $(full_ca_file_path):$(full_ca_file_path):ro,Z"
   fi
   echo "$mount_parameter"
@@ -234,7 +234,7 @@ function ct_mount_ca_file()
 function ct_build_s2i_npm_variables()
 {
   npm_variables=""
-  if [ -n "$NPM_REGISTRY" -a -f "$(full_ca_file_path)" ]; then
+  if [ -n "$NPM_REGISTRY" ] && [ -f "$(full_ca_file_path)" ]; then
     npm_variables="-e NPM_MIRROR=$NPM_REGISTRY $(ct_mount_ca_file)"
   fi
   echo "$npm_variables"
@@ -247,7 +247,6 @@ function ct_npm_works() {
   local tmpdir=$(mktemp -d)
   : "  Testing npm in the container image"
   docker run --rm ${IMAGE_NAME} /bin/bash -c "npm --version" >${tmpdir}/version
-
   if [ $? -ne 0 ] ; then
     echo "ERROR: 'npm --version' does not work inside the image ${IMAGE_NAME}." >&2
     return 1
@@ -489,7 +488,6 @@ ct_s2i_build_as_df()
     local df_name=
     local tmpdir=
     local incremental=false
-    local ca_file=false
     local mount_options=""
 
     # Run the entire thing inside a subshell so that we do not leak shell options outside of the function
@@ -553,10 +551,8 @@ EOF
     # Filter out env var definitions from $s2i_args and create Dockerfile ENV commands out of them
     echo "$s2i_args" | grep -o -e '\(-e\|--env\)[[:space:]=]\S*=\S*' | sed -e 's/-e /ENV /' -e 's/--env[ =]/ENV /' >>"$df_name"
     # Check if CA autority is present on host and add it into Dockerfile
-    if [ -f "$(full_ca_file_path)" ]; then
-      echo "RUN cd /etc/pki/ca-trust/source/anchors && update-ca-trust extract" >>"$df_name"
-      ca_file=true
-    fi
+    [ -f "$(full_ca_file_path)" ] && echo "RUN cd /etc/pki/ca-trust/source/anchors && update-ca-trust extract" >>"$df_name"
+
     # Add in artifacts if doing an incremental build
     if $incremental; then
         echo "RUN mkdir /tmp/artifacts" >>"$df_name"
@@ -578,10 +574,8 @@ EOF
         echo "CMD /usr/libexec/s2i/run" >>"$df_name"
     fi
 
-    if $ca_file; then
-        # Check if -v parameter is present in s2i_args and add it into docker build command
-        mount_options=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*' || true)
-    fi
+    # Check if -v parameter is present in s2i_args and add it into docker build command
+    mount_options=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*' || true)
 
     # Run the build and tag the result
     docker build $mount_options -f "$df_name" --no-cache=true -t "$dst_image" .
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -207,39 +207,37 @@ function ct_doc_content_old() {
   : "  Success!"
 }
 
-
+# full_ca_file_path
+# Return string for full path to CA file
+function full_ca_file_path()
+{
+  echo "/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
+}
 # ct_mount_ca_file
 # ------------------
 # Check if /etc/pki/certs/RH-IT-Root-CA.crt file exists
-# return mount string for containers
-ct_mount_ca_file()
+# return mount string for containers or empty string
+function ct_mount_ca_file()
 {
-    local mount_parameter=""
-    if [ -n $NPM_REGISTRY ]; then
-        local ca_full_path="/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
-        if [ -f "$ca_full_path" ]; then
-            mount_parameter="-v $ca_full_path:$ca_full_path:ro,Z"
-        fi
-    fi
-    echo "$mount_parameter"
+  # mount CA file only if NPM_REGISTRY variable is present.
+  local mount_parameter=""
+  if [ -n "$NPM_REGISTRY" -a -f "$(full_ca_file_path)" ]; then
+    mount_parameter="-v $(full_ca_file_path):$(full_ca_file_path):ro,Z"
+  fi
+  echo "$mount_parameter"
 }
 
 # ct_build_s2i_npm_variables URL_TO_NPM_JS_SERVER
-# ------------------------------
+# ------------------------------------------
 # Function returns -e NPM_MIRROR and -v MOUNT_POINT_FOR_CAFILE
-function ct_build_s2i_npm_variables() {
-    local ca_full_path="/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
-    if [ ! -z $NPM_REGISTRY ]; then
-        # Check if CA file exists and update CA authorities in image.
-        if [ -f "$ca_full_path" ]; then
-          mount_option=$(ct_mount_ca_file)
-          echo "-e NPM_MIRROR=$NPM_REGISTRY $mount_option"
-        else
-          echo ""
-        fi
-    else
-        echo ""
-    fi
+# or empty string
+function ct_build_s2i_npm_variables()
+{
+  npm_variables=""
+  if [ -n "$NPM_REGISTRY" -a -f "$(full_ca_file_path)" ]; then
+    npm_variables="-e NPM_MIRROR=$NPM_REGISTRY $(ct_mount_ca_file)"
+  fi
+  echo "$npm_variables"
 }
 
 # ct_npm_works
@@ -492,6 +490,7 @@ ct_s2i_build_as_df()
     local tmpdir=
     local incremental=false
     local ca_file=false
+    local mount_options=""
 
     # Run the entire thing inside a subshell so that we do not leak shell options outside of the function
     (
@@ -554,7 +553,7 @@ EOF
     # Filter out env var definitions from $s2i_args and create Dockerfile ENV commands out of them
     echo "$s2i_args" | grep -o -e '\(-e\|--env\)[[:space:]=]\S*=\S*' | sed -e 's/-e /ENV /' -e 's/--env[ =]/ENV /' >>"$df_name"
     # Check if CA autority is present on host and add it into Dockerfile
-    if [ -f "/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt" ]; then
+    if [ -f "$(full_ca_file_path)" ]; then
       echo "RUN cd /etc/pki/ca-trust/source/anchors && update-ca-trust extract" >>"$df_name"
       ca_file=true
     fi
@@ -578,14 +577,14 @@ EOF
     else
         echo "CMD /usr/libexec/s2i/run" >>"$df_name"
     fi
+
     if $ca_file; then
         # Check if -v parameter is present in s2i_args and add it into docker build command
-        mount_ca_option=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*')
-    else
-        mount_ca_option=""
+        mount_options=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*')
     fi
+
     # Run the build and tag the result
-    docker build $mount_ca_option -f "$df_name" --no-cache=true -t "$dst_image" .
+    docker build $mount_options -f "$df_name" --no-cache=true -t "$dst_image" .
     )
 }
 
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -208,6 +208,40 @@ function ct_doc_content_old() {
 }
 
 
+# ct_mount_ca_file
+# ------------------
+# Check if /etc/pki/certs/RH-IT-Root-CA.crt file exists
+# return mount string for containers
+ct_mount_ca_file()
+{
+    local mount_parameter=""
+    if [ -n $NPM_REGISTRY ]; then
+        local ca_full_path="/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
+        if [ -f "$ca_full_path" ]; then
+            mount_parameter="-v $ca_full_path:$ca_full_path:ro,Z"
+        fi
+    fi
+    echo "$mount_parameter"
+}
+
+# ct_build_s2i_npm_variables URL_TO_NPM_JS_SERVER
+# ------------------------------
+# Function returns -e NPM_MIRROR and -v MOUNT_POINT_FOR_CAFILE
+function ct_build_s2i_npm_variables() {
+    local ca_full_path="/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
+    if [ ! -z $NPM_REGISTRY ]; then
+        # Check if CA file exists and update CA authorities in image.
+        if [ -f "$ca_full_path" ]; then
+          mount_option=$(ct_mount_ca_file)
+          echo "-e NPM_MIRROR=$NPM_REGISTRY $mount_option"
+        else
+          echo ""
+        fi
+    else
+        echo ""
+    fi
+}
+
 # ct_npm_works
 # --------------------
 # Checks existance of the npm tool and runs it.
@@ -221,7 +255,8 @@ function ct_npm_works() {
     return 1
   fi
 
-  docker run --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
+  docker run $(ct_mount_ca_file) --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
+
   if [ $? -ne 0 ] ; then
     echo "ERROR: npm could not install jquery inside the image ${IMAGE_NAME}." >&2
     return 1
@@ -229,7 +264,6 @@ function ct_npm_works() {
 
   : "  Success!"
 }
-
 
 # ct_path_append PATH_VARNAME DIRECTORY
 # -------------------------------------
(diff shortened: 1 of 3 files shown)


phracek:
[test] Let's try with NodeJS container.

pkubatrh:
[test]

pkubatrh:
centos CI cannot access internal adress for obvious reasons.

rhel7/8 seems to be suffering from some SSL issues?
`npm verb stack Error: unable to get local issuer certificate`

phracek:
That's expected. Because of CA file is not present.
There are two approaches:

1. Check failure / success of command `npm install jquery --verbose` if `https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/jquery` is present.
2. Add a command `npm set config cafile <CAfile>` into `npm install jquery` command. But this should be hidden. We do not want to share it worldwide.

Any idea, I am opened.

pkubatrh:
I know this is WIP but please keep the changes for `ct_s2i_build_as_df` in sync with how the actual `s2i` binary would behave.

eg. I see in https://github.com/sclorg/s2i-php-container/pull/269:
```
nexus_registry="-e NPM_MIRROR=https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org"
```

All this should do is add a `ENV` command to the resulting Dockerfile.
If you want to add the CA files into the assembly container just use the (unimpemented) `-v` argument.

phracek:
> I know this is WIP but please keep the changes for `ct_s2i_build_as_df` in sync with how the actual `s2i` binary would behave.

Good point, but as you can see here https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L520 all variables like `-e` and `--env` are converted to ENV.

phracek:
What I am totally missing here is a test for `ct_s2i_build_as_df` I will try to solve it somehow. It would help us in the future.

pkubatrh:
> Good point, but as you can see here https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L520 all variables like `-e` and `--env` are converted to ENV.

Yes, but then there is this part which builds on top of the (specific) `ENV` definition inside the same function:
https://github.com/sclorg/container-common-scripts/pull/127/files#diff-8554c86efec4fd2b76e014efc7ab6102R535-R558

pkubatrh:
Also one useful thing to know (that I am not sure is documented anywhere) is testing against image PRs - you can add the following string to your commit message if you want to have a specific image's PR tested against a common-scripts PR:
```
Required-by: postgresql-container#347
```

Just modify the repository name and PR number accordingly.

phracek:
> Yes, but then there is this part which builds on top of the (specific) `ENV` definition inside the same function:
> https://github.com/sclorg/container-common-scripts/pull/127/files#diff-8554c86efec4fd2b76e014efc7ab6102R535-R558

Well, I will try to describe the steps a bit more, maybe I have missed something important:

- `ct_s2i_build_as_df` function gets in `s2i_args` variable, like `-e NPM_MIRROR=URL_ADDRESS`
- By this code https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L520 all `-e` or `--env` are converted and added into Dockerfile
- If `NPM_MIRROR` is present in Dockerfile then get `hostname` to check if the `npmjs` register is reachable from the host.
- if `register` is not reachable from the host, then delete ENV variable, because it causes failures. 
- if `ca file` is present then COPY them into Dockerfile. It could be mounted into an image, but I have to find out how to do it. I did not go to the detail yet. But it could be a better way.
- if `ca file` is not present then NPM_MIRROR should be removed. This is a bug and I will fix it by next commit.

The remain functionality of `ct_s2i_build_as_df` remains untouched.
Thanks for any points, which I have missed.

phracek:
The advantage of this approach is that only `ct_s2i_build_as_df` function is modified.
If we want to add a functionality like `-v` then we have to change it in each affected container. Like `s2i-{perl,ruby,nodejs,php}-container`

phracek:
[test] for getting a certificate path.

pkubatrh:
>The advantage of this approach is that only ct_s2i_build_as_df function is modified.

Sure, but does it make sense to have this very specific behaviour inside a function that is supposed to emulate `s2i build --as-dockerfile`? Or does `s2i` already do the behaviour you are proposing automagically?

pkubatrh:
What I expect should happen here:

- add a way to define ENV inside s2i Dockerfile (already present)
- add a way to mount in a cert (or a path in general) during the s2i build (to be done)
- add a function that can be called to check if we can/need to mount ^

The rest should imo happen in the images.

phracek:
> What I expect should happen here:
> 
>     * add a way to define ENV inside s2i Dockerfile (already present)
> 
>     * add a way to mount in a cert (or a path in general) during the s2i build (to be done)
> 
>     * add a function that can be called to check if we can/need to mount ^
> 
> 
> The rest should imo happen in the images.

Now I see your point. I will change it as you proposed. THX.

phracek:
[test] Let's test one more round.

phracek:
Now, CA file is mounted during the build process.

phracek:
[test]

test-lib.sh
```diff
@@ -518,6 +533,29 @@ EOF
...
+            # Check if CA file exists and update CA authorities in image.
+            ca_full_path="/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
+            if [ -e "$ca_full_path" ]; then
+                echo "RUN cd /etc/pki/ca-trust/source/anchors && update-ca-trust extract" >>"$df_name"
+            else
+              sed -ie "s/ENV NPM_MIRROR.*//" "$df_name"
+            fi
+        fi
+    fi
+    set -e
```
pkubatrh: This makes more sense as part of `ct_mount_ca_file` now.

test-lib.sh
```diff
@@ -539,7 +577,7 @@ EOF
         echo "CMD /usr/libexec/s2i/run" >>"$df_name"
     fi
     # Run the build and tag the result
-    docker build -f "$df_name" --no-cache=true -t "$dst_image" .
+    docker build $(ct_mount_ca_file) -f "$df_name" --no-cache=true -t "$dst_image" .
```
pkubatrh: Please do not make the volume argument specific for this use case. Let's future proof and parse `$s2i-args` for any volume definitions and pass them to `docker build` here.

test-lib.sh
```diff
@@ -221,10 +235,11 @@ function ct_npm_works() {
...
   fi
 
-  docker run --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
+  docker run $(ct_mount_ca_file) --rm ${IMAGE_NAME} /bin/bash -c "npm --verbose install jquery && test -f node_modules/jquery/src/jquery.js"
+
   if [ $? -ne 0 ] ; then
-    echo "ERROR: npm could not install jquery inside the image ${IMAGE_NAME}." >&2
-    return 1
+        echo "ERROR: npm could not install jquery inside the image ${IMAGE_NAME}." >&2
+        return 1
```
pkubatrh: Random indent change?

test-lib.sh
pkubatrh: Or maybe a different function that could call `ct_mount_ca_file`, but it definitely should not be part of `ct_s2i_build_as_df`

test-lib.sh
phracek: Caused by PyCharm. Will fix it in next PR.

test-lib.sh
```diff
@@ -518,6 +558,10 @@ EOF
     fi
     # Filter out env var definitions from $s2i_args and create Dockerfile ENV commands out of them
     echo "$s2i_args" | grep -o -e '\(-e\|--env\)[[:space:]=]\S*=\S*' | sed -e 's/-e /ENV /' -e 's/--env[ =]/ENV /' >>"$df_name"
+    mount_ca_option=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]/etc/pki/\.*\S*')
+    if [ x"$mount_ca_option" != "x" ]; then
+      echo "RUN cd /etc/pki/ca-trust/source/anchors && update-ca-trust extract" >>"$df_name"
+    fi
```
pkubatrh: Can this be moved to assembly script and run unconditionally?

test-lib.sh
```diff
@@ -518,6 +558,10 @@ EOF
     fi
     # Filter out env var definitions from $s2i_args and create Dockerfile ENV commands out of them
     echo "$s2i_args" | grep -o -e '\(-e\|--env\)[[:space:]=]\S*=\S*' | sed -e 's/-e /ENV /' -e 's/--env[ =]/ENV /' >>"$df_name"
+    mount_ca_option=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]/etc/pki/\.*\S*')
```
pkubatrh: Again, please make it general. This still only looks for volume arguments that contain `/etc/pki/`

test-lib.sh
```diff
@@ -208,6 +208,46 @@ function ct_doc_content_old() {
...
+function ct_build_s2i_npm_variables() {
+    local npm_host=$1
+    local ca_full_path="/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
+    set +e
+    npm_mirror=$(echo $npm_host | cut -d'/' -f3)
+    if [ x"$npm_mirror" != "" ]; then
+        # npm_host is in format
+        # http{s}://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org
+        ping -c 5 $npm_mirror > /dev/null
+        if [ $? -ne 0 ]; then
```
pkubatrh: You can do `if ping -c 5 "$npm_mirror" > /dev/null; then` instead. That way you can even remove the `set +e` above

test-lib.sh
```diff
@@ -208,6 +208,46 @@ function ct_doc_content_old() {
...
+
+# ct_build_s2i_npm_variables URL_TO_NPM_JS_SERVER
+# ------------------------------
+# Function returns -e NPM_MIRROR and -v MOUNT_POINT_FOR_CAFILE
+function ct_build_s2i_npm_variables() {
+    local npm_host=$1
+    local ca_full_path="/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
+    set +e
+    npm_mirror=$(echo $npm_host | cut -d'/' -f3)
+    if [ x"$npm_mirror" != "" ]; then
```
pkubatrh: This will always be true

test-lib.sh
```diff
@@ -208,6 +208,46 @@ function ct_doc_content_old() {
...
+        # npm_host is in format
+        # http{s}://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org
+        ping -c 5 $npm_mirror > /dev/null
+        if [ $? -ne 0 ]; then
+            echo ""
+        else
+            # Check if CA file exists and update CA authorities in image.
+            if [ -e "$ca_full_path" ]; then
```
pkubatrh: `-f` as we expect a file

test-lib.sh
phracek: This is not needed at all. We will have $NPM_REGISTRY variable defined. For CI systems.
Therefore I don't see any reason to have this check.

test-lib.sh
phracek: This is not needed, I will remove it.

test-lib.sh
phracek: This has  not to be run in `assemble` script, because it is run as `USER 1001`

test-lib.sh
```diff
@@ -208,6 +208,40 @@ function ct_doc_content_old() {
...
 
 
+# ct_mount_ca_file
+# ------------------
+# Check if /etc/pki/certs/RH-IT-Root-CA.crt file exists
+# return mount string for containers
+ct_mount_ca_file()
+{
+    local mount_parameter=""
+    if [ -n $NPM_REGISTRY ]; then
```
pkubatrh: Missing quotes

test-lib.sh
```diff
@@ -208,6 +208,40 @@ function ct_doc_content_old() {
...
+    fi
+    echo "$mount_parameter"
+}
+
+# ct_build_s2i_npm_variables URL_TO_NPM_JS_SERVER
+# ------------------------------
+# Function returns -e NPM_MIRROR and -v MOUNT_POINT_FOR_CAFILE
+function ct_build_s2i_npm_variables() {
+    local ca_full_path="/etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt"
+    if [ ! -z $NPM_REGISTRY ]; then
```
pkubatrh: `-n` instead, to have the same check as in `ct_mount_ca_file`. Missing quotes here as well.

test-lib.sh
```diff
@@ -208,6 +208,40 @@ function ct_doc_content_old() {
...
+        # Check if CA file exists and update CA authorities in image.
+        if [ -f "$ca_full_path" ]; then
+          mount_option=$(ct_mount_ca_file)
+          echo "-e NPM_MIRROR=$NPM_REGISTRY $mount_option"
+        else
+          echo ""
+        fi
+    else
+        echo ""
```
pkubatrh: I do not think you need to explicitly return empty strings

test-lib.sh
```diff
@@ -538,8 +578,14 @@ EOF
     else
         echo "CMD /usr/libexec/s2i/run" >>"$df_name"
     fi
+    if $ca_file; then
+        # Check if -v parameter is present in s2i_args and add it into docker build command
+        mount_ca_option=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*')
+    else
+        mount_ca_option=""
+    fi
```
pkubatrh: If `$ca_file` does not exist a volume will not be mounted into the build containers. Why should the volume argument be limited in such a way?

test-lib.sh
```diff
@@ -221,7 +255,8 @@ function ct_npm_works() {
     return 1
   fi
 
-  docker run --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
+  docker run $(ct_mount_ca_file) --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
```
pkubatrh: Missing quotes

pkubatrh:
Re-running tests as postgresql should not be failing now on centos

[test]

test-lib.sh
phracek: Well, because it is relevant only for mounting CA file. But I can rename it to `mount_option` and we can use it always.

pkubatrh: Yeah I think that would be better. The volume parameter for ca_file mount is controlled by `ct_build_s2i_npm_variables` so it would not make much sense to have it in two places anyway.

test-lib.sh
phracek: Good point. `ct_mount_ca_file` is used by `ct_build_s2i_npm_variables` and `ct_npm_works` functions.

phracek: I guess `ct_build_s2i_npm_variables` will be used only for `-e NPM_MIRROR` and `ct_mount_ca_file` will be used only for `-v <path>:<path>:`.

pkubatrh: I see nothing wrong with `ct_build_s2i_npm_variables` returing both `-e NPM_MIRROR` and `-v <path>:<path>:`

evanshortiss:
Hey folks, this looks good. Will be nice to be able to use it in builds. What will the usage be once it lands?

AFAICT from reading changes it will look for the CA file on the host and require the NPM_REGISTRY variable to be passed?

phracek:
> Hey folks, this looks good. Will be nice to be able to use it in builds. What will the usage be once it lands?
> 
> AFAICT from reading changes it will look for the CA file on the host and require the NPM_REGISTRY variable to be passed?


Yes, that's right. It requires the CA file present on the host.
We don't want to connect to the public npm registry, but instead of that connect
to the internal registry.

test-lib.sh
phracek: It has to be w/o quotes, otherwise you see error like
```bash
/usr/bin/docker-current: Error response from daemon: create  /etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt: " /etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intented to pass a host directory, use absolute path.
```

pkubatrh: Any reason to have the space before the file path?

pkubatrh: Oh nvm, I see the issue now. Yeah it would not really work quoted since it is an argument to the docker applicaton, my bad.

phracek:
[test]

evanshortiss:
Thanks for the info. Will be very useful. I did a crude setup for it in a [Dockerfile here](https://github.com/RedHatWorkshops/dayinthelife-streaming/blob/master/projects/module-1-ui-web/Dockerfile) recently, but this will be much better since I'd rather use s2i when I can 👍

test-lib.sh
```diff
@@ -538,8 +577,14 @@ EOF
     else
         echo "CMD /usr/libexec/s2i/run" >>"$df_name"
     fi
+
+    if $ca_file; then
+        # Check if -v parameter is present in s2i_args and add it into docker build command
+        mount_options=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*')
```
phracek: @pkubatrh In case `ca_file` is present and NPM_REGISTRY variable was not specified, then `grep` command failed, because of no option like `-v` in `s2i_args` is present. There are two possibilities from my point of view.
before command grep add instruction `set -x` and afterward `set -e` back again or wait till internal PR is merged.

Any other idea?

pkubatrh: How about adding `|| true` to the grep command? That way the result value will never cause the script to fail (and we do not care about it anyway).

phracek:
[test]

phracek:
Tests are failing on Python container. NodeJS and Postgresql containers tests were successfull
```
[91m[0m---> Installing application source ...
---> Installing dependencies ...
[91mYou are using pip version 7.1.0, however version 19.3.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
[0mCollecting gunicorn (from -r requirements.txt (line 1))
  Downloading https://files.pythonhosted.org/packages/60/0d/3dbda0324f5bf007f3274e5ea09f0f3bcbf0ca01a75b80ff4f1ff9f8ecfd/gunicorn-20.0.0-py2.py3-none-any.whl (77kB)
Requirement already satisfied (use --upgrade to upgrade): setuptools>=3.0 in /opt/app-root/lib/python2.7/site-packages (from gunicorn->-r requirements.txt (line 1))
Installing collected packages: gunicorn
Successfully installed gunicorn-20.0.0
 ---> 9c902947a4a2
Removing intermediate container b818766b6c9d
Step 9/9 : CMD /usr/libexec/s2i/run
 ---> Running in 3b51275567b5
 ---> c422cdb4fff5
Removing intermediate container 3b51275567b5
Successfully built c422cdb4fff5

[1m[INFO] Waiting for application container to start  ...[0m

---> Running application from Python script (app.py) ...
  File "app.py", line 4, in <module>
    from gunicorn.app.base import BaseApplication
  File "/opt/app-root/lib/python2.7/site-packages/gunicorn/app/base.py", line 38
    print("\nError: %s" % str(e), file=sys.stderr)
                                      ^
SyntaxError: invalid syntax
```

pkubatrh:
The issue mentioned by Petr should be fixed via https://github.com/sclorg/s2i-python-container/pull/353

phracek:
[test] The issue in s2i-python-container is fixed and merged.

phracek:
[test]

test-lib.sh
```diff
@@ -538,8 +577,14 @@ EOF
     else
         echo "CMD /usr/libexec/s2i/run" >>"$df_name"
     fi
+
+    if $ca_file; then
```
pkubatrh: I guess this can now be removed and mount options set unconditionally?

test-lib.sh
```diff
@@ -207,6 +207,38 @@ function ct_doc_content_old() {
...
+}
+# ct_mount_ca_file
+# ------------------
+# Check if /etc/pki/certs/RH-IT-Root-CA.crt file exists
+# return mount string for containers or empty string
+function ct_mount_ca_file()
+{
+  # mount CA file only if NPM_REGISTRY variable is present.
+  local mount_parameter=""
+  if [ -n "$NPM_REGISTRY" -a -f "$(full_ca_file_path)" ]; then
```
pkubatrh: Better to use `[ -n "$NPM_REGISTRY" ] && [ -f "$(full_ca_file_path)" ]` here

test-lib.sh
```diff
@@ -207,6 +207,38 @@ function ct_doc_content_old() {
...
+}
+
+# ct_build_s2i_npm_variables URL_TO_NPM_JS_SERVER
+# ------------------------------------------
+# Function returns -e NPM_MIRROR and -v MOUNT_POINT_FOR_CAFILE
+# or empty string
+function ct_build_s2i_npm_variables()
+{
+  npm_variables=""
+  if [ -n "$NPM_REGISTRY" -a -f "$(full_ca_file_path)" ]; then
```
pkubatrh: Same here

test-lib.sh
```diff
@@ -518,6 +552,11 @@ EOF
     fi
     # Filter out env var definitions from $s2i_args and create Dockerfile ENV commands out of them
     echo "$s2i_args" | grep -o -e '\(-e\|--env\)[[:space:]=]\S*=\S*' | sed -e 's/-e /ENV /' -e 's/--env[ =]/ENV /' >>"$df_name"
+    # Check if CA autority is present on host and add it into Dockerfile
+    if [ -f "$(full_ca_file_path)" ]; then
```
pkubatrh: This could then be shortened to:
```
[ -f "$(full_ca_file_path)" ] && echo "RUN cd /etc/pki/ca-trust/source/anchors && update-ca-trust extract" >>"$df_name"
```

pkubatrh:
Otherwise LGTM!

test-lib.sh
phracek: I will try and test it.

pkubatrh:
[test]

PR #128, opened 2019-11-27
Add cid_file from running container

phracek:
In case we want to test if NEXUS registry
if working properly we need to exec npm
from already running container.

`docker run` does not work for some reason
and npm variable are not present in .npmrc /opt/app-root
folder.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -222,7 +222,7 @@ function ct_mount_ca_file()
   # mount CA file only if NPM_REGISTRY variable is present.
   local mount_parameter=""
   if [ -n "$NPM_REGISTRY" ] && [ -f "$(full_ca_file_path)" ]; then
-    mount_parameter="-v $(full_ca_file_path):$(full_ca_file_path):ro,Z"
+    mount_parameter="-v $(full_ca_file_path):$(full_ca_file_path):Z"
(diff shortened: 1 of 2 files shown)


test-lib.sh
```diff
@@ -252,7 +253,7 @@ function ct_npm_works() {
     return 1
   fi
 
-  docker run $(ct_mount_ca_file) --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
+  docker exec $(cat ${cid_file}) /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
```
pkubatrh: Not sure I understand this change. You will need to run the container somewhere before you are able to exec into it.

pkubatrh: After discussing this online and suggesting a different way to create the files needed for npm to use custom registries it hit me that this might still not work because of https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L554 only being run during a s2i build :|

phracek: Right, I have to do it in a different way.

phracek:
Closing it. This pull request does not make sense.

PR #129, opened 2019-11-28
Run container for npm test.

phracek:
This pull request checks if npm works properly with NPM_REGISTRY
we need to start a container and check npm by `docker exec` function.

Required: s2i-nodejs-container#221

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -246,19 +246,36 @@ function ct_build_s2i_npm_variables()
 function ct_npm_works() {
   local tmpdir=$(mktemp -d)
   : "  Testing npm in the container image"
+  cid_file="${tmpdir}/cid"
   docker run --rm ${IMAGE_NAME} /bin/bash -c "npm --version" >${tmpdir}/version
   if [ $? -ne 0 ] ; then
     echo "ERROR: 'npm --version' does not work inside the image ${IMAGE_NAME}." >&2
     return 1
   fi
 
-  docker run $(ct_mount_ca_file) --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
+  docker run -d --user=100001 $(ct_mount_ca_file) --rm --cidfile="$cid_file" ${IMAGE_NAME}-testapp
+
+    # Wait for the container to write it's CID file
+  ct_wait_for_cid "$cid_file" || return 1
+
+  docker exec $(cat "$cid_file") /bin/bash -c "npm --verbose install jquery && test -f node_modules/jquery/src/jquery.js" >${tmpdir}/jquery 2>&1
 
   if [ $? -ne 0 ] ; then
     echo "ERROR: npm could not install jquery inside the image ${IMAGE_NAME}." >&2
     return 1
   fi
 
+  if [ -n "$NPM_REGISTRY" ] && [ -f "$(full_ca_file_path)" ]; then
+    grep -o "$NPM_REGISTRY" ${tmpdir}/jquery > /dev/null
+    if [ $? -ne 0 ]; then
+        echo "ERROR: Internal repository is NOT set. Even it is requested."
+    fi
+  fi
+
+  if [ -f "$cid_file" ]; then
+      docker stop $(cat "$cid_file")
+      rm "$cid_file"
+  fi
   : "  Success!"
 }
 
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -253,7 +253,7 @@ function ct_npm_works() {
     return 1
   fi
 
-  docker run -d --user=100001 $(ct_mount_ca_file) --rm --cidfile="$cid_file" ${IMAGE_NAME}-testapp
+  docker run -d $(ct_mount_ca_file) --rm --cidfile="$cid_file" ${IMAGE_NAME}-testapp
 
     # Wait for the container to write it's CID file
   ct_wait_for_cid "$cid_file" || return 1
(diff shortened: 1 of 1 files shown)


phracek:
https://github.com/sclorg/s2i-nodejs-container/pull/221

hhorak:
[test]

test-lib.sh
```diff
@@ -246,19 +246,41 @@ function ct_build_s2i_npm_variables()
...
+
+  docker exec $(cat ${cid_file}) /bin/bash -c "ls -la && id && pwd && npm config list && npm --verbose install jquery && test -f node_modules/jquery/src/jquery.js" >${tmpdir}/jquery 2>&1
 
   if [ $? -ne 0 ] ; then
     echo "ERROR: npm could not install jquery inside the image ${IMAGE_NAME}." >&2
     return 1
   fi
 
+  if [ -n "$NPM_REGISTRY" ] && [ -f "$(full_ca_file_path)" ]; then
+    grep -o "https://repository.engineering.redhat.com" ${tmpdir}/jquery > /dev/null
```
hhorak: I'd say we should grep for `"$NPM_REGISTRY"` as this address can be different than what we know now.

test-lib.sh
```diff
@@ -246,19 +246,41 @@ function ct_build_s2i_npm_variables()
...
+    if [ $? -ne 0 ]; then
+        echo "ERROR: Internal repository is NOT set. Even it is requested."
+    fi
+  fi
+
+  info "Cleaning up the test application"
```
hhorak: Is cleaning explicitly necessary? I'd guess that generic clean function should clean well even in this case, unless there is a reason this must be handled separately.

phracek: My thinking was to clean it after the test, but it is cleaned always. So this part of test is not needed at all.

phracek:
[test]

phracek:
[test][test-openshift]

pkubatrh:
Hmm seems like CI jobs for rhel8 version are stuck after finishing testing the python image

pkubatrh:
[test]

pkubatrh:
Narrowed it down to something introducing the hang inside the image tests, will revisit tomorrow in more detail.

test-lib.sh
```diff
@@ -246,19 +246,32 @@ function ct_build_s2i_npm_variables()
...
   : "  Testing npm in the container image"
+  cid_file="${tmpdir}/cid"
   docker run --rm ${IMAGE_NAME} /bin/bash -c "npm --version" >${tmpdir}/version
   if [ $? -ne 0 ] ; then
     echo "ERROR: 'npm --version' does not work inside the image ${IMAGE_NAME}." >&2
     return 1
   fi
 
-  docker run $(ct_mount_ca_file) --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
+  docker run --user=100001 $(ct_mount_ca_file) --rm --cidfile=${cid_file} ${IMAGE_NAME}-testapp &
```
pkubatrh: I expect this is the culprit for the hang. You run the container in the background (why not run it detached instead?) but you never kill it.

pkubatrh: Killed the container manually and the job completed

phracek: I have had in the commit also cleaning the container https://github.com/sclorg/container-common-scripts/pull/129#discussion_r352220415 but delete it.

I was inspired by s2i-nodejs-container. 
https://github.com/sclorg/s2i-nodejs-container/blob/master/test/run#L292

But missing parts are:
* https://github.com/sclorg/s2i-nodejs-container/blob/master/test/run#L92
* https://github.com/sclorg/s2i-nodejs-container/blob/master/test/run#L97

My thoughts were, that once the job is finished, then the container is automatically destroyed.

Running container in detached mode can be another way.

test-lib.sh
```diff
@@ -246,19 +246,36 @@ function ct_build_s2i_npm_variables()
...
   : "  Testing npm in the container image"
+  cid_file="${tmpdir}/cid"
   docker run --rm ${IMAGE_NAME} /bin/bash -c "npm --version" >${tmpdir}/version
   if [ $? -ne 0 ] ; then
     echo "ERROR: 'npm --version' does not work inside the image ${IMAGE_NAME}." >&2
     return 1
   fi
 
-  docker run $(ct_mount_ca_file) --rm ${IMAGE_NAME} /bin/bash -c "npm install jquery && test -f node_modules/jquery/src/jquery.js"
+  docker run -d --user=100001 $(ct_mount_ca_file) --rm --cidfile="$cid_file" ${IMAGE_NAME}-testapp
```
pkubatrh: Why `--user=100001`. Should we not just use the default user for the image?

test-lib.sh
```diff
@@ -246,19 +246,36 @@ function ct_build_s2i_npm_variables()
...
+  docker exec $(cat "$cid_file") /bin/bash -c "npm --verbose install jquery && test -f node_modules/jquery/src/jquery.js" >${tmpdir}/jquery 2>&1
 
   if [ $? -ne 0 ] ; then
     echo "ERROR: npm could not install jquery inside the image ${IMAGE_NAME}." >&2
     return 1
   fi
 
+  if [ -n "$NPM_REGISTRY" ] && [ -f "$(full_ca_file_path)" ]; then
+    grep -o "$NPM_REGISTRY" ${tmpdir}/jquery > /dev/null
+    if [ $? -ne 0 ]; then
```
pkubatrh: How about:
```
if grep -qo "$NPM_REGISTRY" "${tmpdir}/jquery"
```

pkubatrh:
Looks good, tests are green now and complete. Just some small details left before we can merge

phracek:
[test]

phracek:
[test-openshift]

phracek:
Now, If NPM_REGISTRY variable is present and CA file is present as well, then if NPM_REGISTRY entry is not mentioned in `jquery` log file, then test fails and return 1.

phracek:
In `rhel7` test there is something weird. From log https://gist.github.com/rhscl-bot/b333e062505750485ec2f2b3bf55dfef#file-container-common-scripts-135-sh-L1211 I see, that centos7 image is used instead of rhel7. NPM_REGISTRY should not be present for centos7.

I will look at it deeply.

phracek:
[test]

phracek:
[test]

pkubatrh:
LGTM now, thanks!

Issue #42, opened 2017-11-30
add 'make syntaxcheck'


praiskup:
We could employ ShellCheck or some other static analysis tools (or even plain sed calls...)

Moved from sclorg/postgresql-container#210.

phracek:
This is already solved. `make shellcheck` does the magic, right?

phracek:
Feel free to reopen it.

PR #130, opened 2019-11-29
Add shellcheck script and fix many shell issues found by shellcheck

hhorak:
This adds a new make command `make shellcheck` that is supposed to be run by CI (travis-ci might be used). It only checks the scripts *.sh so far (no tests/*) and the idea is to do similarly over time in other containers, to avoid issues coming from poor shell scripting.

The idea is to specify `SHELLCHECK_FILES` in the `Makefile` for each image to scan only files we want. By default it scans the whole repository, which might be too much.

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -299,6 +299,7 @@ function ct_os_new_project() {
 # --------------------
 # Deletes the specified project in the openshfit
 # Arguments: project - project name, uses the current project if omitted
+# shellcheck disable=SC2120
 function ct_os_delete_project() {
   if [ "${CT_SKIP_NEW_PROJECT:-false}" == 'true' ] ; then
     echo "Deleting project skipped, cleaning objects only."
@@ -317,7 +318,7 @@ function ct_os_delete_project() {
 # Handy when we have one project and want to run more tests.
 function ct_delete_all_objects() {
   for x in bc builds dc is isimage istag po pv pvc rc routes secrets svc ; do
-    oc delete $x --all
+    oc delete "$x" --all
   done
   # for some objects it takes longer to be really deleted, so a dummy sleep
   # to avoid some races when other test can see not-yet-deleted objects and can fail
@@ -847,6 +848,7 @@ ct_os_test_image_update() {
   local ip="" check_command_exp=""
 
   echo "Running image update test for: $image_name"
+  # shellcheck disable=SC2119
   ct_os_new_project
 
   # Get current image from repository and create an imagestream
@@ -871,6 +873,7 @@ ct_os_test_image_update() {
   check_command_exp=${check_function//<IP>/$ip}
   ct_assert_cmd_success "$check_command_exp"
 
+  # shellcheck disable=SC2119
   ct_os_delete_project
 }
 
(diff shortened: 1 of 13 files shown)


hhorak:
[test]

phracek:
Failing test on rhel8 should be fixed by https://github.com/sclorg/container-common-scripts/pull/129

praiskup:
Aaaawesome!

praiskup:
This should close https://github.com/sclorg/container-common-scripts/issues/42

hhorak:
Notes from the meeting: this should be divided to more commits (maybe based on functions).

hhorak:
So, the PR is divided to more commits for easier review. HTH.

test-lib.sh
```diff
@@ -307,8 +307,10 @@ ct_path_foreach ()
 function ct_run_test_list() {
   for test_case in $TEST_LIST; do
     : "Running test $test_case"
-    [ -f test/$test_case ] && source test/$test_case
-    [ -f ../test/$test_case ] && source ../test/$test_case
+    # shellcheck source=/dev/null
+    [ -f test/"$test_case" ] && source test/"$test_case"
```
phracek:
```suggestion
    [ -f "test/$test_case" ] && source "test/$test_case"
```

test-lib.sh
```diff
@@ -307,8 +307,10 @@ ct_path_foreach ()
...
+    # shellcheck source=/dev/null
+    [ -f ../test/"$test_case" ] && source ../test/"$test_case"
```
phracek:
```suggestion
    [ -f "../test/$test_case" ] && source "../test/$test_case"
```

test-lib-openshift.sh
```diff
@@ -740,17 +751,24 @@ function ct_os_test_template_app_func() {
   # get the template file from remote or local location; if not found, it is
   # considered an internal template name, like 'mysql', so use the name
   # explicitly
-  local local_template=$(ct_obtain_input "${template}" 2>/dev/null || echo "--template=${template}")
+  local local_template
   local namespace=${CT_NAMESPACE:-$(oc project -q)}
```
phracek:
```suggestion
  local namespace
  
  namespace=${CT_NAMESPACE:-$(oc project -q)}
```

phracek:
I have reviewed all changes most of them are `local` usage, `double-quotes`, using `$? -eq 0`.
or not used `cat`.

Please rebase it into the latest changes.
`test-lib.sh` has conflict. We can run again test.

We already have nightly build tests and therefore we can catch potential problems soon.

hhorak:
Thanks a lot for the review. The PR is now rebased and proposed changes committed, so should be ready for merge. We can start updating the common sub-module in other containers once code from github is taken for the next containers respin. And ideally fix the remaining shellcheck issues in the particular containers with that.

hhorak:
@phracek Keeping it open for re-check that the requested changes are done.

test-lib.sh
praiskup: Suggestion: put the project-preferences to guidelines document.

In this case, both variants are entirely correct, and I for one usually
prefer the former variant.

hhorak:
Thanks, let's give it a try :)

PR #131, opened 2019-12-09
Remove ro option from mount_parameter

phracek:
We need to remove `ro` parameter from mount command.
It fails on RHEL-8


phracek:
[test]

pkubatrh:
lgtm, merging. Thanks!

PR #134, opened 2020-02-06
Use local var for cidfile

pkubatrh:
So that we do not replace some value that is still in-use in caller
Required-by: s2i-ruby-container#248


pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
Hmm it seems like the `Required-by` did not run any ruby tests even though it got picked up properly:
```
PR commits ask for testing s2i-ruby-container from PR 248
```

pkubatrh:
@phracek PTAL

PR #136, opened 2020-02-21
do not ignore import_image when uploading images

pkubatrh:


pkubatrh:
Note that this will break some existing test cases (eg. https://github.com/sclorg/s2i-ruby-container/blob/master/test/run-openshift-local-cluster#L24) as the `import_image` argument was previously ignored completely but some function calls passed it on anyway

phracek:
I will double check s2i-ruby-container with `import_image` and rewrite the test if it is needed.

PR #139, opened 2020-03-03
test-lib: add binary availability in Dockerfile check

pkubatrh:
Required-by: s2i-nodejs-container#230

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -301,10 +301,10 @@ function ct_binary_found_from_df() {
   # Create Dockerfile that looks for the binary
   cat <<EOF >"$tmpdir/Dockerfile"
 FROM $IMAGE_NAME
-RUN which $binary
+RUN which $binary | grep "$binary_path"
 EOF
   # Build an image, looking for expected path in the output
-  if ! docker build -f "$tmpdir/Dockerfile" --no-cache "$tmpdir" 2>&1 | grep "$binary_path"; then
+  if ! docker build -f "$tmpdir/Dockerfile" --no-cache "$tmpdir"; then
     echo "  ERROR: Failed to find $binary in Dockerfile!" >&2
     return 1
   fi
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -285,6 +285,32 @@ function ct_npm_works() {
   : "  Success!"
 }
 
+# ct_binary_found_from_df binary [path]
+# --------------------
+# Checks if a binary can be found in PATH during Dockerfile build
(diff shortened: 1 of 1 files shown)


pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
rhel CI failing because of:
```
stderr: fatal: unable to connect to github.com:
github.com[0: 192.30.253.113]: errno=Connection timed out
```

pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
nodejs PR needed rebasing for centos fixes, one more [test]

pkubatrh:
Made the test more generic so we can now check any binary not only npm.

[test]

pkubatrh:
[test]

test-lib.sh
```diff
@@ -285,6 +285,32 @@ function ct_npm_works() {
...
+  tmpdir=$(mktemp -d)
+  : "  Testing $binary in build from Dockerfile"
+
+  # Create Dockerfile that looks for the binary
+  cat <<EOF >"$tmpdir/Dockerfile"
+FROM $IMAGE_NAME
+RUN which $binary
+EOF
+  # Build an image, looking for expected path in the output
+  if ! docker build -f "$tmpdir/Dockerfile" --no-cache "$tmpdir" 2>&1 | grep "$binary_path"; then
```
phracek, line 307: Is it enough to check only `/opt/rh`? What about to add as a parameter also `version` and check `full path` like was before in `grep -E`. To be more precise.

pkubatrh: You can already do that via the second argument, `/opt/rh` is only the default.

phracek: What about `grep "^$binary_path"`?

phracek: I am still thinking if it makes sense to somehow automatically "calculate" full binary path for a binary but the effort seems to be really high then expectation. Let's leave it to `^$binary_path'.

pkubatrh: I would leave that up to the user of the function. I have added the `^` to the default value though. Thanks for the idea!

pkubatrh:
[test]

test-lib.sh
phracek: Thanks. LGTM.

pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
Thanks for review, tests are green now so merging

PR #141, opened 2020-03-09
test-lib-openshift: use pullthrough when importing images

pkubatrh:
This way we can workaround the issue of the builder pod not being
able to access registry.redhat.io


pkubatrh:
@phracek PTAL

phracek:
LGTM, But I would be really appropriated to have a comment why `--reference-policy=local`

pkubatrh:
[test]

pkubatrh:
[test]

PR #142, opened 2020-03-22
Add a new funciton ct_wait_container_port

hhorak:
This function that can be used for starting up a container, instead of sleep calls.

Using it can hopefully help to get rid of random test failures seen at https://github.com/sclorg/httpd-container/pull/92.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -650,4 +650,40 @@ ct_check_image_availability() {
   fi
 }
 
+# ct_wait_container_port CONTAINER_ID PORT [ VERBOSE TIMEOUT ]
+# ----------------------------
+# Uses 'nc' for checking whether the container specified by CONTAINER_ID is still
+# running and exits with 0 once the PORT of that container listens for connections.
+# Argument: CONTAINER_ID - id of the container to query the status and ip
+# Argument: PORT - what port to look at
+# Argument: VERBOSE(optional) - if set to 1, progress output will be printed
+# Argument: TIMEOUT(optional) - how long to wait for a port
+ct_wait_container_port() {
+  local container_id=$1
+  local port=$2
+  local verbose=${3:-0}
+  local timeout=${4:-60}
+  local cip
+  local status
+
+  cip=$(docker inspect --format='{{.NetworkSettings.IPAddress}}' "${container_id}")
+  SECONDS=0
+  [ "$verbose" == "1" ] && echo -n "Checking whether port $cip:$port is listening ..." >&2
+  while [ "$SECONDS" -lt "$timeout" ] ; do
+    status=$(docker inspect -f '{{.State.Status}}' "${container_id}")
+    if [ "${status}" != 'running' ] ; then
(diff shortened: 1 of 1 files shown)


praiskup:
At least this isn't enough in case of PostgreSQL ... there's pretty sophisticated method of
checking whether postgresql booted fine (see the liveness/readiness check in PG container).

hhorak:
Right, for db containers, we should probably use more specific method. This might be useful for webapps, where having port opened is usually ok. Although I'm thinking now whether it shouldn't be rather more specific functionality and instead of checking a port readiness, check output of curl call repeatedly and timeout if the daemon does not return expected string.

praiskup:
On Monday, March 23, 2020 7:15:30 PM CET Honza Horak wrote:
> Right, for db containers, we should probably use more specific method.

Perhaps, if specified, you could use the defined readiness/liveness check
directly?

hhorak:
> On Monday, March 23, 2020 7:15:30 PM CET Honza Horak wrote: Right, for db containers, we should probably use more specific method.
> Perhaps, if specified, you could use the defined readiness/liveness check directly?

Not sure I'm still following, do you mean like `podman exec /usr/libexec/check-container`? (https://github.com/sclorg/postgresql-container/blob/master/examples/postgresql-ephemeral-template.json#L133)

praiskup:
On Monday, March 23, 2020 7:35:51 PM CET Honza Horak wrote:
> do you mean like `podman exec /usr/libexec/check-container`?

That's what I meant but I'm not sure it is possible.  Sorry for interrupt,
feel free to continue with this...

pkubatrh:
I agree. Would be nice to check container readiness the same way we do it in the templates if possible.

hhorak:
> I agree. Would be nice to check container readiness the same way we do it in the templates if possible.

Agreed, where we have templates, it's the way to go.

But we don't have such templates for s2i images like httpd (that this PR was originally inspired by) -- those need either "curl in loop" or in worse case (e.g. httpd) they use "sleep X" and see test failures when X is not enough from some reason (that is the reason why tests are failing in sclorg/httpd-container#92 now).

So, what I'm after here is that exact case -- either have a generic function that waits until a port is listening, or till that port also serves some content (in case HTTP protocol is used in such cases, which seems to be the case). It doesn't seem to be worth re-defining such function in every container test script.

hhorak:
It looks like the waiting in loop can be achieved also with the `ct_test_response` function, at least in the case of httpd container that was my primary concern. The tests could be converted to not use sleep as seen in https://github.com/sclorg/httpd-container/pull/94. Therefore, closing this PR for now as there is probably not good enough need for this new function.

praiskup:
It's a pity this stayed unmerged.  I didn't want to discourage you from
merging.

hhorak:
> It's a pity this stayed unmerged. I didn't want to discourage you from merging.

No worry, I just didn't see much value in it after all, and it's always useful to save some code.. removed code, good code :)

hhorak:
Still see no real need for this, closing.

PR #143, opened 2020-04-14
openshift: wait for build pod separately

pkubatrh:
Fixes the openshift CI fail in https://github.com/sclorg/s2i-php-container/pull/283

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -139,6 +139,17 @@ function ct_os_get_pod_status() {
                             | awk '{print $1}' | head -n 1
 }
 
+# ct_os_get_build_pod_status POD_PREFIX
+# --------------------
+# Returns status of the build pod specified by prefix [pod_prefix].
+# Arguments: pod_prefix - prefix or whole ID of the pod
+function ct_os_get_build_pod_status() {
+  local pod_prefix="${1}" ; shift
+  local query="custom-columns=Ready:status.containerStatuses[0].state.terminated.reason,NAME:.metadata.name"
+  oc get pods -o "$query" | grep -e "${pod_prefix}" | grep -E "\-build$" \
+                          | awk '{print $1}' | head -n 1
+}
+
 # ct_os_get_pod_name POD_PREFIX
 # --------------------
 # Returns the full name of pods specified by prefix [pod_prefix].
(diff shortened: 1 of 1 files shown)


pkubatrh:
@phracek PTAL

test-lib-openshift.sh
```diff
@@ -177,6 +188,18 @@ function ct_os_check_pod_readiness() {
...
   local pod_prefix="${1}" ; shift
   local timeout="${1}" ; shift
+  # If there is a build pod - wait for it to finish first
+  sleep 3
+  if ct_os_get_all_pods_name | grep -E "${pod_prefix}.*-build"; then
+    SECONDS=0
+    echo -n "Waiting for ${pod_prefix} build pod to finish ..."
+    while ! [ "$(ct_os_get_build_pod_status "${pod_prefix}")" == "Completed" ] ; do
+      echo -n "."
+      [ "${SECONDS}" -gt "${timeout}0" ] && echo " FAIL" && return 1
```
phracek, line 198:
```suggestion
      [ "${SECONDS}" -gt "${timeout}0" ] && echo " FAIL" && ct_os_get_status && return 1
```

pkubatrh: We already have: https://github.com/sclorg/container-common-scripts/blob/master/test-lib-openshift.sh#L60 for debugging output in case of a failure. Would rather have it in one place than in every function.

pkubatrh:
Not sure if `ct_os_wait_rc_ready` is being used with s2i builds, but if yes I would rather split the function in two (one to check s2i build result and one to check for rc status).

phracek:
@pkubatrh In case of POD failed, like here: https://gist.github.com/rhscl-bot/399ec0aac1d605f5183bcb004f1629b5#file-s2i-php-container-48-sh-L4513 I don't see any output from `oc`?
That was the reason why I have added a function `ct_os_get_status`. But we can use `ct_os_print_logs` instead of `ct_os_get_status`

The function `ct_os_enable_print_logs` is not used by in test-lib-openshift.sh. I did not find out any call. It has to be called directly from container. Like here: https://github.com/sclorg/postgresql-container/blob/master/test/run-openshift-local-cluster#L422

pkubatrh:
Well yes, `ct_os_enable_print_logs` needs to be run in the container tests to enable the print logs feature, it is not enabled by default. But once it is enabled the logs get printed out on any failure during the test run.

If we added your suggestion to this PR then the logs would be printed out twice for this specific failure, if the print log feature is enabled. Is this explanation more understandable?

PR #149, opened 2020-05-12
test-lib: add function to check env vars presence during exec

pkubatrh:


pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
@phracek PTAL

test-lib.sh
```diff
@@ -311,6 +311,49 @@ EOF
...
+  # Get environment variables from `docker exec`
+  ct_create_container "test_exec_envs" bash -c "sleep 1000" >/dev/null
+  cid=$(ct_get_cid "test_exec_envs")
+  docker exec "$cid" env | sort > "$tmpdir/exec_envs"
+  # Filter out variables we are not interested in
+  # Always check X_SCLS, ignore PWD
+  # Check variables from `docker run` that have alternative paths inside (/opt/rh, /opt/app-root)
+  exec_envs=$(cat "$tmpdir/exec_envs")
+  run_envs=$(grep "^X_SCLS=\|/opt/rh\|/opt/app-root" "$tmpdir/run_envs" | grep -v "^PWD=")
+  for variable in $run_envs; do
```
hhorak, line 336: While it's very unlikely in the container, where the `env` is under our control, this would not work great if there is a whitespace in the path. I think it might be better without the `$run_envs` variable, but using `while read -r variable ; do ... done <<<$(grep "^X_S...)`.

test-lib.sh
```diff
@@ -311,6 +311,49 @@ EOF
...
+    for value in $stripped; do
+        if [ -n "${filtered_envs##*$value*}" ]; then
+            echo " Value $value is missing from variable $var_name"
+            return 1
+        fi
+    done
+  done
+  IFS=$old_IFS
+  echo " All values present in \`docker exec\`"
+  ct_cleanup
```
hhorak, line 353: I'm not sure whether it is a good idea to run `ct_cleanup` in one of the library functions like this, since it has effect on created containers that were not created in this function. It would remove them and displayed its log, but it would be confusing because the other containers are not related to this test. I'd suggest to rather either remove just the single created container or leave it up to the calling script to run `ct_cleanup`.

hhorak: Maybe we should make clear in the comment of `ct_cleanup` function how it is supposed to be used (whether it can be called multiple times). And we can also divide it to two functions: `ct_cleanup` working with all previously created containers, and `ct_cleanup_single` to cleanup only a single container.

test-lib.sh
```diff
@@ -311,6 +311,49 @@ EOF
...
+  docker run --rm "$IMAGE_NAME" /bin/bash -c "env" | sort > "$tmpdir/run_envs"
+  # Get environment variables from `docker exec`
+  ct_create_container "test_exec_envs" bash -c "sleep 1000" >/dev/null
+  cid=$(ct_get_cid "test_exec_envs")
+  docker exec "$cid" env | sort > "$tmpdir/exec_envs"
+  # Filter out variables we are not interested in
+  # Always check X_SCLS, ignore PWD
+  # Check variables from `docker run` that have alternative paths inside (/opt/rh, /opt/app-root)
+  exec_envs=$(cat "$tmpdir/exec_envs")
+  run_envs=$(grep "^X_SCLS=\|/opt/rh\|/opt/app-root" "$tmpdir/run_envs" | grep -v "^PWD=")
```
phracek, line 335: What about to grep variable `^X_SCLS=\|/opt/rh\|/opt/app-root` as an argument do we can check also the other variables. E.g. NPM_MIRROR etc.

test-lib.sh
phracek, line 353: Please remove it. This should be called from the test suite and not from function itself.

phracek:
I have tested it locally with a couple more `echo` and the result is:

```bash
LD_LIBRARY_PATH not found during ` docker exec`
```

test-lib.sh
pkubatrh: Thanks for catching this! I wanted to move the call to nodejs testuite anyway, so let us do that. Makes more sense to do cleanup with log dumps after the test suite completes anyway.

test-lib.sh
pkubatrh: Makes sense, will add.

test-lib.sh
pkubatrh: Good catch and I agree, let us go for the safer variant.

pkubatrh:
Ah damn, I wanted

> I have tested it locally with a couple more `echo` and the result is:
> 
> ```shell
> LD_LIBRARY_PATH not found during ` docker exec`
> ```

Did you run the test case against https://github.com/sclorg/s2i-nodejs-container/pull/230 ?

phracek:
I have checked it against https://github.com/sclorg/s2i-nodejs-container/pull/230 and all works fine.

test-lib.sh
```diff
@@ -311,6 +311,49 @@ EOF
...
+# Uses: $IMAGE_NAME - name of the image being tested
+function ct_check_exec_env_vars() {
+  local tmpdir exec_envs run_envs cid old_IFS
+  local var_name stripped filtered_envs
+  tmpdir=$(mktemp -d)
+  CID_FILE_DIR=${CID_FILE_DIR:-$(mktemp -d)}
+  # Get environment variables from `docker run`
+  docker run --rm "$IMAGE_NAME" /bin/bash -c "env" | sort > "$tmpdir/run_envs"
+  # Get environment variables from `docker exec`
+  ct_create_container "test_exec_envs" bash -c "sleep 1000" >/dev/null
```
phracek, line 328: Based on the PR https://github.com/sclorg/s2i-nodejs-container/pull/230 container is already running by this command https://github.com/sclorg/s2i-nodejs-container/blob/f4beccc99db5dde479965dccebe3f655cac838c9/test/run#L440 so you can only connect to already exists container and not create the new one. cid_file is already present and you can passed it as argument https://github.com/sclorg/s2i-nodejs-container/blob/f4beccc99db5dde479965dccebe3f655cac838c9/test/run#L448

pkubatrh: That is a possibility, but I would rather have the test case standalone for cases where no such running container is available. Considering we do not really wait on the container finishing some work I do not think it much of a problem to create a new one every time.

pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
Merging, thanks for review!

PR #151, opened 2020-05-18
test-lib: check if all envs are set after second scl enable

pkubatrh:
Required-by: s2i-nodejs-container#230

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -50,6 +50,33 @@ function ct_enable_cleanup() {
   trap ct_cleanup EXIT SIGINT
 }
 
+function ct_check_envs_set {
+  local env_filter check_envs env_format
+  env_filter=$1; shift
+  check_envs=$1; shift
+  loop_envs=$1; shift
+  env_format=${1:-"*VALUE*"}
+  while read -r variable; do
+    var_name=$(echo "$variable" | awk -F= '{ print $1 }')
+    stripped=$(echo "$variable" | awk -F= '{ print $2 }')
+    filtered_envs=$(echo "$check_envs" | grep "^$var_name=")
+    [ -z "$filtered_envs" ] && { echo "$var_name not found during \` docker exec\`"; return 1; }
+    old_IFS=$IFS
+    # For each such variable compare its content with the `docker exec` result, use `:` as delimiter
+    IFS=:
+    for value in $stripped; do
+        # If the falue checked does not go through env_filter we do not care about it
+        echo "$value" | grep -q "$env_filter" || continue
+        if [ -n "${filtered_envs##${env_format//VALUE/$value}}" ]; then
+            echo " Value $value is missing from variable $var_name"
+            echo "$filtered_envs"
+            return 1
+        fi
+    done
+  done <<< "$(echo "$loop_envs" | grep "$env_filter" | grep -v "^PWD=")"
+  IFS=$old_IFS
+}
+
 # ct_get_cid [name]
 # --------------------
 # Prints container id from cid_file based on the name of the file.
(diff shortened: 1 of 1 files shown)


pkubatrh:
[test]

test-lib.sh
```diff
@@ -322,40 +349,54 @@ EOF
...
 function ct_check_exec_env_vars() {
   local tmpdir exec_envs cid old_IFS env_filter
-  local var_name stripped filtered_envs
+  local var_name stripped filtered_envs loop_envs
   env_filter=${1:-"^X_SCLS=\|/opt/rh\|/opt/app-root"}
   tmpdir=$(mktemp -d)
   CID_FILE_DIR=${CID_FILE_DIR:-$(mktemp -d)}
   # Get environment variables from `docker run`
-  docker run --rm "$IMAGE_NAME" /bin/bash -c "env" | sort > "$tmpdir/run_envs"
+  loop_envs=$(docker run --rm "$IMAGE_NAME" /bin/bash -c "env")
```
phracek, line 357: Why `loop_envs`? `run_envs` was better

test-lib.sh
```diff
@@ -50,6 +50,33 @@ function ct_enable_cleanup() {
   trap ct_cleanup EXIT SIGINT
 }
 
+function ct_check_envs_set {
```
phracek, line 53: Some comment before this function whould be nice.

test-lib.sh
```diff
@@ -50,6 +50,33 @@ function ct_enable_cleanup() {
...
+  local env_filter check_envs env_format
+  env_filter=$1; shift
+  check_envs=$1; shift
+  loop_envs=$1; shift
+  env_format=${1:-"*VALUE*"}
+  while read -r variable; do
```
phracek, line 59: Please add there a check, that all local variables has to be set.
For case, that somebody will call it directly.

test-lib.sh
```diff
@@ -322,40 +349,54 @@ EOF
...
+# Uses: $IMAGE_NAME - name of the image being tested
+function ct_check_scl_enable_vars() {
+  local tmpdir exec_envs cid old_IFS env_filter enabled_scls
+  local var_name stripped filtered_envs loop_envs
+  env_filter=$1
+  tmpdir=$(mktemp -d)
+  enabled_scls=$(docker run --rm "$IMAGE_NAME" /bin/bash -c "echo \$X_SCLS")
+  if [ -z "$env_filter" ]; then
+    for scl in $enabled_scls; do
+      [ -z "$env_filter" ] && env_filter=/$scl && continue
```
phracek, line 386: What about this one:

```bash
  if [ -z "$env_filter ]; then
     $env_filter="/$scl"
  else
     $env_filter="env_filter|/$scl"
  fi
```

test-lib.sh
```diff
@@ -322,40 +349,54 @@ EOF
...
+#   choosing which variables to check in the test case.
+#   Defaults to paths containing enabled SCLS in the image
+# Uses: $IMAGE_NAME - name of the image being tested
+function ct_check_scl_enable_vars() {
+  local tmpdir exec_envs cid old_IFS env_filter enabled_scls
+  local var_name stripped filtered_envs loop_envs
+  env_filter=$1
+  tmpdir=$(mktemp -d)
+  enabled_scls=$(docker run --rm "$IMAGE_NAME" /bin/bash -c "echo \$X_SCLS")
+  if [ -z "$env_filter" ]; then
```
phracek, line 384:
```suggestion
  if [ -z "$env_filter" ]; then
      $env_filter=""
```

test-lib.sh
phracek, line 386: And this can be overwritten by:
env_filter is set to empty and defined before `for` cycle
```bash
 env_filter="$env_filter|/$scl"
```

test-lib.sh
pkubatrh: True, will add. Thanks!

test-lib.sh
pkubatrh: I can add something, but I think we are missing such a check for most of the library calls. I guess the expectation is that the caller should make sure its passed correctly?

test-lib.sh
pkubatrh: I wonder how that went in here. Probably some copy paste error while moving the code. Will change back. Thanks for spotting this!

test-lib.sh
pkubatrh: Can you elaborate on this please?

test-lib.sh
pkubatrh: You would end up with eg. `|/rh-nodejs10|/rh-postgresql10` this way.

phracek, line 386:
```suggestion
      [ -z "$env_filter" ] && env_filter="/$scl" && continue
```

test-lib.sh
phracek: ok, let's leave and we will see in the future.

pkubatrh:
Added changes based on review

[test]

pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
Thanks for review, merging

PR #152, opened 2020-05-26
Add a function to test an image stream

hhorak:
An example of the usage:
```
ct_os_test_image_stream \
    "imagestreams/postgresql-rhel7.json" \
    "examples/postgresql-ephemeral-template.json" \
    postgresql \
    "-p POSTGRESQL_VERSION=${VERSION}"
```

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -1073,13 +1073,24 @@ function ct_os_test_image_stream() {
   local service_name=$3
   local template_params=${4:-}
 
+  if [ $# -lt 3 ] || [ -z "${1}" ] || [ -z "${2}" ] || [ -z "${3}" ]; then
+    echo "ERROR: ct_os_test_image_stream() requires at least 3 arguments that cannot be emtpy." >&2
+    return 1
+  fi
+
   echo "Running image stream test for stream $image_stream_file and template $template_file"
   # shellcheck disable=SC2119
   ct_os_new_project
 
   oc create -f "${image_stream_file}"
   # shellcheck disable=SC2086
-  oc process -p NAMESPACE="$(oc project -q)" ${template_params} -f "${template_file}" | oc create -f -
+  if ! ct_os_deploy_template_image "${template_file}" -p NAMESPACE="$(oc project -q)" ${template_params} ; then
+    echo "ERROR: ${template_file} could not be loaded"
+    return 1
+    # Deliberately not runnig ct_os_delete_project here because user either
+    # might want to investigate or the cleanup is done with the cleanup trap.
+    # Most functions depend on the set -e anyway at this point.
+  fi
   ct_os_wait_pod_ready "${service_name}" 120
 
   # shellcheck disable=SC2119
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -1060,4 +1060,27 @@ function ct_os_check_cmd_internal() {
   return 1
 }
 
+# ct_os_test_image_stream
+# ------------------------
+# Creates an image stream and deploys a specified template. Then checks that a pod runs.
+# Argument: image_stream_file - local file name with an image stream
+# Argument: template_file - local file name with a template
+# Argument: service_name - how the pod will be named (prefix)
+function ct_os_test_image_stream() {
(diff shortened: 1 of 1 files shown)


hhorak:
I have some further enhancements in progress, so turning this to draft for a moment.

hhorak:
> I have some further enhancements in progress, so turning this to draft for a moment.

Disregard, it looks ok as it is.

test-lib-openshift.sh
```diff
@@ -1060,4 +1060,30 @@ function ct_os_check_cmd_internal() {
...
+  local service_name=$3
+  local template_params=${4:-}
+
+  echo "Running image stream test for stream $image_stream_file and template $template_file"
+  # shellcheck disable=SC2119
+  ct_os_new_project
+
+  oc create -f "${image_stream_file}"
+  # shellcheck disable=SC2086
+  oc process -p NAMESPACE="$(oc project -q)" ${template_params} -f "${template_file}" | oc create -f -
```
phracek, line 1082: What about if `oc process` failed. Would it be possible to add there check that `ct_os_wait_pod_ready` will be called only in success?

test-lib-openshift.sh
```diff
@@ -1060,4 +1060,30 @@ function ct_os_check_cmd_internal() {
...
+# Argument: template_file - local file name with a template
+# Argument: service_name - how the pod will be named (prefix)
+# Argument: template_params (optional) - parameters for the template, like image stream version
+function ct_os_test_image_stream() {
+  local image_stream_file=$1
+  local template_file=$2
+  local service_name=$3
+  local template_params=${4:-}
+
+  echo "Running image stream test for stream $image_stream_file and template $template_file"
```
phracek, line 1076: Please add there a check that files really exist. I would like to avoid the typos.

phracek: and of course, `image_stream_file`, `template_file`, and `service_name` should not be empty, otherwise `oc create -f` will wait and maybe block the test forever.

test-lib-openshift.sh
phracek: By the way. This function is partially done by https://github.com/sclorg/container-common-scripts/blob/master/test-lib-openshift.sh#L275. What about to use it?

test-lib-openshift.sh
hhorak: Ok, thanks for the suggestions, will add some more sanity checks.

test-lib-openshift.sh
hhorak: Good catch, will call `ct_os_deploy_template_image` instead and also add the check there.

hhorak: Actually, for the file check itself, I believe the extra check is not necessary, because if the file does not exist, user will see pretty clear error message from the `oc process` command.

test-lib-openshift.sh
hhorak: Actually, for the file check itself, I believe the extra check is not necessary, because if the file does not exist, user will see pretty clear error message from the `oc process` command. I'll only add checks for having the values specified.

hhorak:
Hope the notes were addressed in the latest commit.

test-lib-openshift.sh
```diff
@@ -1060,4 +1060,41 @@ function ct_os_check_cmd_internal() {
...
+# Argument: service_name - how the pod will be named (prefix)
+# Argument: template_params (optional) - parameters for the template, like image stream version
+function ct_os_test_image_stream() {
+  local image_stream_file=$1
+  local template_file=$2
+  local service_name=$3
+  local template_params=${4:-}
+
+  if [ $# -lt 3 ] || [ -z "${1}" ] || [ -z "${2}" ] || [ -z "${3}" ]; then
+    echo "ERROR: ct_os_test_image_stream() requires at least 3 arguments that cannot be emtpy." >&2
```
phracek, line 1077:
```suggestion
    echo "ERROR: ct_os_test_image_stream() requires at least 3 arguments that cannot be empty." >&2
```

hhorak:
Thanks for the typo commit. I see no other issues reported, so merging now, because #155 depends on this.

PR #153, opened 2020-05-28
Add a function to test the latest version in an image stream

phracek:
The commit contains these changes:
* test for check latest image stream
* python file which checks the latest version against image
stream tags
* function `ct_check_latest_imagestreams`

Required-by: s2i-nodejs-container#241

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ check-squash:
 	./tests/squash/squash.sh
 
 check-latest-imagestream:
-	cp check_imagestreams.py tests/ && cd tests && ./check_imagestreams.sh
+	cd tests && ./check_imagestreams.sh
 
 check: check-failures check-squash check-latest-imagestream
 	TESTED_IMAGES="$(TESTED_IMAGES)" tests/remote-containers.sh
--- a/Makefile
+++ b/Makefile
@@ -39,5 +39,8 @@ check-failures: check-test-lib
 check-squash:
 	./tests/squash/squash.sh
 
-check: check-failures check-squash
+check-latest-imagestream:
+	cp check_imagestreams.py tests/ && cd tests && ./check_imagestreams.sh
+
+check: check-failures check-squash check-latest-imagestream
 	TESTED_IMAGES="$(TESTED_IMAGES)" tests/remote-containers.sh
--- a/check_imagestreams.py
+++ b/check_imagestreams.py
@@ -1,0 +1,63 @@
+#! /bin/python
+
+import os
+import sys
+import json
+import logging
+
+from pathlib import Path
+from typing import Dict
+
+IMAGESTREAMS_DIR: str = "imagestreams"
+
+
+class ImageStreamChecker(object):
+    version: str = ""
+    results: Dict = {}
+
+    def __init__(self, version: str):
+        self.version = version
+
+    def load_json_file(self, filename: Path):
+        with open(str(filename)) as f:
+            return json.load(f)
+
+    def check_version(self, json_dict: Dict):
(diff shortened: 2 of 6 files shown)


test-lib.sh
```diff
@@ -773,4 +773,10 @@ ct_check_image_availability() {
   fi
 }
 
+ct_check_latest_is() {
+    local latest_version=
+
+    latest_version=$(grep 'VERSIONS' Makefile | rev | cut -d ' ' -f 1 | rev )
+    python3 "./check_imagestream.py" "$latest_version" | return 1
```
hhorak, line 780: I guess you wanted to use `... || return 1` but it is not necessary either, because return value of the function is what the last command returns. So, I believe you can remove the `| return 1` entirely.

phracek: Good catch

pkubatrh:
[test]

Makefile
```diff
@@ -39,5 +39,8 @@ check-failures: check-test-lib
 check-squash:
 	./tests/squash/squash.sh
 
-check: check-failures check-squash
+check-latest-imagestream:
+	cp check_imagestreams.py tests/ && cd tests && ./check_imagestreams.sh
```
hhorak, line 43: Why the script is copied? Wouldn't it be enough to `cd tests && ../check_imagestreams.sh`?

check_imagestreams.py
```diff
@@ -0,0 +1,63 @@
+#! /bin/python
```
hhorak, line 1: Having my Python hat on, shebangs should always have version. Another thing is that on some systems (RHEL-7), the python might be provided by SCL, so the shebang might be `#!/bin/env python3`. For example on our RHEL-7 CI machine, the python3 is available through virtualenv:
```
# which python3
/usr/local/python36-tools/bin/python3
```

test-lib.sh
```diff
@@ -773,4 +773,16 @@ ct_check_image_availability() {
...
+# ct_check_latest_imagestreams
+# -----------------------------
+# Check if the latest version present in Makefile in the variable VERSIONS
+# is present in all imagestreams.
+# Also the latest tag in the imagestreams has to contain the latest version
+ct_check_latest_imagestreams() {
+    local latest_version=
+
+    latest_version=$(grep 'VERSIONS' Makefile | rev | cut -d ' ' -f 1 | rev )
+    python3 ./common/check_imagestreams.py "$latest_version"
```
hhorak, line 785: `./common` will work only when we are in the git repo root. It might be enough for the default case (`make check`), but will not work when I run `./run` from the `tests/` dir. I'm not sure what is the best solution here, but there is already `$(dirname "$(readlink -f "$0")")` used below, so maybe we can use the same trick and call the `check_imagestreams.py` from the same directory as this `test-lib.sh` is in..

Makefile
phracek: I will try it. For some reason it failed when I have implemented.

check_imagestreams.py
phracek: And my stupid typo in shebang;)

Makefile
phracek: Good catch. works.

phracek:
I have fixed all @hhorak  issues by commit https://github.com/sclorg/container-common-scripts/pull/153/commits/dd8eb621c29c858a008db3d6b68c89b06692e457

phracek:
[test]

phracek:
Rebased on the top of master.

[test]

hhorak:
I don't see any issues now.

PR #156, opened 2020-06-18
Fix issue with VERSIONS in Makefile

phracek:
Makefile for postresql-container https://github.com/sclorg/postgresql-container/blob/master/Makefile#L5
contains two VERSIONS in case we use grep.
Line has to start with VERSIONS.

This is easy fix.


phracek:
[test]

phracek:
@pkubatrh PTAL

pkubatrh:
lgtm. thanks!

PR #161, opened 2020-07-27
Add ct_test_app_dockerfile function for testing s2i usage using a Dockerfile

hhorak:
This function is supposed to be called for example applications to check that:

* the example apps also work fine when using with a Dockerfile
* the Dockerfile works as expected

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -832,6 +832,11 @@ ct_test_app_dockerfile() {
   local ret
   local cname=app_dockerfile
 
+  if [ -z "$app_dir" ] ; then
+    echo "ERROR: Option app_dir not set. Terminating the Dockerfile build."
+    return 1
+  fi
+
   if ! [ -r "${dockerfile}" -a -s "${dockerfile}" ] ; then
     echo "ERROR: Dockerfile ${dockerfile} does not exist or is empty."
     echo "Terminating the Dockerfile build."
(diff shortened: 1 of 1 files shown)


test-lib.sh
```diff
@@ -788,4 +788,53 @@ ct_check_latest_imagestreams() {
...
+#                        (must work with an application directory called 'app-src')
+# Argument: app_url - git URI with a testing application
+# Argument: body_regexp - PCRE regular expression that must match the response body
+# Argument: port - Optional port number (default: 8080)
+ct_test_app_dockerfile() {
+  local dockerfile=$1
+  local app_url=$2
+  local expected_text=$3
+  local port=${4:-8080}
+  local app_dir=app-src # this is a directory that must match with the name in the Dockerfile
```
phracek, line 803: Double quotes are missing, right?

phracek, line 803:
```suggestion
  local app_dir="app-src" # this is a directory that must match with the name in the Dockerfile
```

hhorak: I've changed this -- it's now an argument that must be configured.

hhorak: I've changed this -- it's now an argument that must be configured.

hhorak:
@pkubatrh @phracek I believe this is ready for review. Is anybody from you able to take a look?

hhorak:
Sure, `Required-by: s2i-nodejs-container#247` added. I'll need to study how this works, this is something I've not used before.

pkubatrh:
Thanks! Will also need to be rebased against master.\

>I'll need to study how this works, this is something I've not used before.

What it does is when the tests are run for this PR, CI will also clone the PR mentioned in the commit and run its tests with the common submodule switched to this PR. So we can test both changes at the same time.

pkubatrh:
[test]

hhorak:
Rebased

PR #167, opened 2020-08-21
Fix check_imagestreams.py after recent changes in imagestreams

hhorak:
The latest tag does not need to necessarily link to `<stream>`, but also `<stream>-elX` or `<stream>-ubiX`. See example at https://github.com/sclorg/httpd-container/pull/98.


hhorak:
A one-liner, that looks very obvious and blocks other PRs. Hope it is ok to merge without a rigid review.

PR #168, opened 2020-08-25
move .exclude file checks to the shell wrapper

pkubatrh:
The exclude file existence check was moved to the shell wrapper.
We need to check for the files only in the version that is latest.
But do not ignore the run if we find that it is excluded. Rather
take the next version down the line and run the test for it instead.
eg. when node 14 is latest in Makefile, but .exclude is found,
run the test for node 12.

Also adds additional check for limiting the test runs, so we
only run the test when the latest version matches the version
being tested.


pkubatrh:
[test]

hhorak:
LGTM

PR #169, opened 2020-08-26
fix version check in check_imagestreams.py

pkubatrh:
Required-by: s2i-python-container#394


pkubatrh:
@hhorak PTAL

[test]

frenzymadness:
The code looks reasonable and I can confirm that it works in testing PR: https://github.com/sclorg/s2i-python-container/pull/396

PR #170, opened 2020-09-22
Do not try to check imagestreams for Fedora images

hhorak:
Should help https://github.com/sclorg/s2i-nodejs-container/pull/254 to pass for Fedora image.


pkubatrh:
lgtm, thanks for the fix Honza!

PR #172, opened 2020-09-23
Add testing image in OpenShift 4 into test.sh

phracek:
This commit fixes problem in case image is not built.
Also removes duplicate code which is present in
test-remote-cluster.sh

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -59,7 +59,7 @@ for dir in ${VERSIONS}; do
     if [[ -x test/run-openshift ]]; then
       VERSION=$dir test/run-openshift
     else
-      echo "-> OpenShift tests are not present, skipping"
+      echo "-> OpenShift 3 tests are not present, skipping"
     fi
   fi
 
--- a/README.md
+++ b/README.md
@@ -35,8 +35,12 @@ It expects the test to be available at `$gitroot/$version/test/run`
 Depends on `tag` as some tests might need to have the images tagged (s2i).
 
 `make test-openshift`  
-Similar to `make test` but runs testsuite for Openshift, expected to be found at
+Similar to `make test` but runs testsuite for Openshift 3, expected to be found at
 `$gitroot/$version/test/run-openshift`
+
+`make test-openshift-4`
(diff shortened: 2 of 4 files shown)


phracek:
[test]

phracek:
@pkubatrh Can you please look at it thanks. Generally, if the image is not built, then do not run OpenShift 4 tests.

test.sh
```diff
@@ -48,20 +61,14 @@ for dir in ${VERSIONS}; do
   fi
 
   if [ -n "${TEST_OPENSHIFT_4}" ]; then
+    export IMAGE_NAME="${NAMESPACE}${BASE_IMAGE_NAME}-${dir//./}-${OS}"
```
pkubatrh, line 64: Is this necessary? `IMAGE_NAME` is set a few lines above, at the start of the cycle.

test.sh
```diff
@@ -48,20 +61,14 @@ for dir in ${VERSIONS}; do
...
   fi
 
-  if [ -n "${TEST_OPENSHIFT_MODE}" ]; then
-    if [[ -x test/run-openshift ]]; then
-      VERSION=$dir test/run-openshift
-    else
-      echo "-> OpenShift tests are not present, skipping"
-    fi
-  fi
-
```
pkubatrh, line 65: This part runs Openshift 3 locally, I do not think we want to remove these right?

test.sh
```diff
@@ -8,6 +8,19 @@
...
+    NAMESPACE=${NAMESPACE:-rhscl/}
+    REGISTRY=${REGISTRY:-registry.access.redhat.com/}
+  else
+    NAMESPACE=${NAMESPACE:-centos/}
+  fi
+
+  export NAMESPACE
+  export REGISTRY
+fi
+
```
pkubatrh, line 23: Is this part needed? I do not see a mention of `$REGISTRY` or `$NAMESPACE` in the nodejs PR. OS should always be set at this point (from Makefile).

test.sh
phracek: Stupid typo :(

pkubatrh:
LGTM, thanks!

PR #175, opened 2020-10-08
Imagestream aarch64 are not support on RHEL7

phracek:
The image streams `aarch64` are not supported on RHEL7.

Therefore skip the tests.

### DIFF
--- a/check_imagestreams.py
+++ b/check_imagestreams.py
@@ -3,6 +3,7 @@
 import sys
 import json
 import logging
+import os
 
 from pathlib import Path
 from typing import Dict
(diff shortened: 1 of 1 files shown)


check_imagestreams.py
```diff
@@ -47,6 +48,9 @@ def check_imagestreams(self):
             print(f"No json files present in {IMAGESTREAMS_DIR}.")
             return 0
         for f in json_files:
+            if os.environ.get("TARGET") in ("rhel7", "centos7") and str(f).endswith("aarch64.json"):
+                print("Imagestream arch64 is not supported on rhel7")
```
pkubatrh, line 52: Typo: arch64 -> aarch64

check_imagestreams.py
```diff
@@ -47,6 +48,9 @@ def check_imagestreams(self):
             print(f"No json files present in {IMAGESTREAMS_DIR}.")
             return 0
         for f in json_files:
+            if os.environ.get("TARGET") in ("rhel7", "centos7") and str(f).endswith("aarch64.json"):
```
pkubatrh, line 51: Would check for substring `aarch64` instead, in case we end up renaming the imagestream files yet again.

pkubatrh (commented):
Some minor issues, otherwise LGTM.

check_imagestreams.py
phracek: It was my first proposal, but for renaming case it make sense.

phracek:
Thanks for the review. I have updated code based on your proposals.

PR #176, opened 2020-10-12
Support testing imagestreams in OpenShift 4

phracek:
This pull request supports testing
imagestreams in OpenShift 4


phracek:
@hhorak @pkubatrh  PTAL. This pull request supports testing imagestreams in OpenShift 4.  Hopefully, this is the final Pull Request and we have full testing support in OpenShift 4

phracek:
Each imagestream contains a definition of which container from which registry is used. By default `registry.redhat.io.`
The default OpenShift 4 installation contains several samples like `registry.redhat.io/rhscl/nodejs-12-rhel7`. But we would like to test imagestream against our built image.

Command `oc image mirror registry.redhat.io/rhscl/nodejs-12-rhel7 <OUR DOCKER REGISTRY>` is used for replacing the default container from imagestream into ours. We don't want to modify imagestream to ours.

hhorak:
Wouldn't it be better to define the imagestream by ourselves? I'm thinking about the case when OpenShift 4.x that we'd use does not have a stream for a new version -- in that case, we cannot test the particular version with the default imagestream. But if we added a new imagestream, we can and also can point the image to an appropriate image location. But maybe I'm missing something.

phracek:
That sounds good.

My idea.

We have a JSON file e.g. https://github.com/sclorg/s2i-nodejs-container/blob/master/imagestreams/nodejs-rhel.json.
We can take a version from this imagestream and update it with our image exported to the docker-registry.
And if the version does not exist, then select a version that can be used as a template like `10`.
This will be defined in the test.
I will create a new python script that will do the magic.

@hhorak @pkubatrh WDYT

pkubatrh:
@hhorak I am not sure I follow. As far as I can tell the test in question is about creating an imagestream we maintain in this repository inside a remote OCP cluster and running a templated app using the imagestream. Considering all pieces come from this repository, how could we have something missing?
I guess the image might not yet be released, so it could be missing from registries, but that could be added as a special case.

phracek:
What about to test image from upstream defined by imagestream in OpenShift.

E.g. ImageStream s2i-nodejs-container for version:
```json
      {
        "name": "10",
        "annotations": {
          "openshift.io/display-name": "Node.js 10",
          "openshift.io/provider-display-name": "Red Hat, Inc.",
          "description": "Build and run Node.js 10 applications on RHEL 7. For more information about using this builder image, including OpenShift considerations, see https://github.com/sclorg/s2i-nodejs-container/blob/master/10/README.md.",
          "iconClass": "icon-nodejs",
          "tags": "builder,nodejs,hidden",
          "version": "10",
          "sampleRepo": "https://github.com/sclorg/nodejs-ex.git"
        },
        "from": {
          "kind": "DockerImage",
          "name": "registry.redhat.io/rhscl/nodejs-10-rhel7:latest"
        },
        "referencePolicy": {
          "type": "Local"
        }
      }

```
We have built the image `rhscl/nodejs-12-rhel7` in CI and would like to test it by imagestream. The built image is uploaded into our external repostory `docker-registry....`
But this repository is not present in the imagestream. We we update imagestream so it looks like:

```json
      {
        "name": "10",
        "annotations": {
          "openshift.io/display-name": "Node.js 10",
          "openshift.io/provider-display-name": "Red Hat, Inc.",
          "description": "Build and run Node.js 10 applications on RHEL 7. For more information about using this builder image, including OpenShift considerations, see https://github.com/sclorg/s2i-nodejs-container/blob/master/10/README.md.",
          "iconClass": "icon-nodejs",
          "tags": "builder,nodejs,hidden",
          "version": "10",
          "sampleRepo": "https://github.com/sclorg/nodejs-ex.git"
        },
        "from": {
          "kind": "DockerImage",
          "name": "docker-registry....../nodejs-10-rhel7"
        },
        "referencePolicy": {
          "type": "Local"
        }
      }
```
then we can test it also from imagestream point of view. Including new versions which does not exist yet in OpenShift 4.X.

phracek:
@pkubatrh @hhorak I have investigated the issue a little bit more and found some a bit tricky issue for testing imagestreams inside OpenShift 4. We have to update two files for testing imagestreams in OpenShift 4.

E.g. s2i-nodejs-container:
Imagestream file used for testing is https://github.com/sclorg/s2i-nodejs-container/blob/master/imagestreams/nodejs-rhel.json
The template file used for testing is https://github.com/sclorg/nodejs-ex/blob/master/openshift/templates/nodejs.json.

E.g. postgresql-container:
Imagestream file used for testing is: https://github.com/sclorg/postgresql-container/blob/master/imagestreams/postgresql-rhel.json
The template used for testing is https://github.com/sclorg/postgresql-container/blob/master/examples/postgresql-ephemeral-template.json

It is no problem to update both files. Pros: we can test whatever we want, including new versions. Cons: We have to create a script which will update it.

But we can use `oc image mirror`. Pros. It already works with all 'old' versions. Cons: I don't know if it works with the latest versions which are not released yet in OpenShift.

hhorak:
> @hhorak I am not sure I follow. As far as I can tell the test in question is about creating an imagestream we maintain in this repository inside a remote OCP cluster and running a templated app using the imagestream. Considering all pieces come from this repository, how could we have something missing?

I meant this case: Since OCP takes imagestreams and templates sometime during its development, there might be more pieces missing than just what we didn't release. If we test on e.g. OCP 4.4, the imagestreams available are a year old. So even images that we maintain in production for few months are missing there by default. Therefore we need own imagestream.

hhorak:
> It is no problem to update both files. Pros: we can test whatever we want, including new versions. Cons: We have to create a script which will update it.
> 
> But we can use `oc image mirror`. Pros. It already works with all 'old' versions. Cons: I don't know if it works with the latest versions which are not released yet in OpenShift.

You mean to mirror the new image (e.g. postgresql 12) as rhel8/postgresql-10? That might lead to confusions and would make debugging harder IMO. So I personally prefer the first option -- rewrite the imagestream and a template.

Or, we can use a different approach and for the purpose of testing the specific image as part of PR tests, to have a dedicated imagestream and also template. It would basically mean to not mix the imagestreams that we maintain for usage by OpenShift with files that we use for testing. The testing files would not change when adding a new version, so it is not a big maintenance cost. It means to have e.g. `test/imagestreams/postgresql-is.json` and `test/templates/postgresql-ephemeral.json` and only work with those copies. A test for official `imagestreams/*json` and `examples/*json` would be done separately. It would be more clear IMO. (just an idea)

pkubatrh:
> I meant this case: Since OCP takes imagestreams and templates sometime during its development, there might be more pieces missing than just what we didn't release. If we test on e.g. OCP 4.4, the imagestreams available are a year old. So even images that we maintain in production for few months are missing there by default. Therefore we need own imagestream.

Ah, that must have been the part I was missing - I thought we are talking about testing the imagestream configuration files we maintain in our repositories, not the one that is included in a specific OCP cluster. Another question pops into my mind though - why test an already released imagestream?

phracek:
> - why test an already released imagestream?

My thinking was to test the built image from PR in CI with the imagestream which we have in our repository. Whether it is working or not.
Our tests use as imagestreams present in the container repository as template stored either in the same repository or in the examples repository, like `nodejs-ex`.

hhorak:
> why test an already released imagestream?

Now I'm not sure what you mean by "already released imagestream" -- is it the imagestream available as default in OCP or is it the file we have in our github registry under imagestream/ dir?

pkubatrh:
> Now I'm not sure what you mean by "already released imagestream" -- is it the imagestream available as default in OCP or is it the file we have in our github registry under imagestream/ dir?

Here I meant any imagestream that is released as part of an OCP release. For example any imagestream currently present in OCP 4.4.

After thinking about it for a while there might still be a value in testing both released, and unreleased (part of the PR tested) imagestreams - to check whether an image works with both imagestreams in case there is some change regarding the example application (moving to some other repository?).

pkubatrh:
In that case I think the best we can do is to import and test whatever imagestream is currently part of Origin master. That imagestream should have all the image streams that we care about right?

phracek:
@pkubatrh @hhorak  NAMESPACE parameter works as we expected. PTAL on it. as a reference PR in s2i-perl-container you can use https://github.com/sclorg/s2i-perl-container/pull/188

As soon as this pull request is merged I will try to test it directly in s2i-per-container pull request [test-openshift-4]

hhorak (approved):
@phracek good work, LGTM.

PR #179, opened 2020-11-02
Missing which and too fast failures

hhorak:
We don't have `which` in all containers. Also, in some cases when, when tests fail before the first container is created, the cleanup can fail by trying to remote empty string.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -341,7 +341,7 @@ function ct_binary_found_from_df() {
   # Create Dockerfile that looks for the binary
   cat <<EOF >"$tmpdir/Dockerfile"
 FROM $IMAGE_NAME
-RUN which $binary | grep "$binary_path"
+RUN command -v $binary | grep "$binary_path"
 EOF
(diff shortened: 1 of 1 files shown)


test-lib.sh
```diff
@@ -27,6 +27,7 @@ EXPECTED_EXIT_CODE=0
 function ct_cleanup() {
   ct_show_resources
   for cid_file in "$CID_FILE_DIR"/* ; do
+    [ -f "$cid_file" ] || break
```
phracek, line 30: If dir `$CID_FILE_DIR` contains more files then only first is checked. What about `continue`?

phracek (changes requested):

test-lib.sh
hhorak: good idea.

hhorak: fixed.

phracek (approved):
LGTM. Thanks for the correction.

phracek:
[test]

PR #181, opened 2020-11-19
Add support for quay.io for CentOS7 images

phracek:
This pull request adds support for quay.io registry.
All RHSCL CentOS images will be soon moved into quay.io registry because of docker pull rate limit.

### DIFF
--- a/tag.sh
+++ b/tag.sh
@@ -18,11 +18,11 @@ for dir in ${VERSIONS}; do
   commit_date=$(git show -s HEAD --format=%cd --date=short | sed 's/-//g')
   date_and_hash="${commit_date}-$(git rev-parse --short HEAD)"
 
-
-  echo "-> Tagging image '$IMAGE_ID' as '$REGISTRY$name:$version' and '$REGISTRY$name:latest' and '$REGISTRY$name:$date_and_hash'"
-  docker tag "$IMAGE_ID" "$REGISTRY$name:$version"
-  docker tag "$IMAGE_ID" "$REGISTRY$name:latest"
-  docker tag "$IMAGE_ID" "$REGISTRY$name:$date_and_hash"
+  full_reg_name="$REGISTRY$name"
+  echo "-> Tagging image '$IMAGE_ID' as '$full_reg_name:$version' and '$full_reg_name:latest' and '$full_reg_name:$date_and_hash'"
+  docker tag "$IMAGE_ID" "$full_reg_name:$version"
+  docker tag "$IMAGE_ID" "$full_reg_name:latest"
+  docker tag "$IMAGE_ID" "$full_reg_name:$date_and_hash"
 
   for suffix in squashed raw; do
     id_file=.image-id.$suffix
--- a/tag.sh
+++ b/tag.sh
@@ -18,10 +18,11 @@ for dir in ${VERSIONS}; do
   commit_date=$(git show -s HEAD --format=%cd --date=short | sed 's/-//g')
   date_and_hash="${commit_date}-$(git rev-parse --short HEAD)"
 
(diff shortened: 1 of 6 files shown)


frenzymadness:
I wanted to comment that I'd prefer the trailing slash to not be a path of the REGISTRY variable but now I see that the variable might be empty which would lead to a slash at the beginning of the image name.

The change itself looks good to me and CI is green :+1:

tag.sh
```diff
@@ -18,10 +18,11 @@ for dir in ${VERSIONS}; do
   commit_date=$(git show -s HEAD --format=%cd --date=short | sed 's/-//g')
   date_and_hash="${commit_date}-$(git rev-parse --short HEAD)"
 
-  echo "-> Tagging image '$IMAGE_ID' as '$name:$version' and '$name:latest' and '$name:$date_and_hash'"
-  docker tag "$IMAGE_ID" "$name:$version"
-  docker tag "$IMAGE_ID" "$name:latest"
-  docker tag "$IMAGE_ID" "$name:$date_and_hash"
+
+  echo "-> Tagging image '$IMAGE_ID' as '$REGISTRY$name:$version' and '$REGISTRY$name:latest' and '$REGISTRY$name:$date_and_hash'"
```
pkubatrh, line 22: How about creating a new variable (eg. `full_name`) and use it instead of `$REGISTRY$name`. Might  be a bit more readable

pkubatrh:
Could we modify references to registry in test-lib as well? eg. https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L563

tests/test-lib/public_image_name
```diff
@@ -5,7 +5,7 @@ set -e
 . test-lib.sh
 
 combinations="rhel7:registry.redhat.io/rhscl/postgresql-10-rhel7
-centos7:docker.io/centos/postgresql-10-centos7
+centos7:quay.io/centos/ruby-27-centos7
```
pkubatrh, line 8: centos7 here I expect

pkubatrh:
Otherwise lgtm

pkubatrh:
[test]

phracek:
I guess, centos7 test will failed because of postresql-container, s2i-python-container and s2i-nodejs-container does not contain fixes for centos7. But let's see.

phracek:
But what is wrong is that New Jenkins did not trigger [test]. I will fix it soon.

test.sh
```diff
@@ -16,7 +16,11 @@ for dir in ${VERSIONS}; do
   IMAGE_VERSION=$(docker inspect -f "{{.Config.Labels.version}}" "$IMAGE_ID")
   # Kept also IMAGE_NAME as some tests might still use that.
   IMAGE_NAME="$(docker inspect -f "{{.Config.Labels.name}}" "$IMAGE_ID"):$IMAGE_VERSION"
-  export IMAGE_NAME
+  if [ x"${OS}" == "xcentos7" ]; then
+    export IMAGE_NAME="$REGISTRY$IMAGE_NAME"
+  else
+    export IMAGE_NAME
+  fi
```
pkubatrh, line 23: How about:

```suggestion
  export IMAGE_NAME="${REGISTRY-}$IMAGE_NAME"
```

pkubatrh:
From the logs:

```
14:29:18   RUN TEST 'image_availability'
14:30:07 quay.io/centos7/ruby-24-centos7 could not be downloaded via 'docker'
14:30:07 image_availability test completed successfully.
```
Seems like the result of this test is not failing the test suite?

phracek:
[test]

phracek:
[test]

phracek:
`image_availability` is passing. Thanks, @pkubatrh. centos7 failed on Docker rate limit.

phracek:
[test]

PR #182, opened 2020-11-23
tests: fail image_availability test case on failure

pkubatrh:


pkubatrh:
[test]

pkubatrh:
Ah right and now the tests will fail because of dockerhub :/

phracek (approved):
LGTM. Thanks for the fix.

PR #184, opened 2020-12-02
Return back old IFS in while loop

phracek:
This commit fixes the error with IFS.

In function `ct_check_envs_set` IFS is not set back properly.
In the while cycle, we set IFS to `:` but in the case of another cycle IFS is set to `:` instead of '\n'.


fila43:
looks good

PR #187, opened 2021-01-19
Support branch definition in URL for git clone command

frenzymadness:
In Python container, we need to test two different branches of [Django example project](https://github.com/sclorg/django-ex) because the old LTS version is EOL and the new one does not work on RHEL/Centos 7.

Inspired by [pip](https://pip.pypa.io/en/stable/reference/pip_install/#git), I've implemented a way to get a branch name from git repo URL for `ct_test_app_dockerfile` function. Basically, whatever you put after a `@` in a git repository URL is considered as a name of a branch and used in `git clone --branch <name>`.

I'm not sure on how many places and in how many ways we are doing `git clone` but, if accepted, we can implement the same behavior for all `git clone` and `s2i build` commands.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -827,7 +827,7 @@ ct_show_resources()
 # -----------------------------
 # Argument: dockerfile - path to a Dockerfile that will be used for building an image
 #                        (must work with an application directory called 'app-src')
-# Argument: app_url - git URI with a testing application
+# Argument: app_url - git URI with a testing application, supports "@" to indicate a different branch
 # Argument: body_regexp - PCRE regular expression that must match the response body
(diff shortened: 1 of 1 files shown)


frenzymadness:
[test]

frenzymadness:
`toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit` :disappointed:

frenzymadness:
[test]

pkubatrh:
[test]

phracek (approved):
LGTM. I have tested locally and the code is fine. Thanks for the PR!

phracek:
I will test this PR locally whether tests are passing.

frenzymadness:
Thank you!

Issue #190, opened 2021-02-18
test_from_dockerfile is broken


frenzymadness:
I've broken test_from_dockerfile by the implementation of the branch name support in https://github.com/sclorg/container-common-scripts/commit/b3d5a0ad020be4d5f165560c24cd1188b9ae6b87

It seems that the systems our CI is running on are too old and the `readarray` command there doesn't provide `-d` option. From my testing, the option is supported in bash version 5 but not in version 4.

I'll try to fix it asap. Please do not update in the meantime.

frenzymadness:
Correction: this breaks only if you use the new feature. If you don't, you can just expect an error message in an output like:
```
14:01:46 test/test-lib.sh: line 870: readarray: -d: invalid option
14:01:46 readarray: usage: readarray [-n count] [-O origin] [-s count] [-t] [-u fd] [-C callback] [-c quantum] [array]
14:01:46 Cloning into 'app-src'...
14:01:46 fatal: unable to access 'https://github.com/sclorg/django-ex.git@2.2.x/': The requested URL returned error: 400
14:01:46 ERROR: Git repository https://github.com/sclorg/django-ex.git@2.2.x cannot be cloned into app-src.
```
But the test continues with the original URL.

PR #191, opened 2021-02-18
Fix split command for old bash

frenzymadness:
Bash 4 does not support option -d for readarray command

Fixes: https://github.com/sclorg/container-common-scripts/issues/190

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -867,7 +867,7 @@ ct_test_app_dockerfile() {
 
   # If app_url contains @, the string after @ is considered
   # as a name of a branch to clone instead of the main/master branch
-  readarray -d "@" -t git_url_parts<<<"${app_url}"
+  IFS='@' read -ra git_url_parts <<< "${app_url}"
 
   if [ -n "${git_url_parts[1]}" ]; then
     git_clone_cmd="git clone --branch ${git_url_parts[1]} ${git_url_parts[0]} ${app_dir}"


test-lib.sh
```diff
@@ -867,7 +867,7 @@ ct_test_app_dockerfile() {
 
   # If app_url contains @, the string after @ is considered
   # as a name of a branch to clone instead of the main/master branch
-  readarray -d "@" -t git_url_parts<<<"${app_url}"
+  IFS='@' read -ra git_url_parts <<< "${app_url}"
```
phracek, line 870: I would be glad to have here `old_IFS` and after this row returns back.

Like here and bellow https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L78
Take an inspiration.

phracek (changes requested):
Thanks for the PR!. Please fix IFS as I have written in my comment.

test-lib.sh
phracek: And here returns back https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L91

frenzymadness: I don't think it's necessary to switch the variable back if you change it for only one command. Tested on Centos 7:
```
# echo $IFS

# IFS='@' read -ra git_url_parts <<< "test@testbranch"
# echo ${git_url_parts[0]}
test
# echo ${git_url_parts[1]}
testbranch
# echo $IFS

#
```

In this way, the IFS is an environment variable passed to one command only instead of a shell variable modified for subsequent commands.

torsava: @phracek I think the use case you linked is different, because you set IFS for the entire environment, but Lumir is setting it only for that one command, so it's not necessary to restore the variable afterwards.

phracek (approved):
Thanks for this PR!. After review comments. LGTM.

frenzymadness:
Thank you!

PR #192, opened 2021-03-11
Skip OpenShift tests

phracek:
if `.exclude-openshift` file is present then as OpenShift-3 as OpenShift-4 tests are skipped. E.g. nodejs-10 imagestream was remove before alone NodeJS-10 EOL and OpenShift tests are failing.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -52,18 +52,31 @@ for dir in ${VERSIONS}; do
   fi
 
   if [ -n "${TEST_OPENSHIFT_4}" ]; then
-    if [[ -x test/run-openshift-remote-cluster ]]; then
-        VERSION=$dir test/run-openshift-remote-cluster
+    # In case only imagestream is deprecated
+    # and the other tests should be working
+    if [ -e ".exclude-openshift" ]; then
+      echo "-> .exclude-openshift file exists for version $dir, skipping OpenShift-4 tests."
     else
-      echo "-> Tests for OpenShift 4 are not present. Add run-openshift-remote-cluster script, skipping"
+      if [[ -x test/run-openshift-remote-cluster ]]; then
+        VERSION=$dir test/run-openshift-remote-cluster
+      else
+        echo "-> Tests for OpenShift 4 are not present. Add run-openshift-remote-cluster script, skipping"
+      fi
     fi
+
   fi
 
   if [ -n "${TEST_OPENSHIFT_MODE}" ]; then
-    if [[ -x test/run-openshift ]]; then
-      VERSION=$dir test/run-openshift
+    # In case only imagestream is deprecated
+    # and the other tests should be working
+    if [ -e ".exclude-openshift" ]; then
+      echo "-> .exclude-openshift file exists for version $dir, skipping OpenShift tests."
     else
-      echo "-> OpenShift 3 tests are not present, skipping"
+      if [[ -x test/run-openshift ]]; then
+        VERSION=$dir test/run-openshift
+      else
+        echo "-> OpenShift 3 tests are not present, skipping"
+      fi
     fi
   fi
 


phracek:
@hhorak @pkubatrh This PR is a bit simple but solves the main issue with OpenShift tests and disables `conu` tests. WDYT. I did not create two PR because changes are simple. If you want to have two PR, please let me know. Commits are of course separated.

test.sh
```diff
@@ -7,6 +7,18 @@
...
+  if test -f .image-id.raw; then
+      local previous_id
+      previous_id=$(cat .image-id.raw)
+      if test "$IMAGE_ID" != "$previous_id"; then
+          # Also remove squashed image since it will change anyway
+          docker rmi "$previous_id" "$(cat .image-id)" || :
+          rm -f ".image-id.raw" ".image-id" || :
+      fi
+  fi
+}
```
pkubatrh, line 21: Does this not duplicate https://github.com/sclorg/container-common-scripts/blob/master/clean.sh?

phracek: yes. I did not source it from test.sh, yet ;)

test.sh
```diff
@@ -52,6 +66,13 @@ for dir in ${VERSIONS}; do
   fi
 
   if [ -n "${TEST_OPENSHIFT_4}" ]; then
+    # In case only imagestream is deprecated
+    # and the other tests should be working
+    if [ -e ".exclude-openshift" ]; then
+      echo "-> .exclude-openshift file exists for version $dir, skipping OpenShift-4 tests."
+      clean_image
+      continue
+    fi
```
pkubatrh, line 75: I would rather see this being tested based on the current versions located inside the imagestreams rather than have a hidden file inside the versioned directories. Seems error-prone (if we forget to remove it for some future version).

test.sh
```diff
@@ -60,6 +81,13 @@ for dir in ${VERSIONS}; do
   fi
 
   if [ -n "${TEST_OPENSHIFT_MODE}" ]; then
+    # In case only imagestream is deprecated
```
pkubatrh, line 84: Same over here

test.sh
```diff
@@ -27,6 +39,8 @@ for dir in ${VERSIONS}; do
   fi
 
   if [ -n "${TEST_CONU_MODE}" ]; then
+    echo "Conu tests are SUPPRESSED now."
+    continue
```
pkubatrh, line 43: We do not want to continue here, the openshift tests would be skipped that way.

pkubatrh: Maybe we should just remove this block. I do not see us using it again in the future tbh

test.sh
phracek: You are right, but what about this case. We remove the stream NodeJS-10 two months ago before NodeJS-10 is EOL.
Like we are planning.
As soon as I would like to test NodeJS-10 in OpenShift 3 or 4 then the imagestream does not exist, because it is not specified in imagestream. But a template requires imagestream for version 10.
The OpenShift tests will fail.

test.sh
pkubatrh: Its used via `make clean`

test.sh
pkubatrh: Imo checking that the imagestream version is not present anymore and skipping the tests is a valid strategy here - the imagestream is EOL, it makes sense to not test it anymore.

test.sh
phracek: The function is present here: https://github.com/sclorg/container-common-scripts/blob/master/build.sh#L67

phracek: I have imported the function from `build.sh` script

phracek:
@pkubatrh I have proposed a better solution which can be merged, I guess

phracek:
@pkubatrh PTAL on it one more time.

hhorak:
LGTM

PR #194, opened 2021-03-18
Support 'make betka'

phracek:
This PR adds new functionality to generate downstream sources based on the upstream sources.

What is covered by this pull request:
- [x] centos7 is not supported and checked in the scripts. There is no plan to support it.
- [x] fedora is supported. If DOWNSTREAM_BRANCH is not specified, then `main` is default. `master` does not exist at all. DOWNSTREAM_NAME has to be specified.
- [x] rhel is supported. DOWNSTREAM_NAME has to be specified because we don't know what to generate. DOCKER_IMAGE has to be specified. Like `rh-nodejs10`. DOWNSTREAM_BRANCH has to be specified, like `rhel-8.4.0` or `rhscl-3.6-rhel7`
- [x] tests are covered also for `make betka`
- [x] Documentation updated

Currently `make betka` is written in BASH and I don't think that Python will save time, but we will see.

Open Items:
If `DOWNSTREAM_NAME` and/or `DOWNSTREAM_BRANCH` is not specified, then we can have in container-common-scripts a file with mapping for default testing, like:
`s2i-nodejs-container    rh-nodej-14   rhscl-3.6-rhel7`

There is a PR https://github.com/sclorg/container-workflow-tool/pull/28 which fixes correct Dockerfile tag generation.

### DIFF
--- a/betka.sh
+++ b/betka.sh
@@ -1,0 +1,205 @@
+#!/bin/bash
+
+# This script is used to build images for dist-git repositories.
+#
+# The result directory will contain generated sources from upstream directories
+#
+
+set -ex
+
+CUR_DIR=$(pwd)
+TMP_DIR=$(mktemp -d)
+UPSTREAM_IMAGE_NAME=$(basename "$CUR_DIR")
+
+echo "Temporary dir is ${TMP_DIR}."
+echo "Upstream image name is: $UPSTREAM_IMAGE_NAME."
+
+function create_temp_dir() {
+  CURDIR_RESULTS="${CUR_DIR}/results-${dir}"
+  if [[ -d "${CURDIR_RESULTS}" ]]; then
+    echo "results directory already exist. Delete it."
+    rm -rf "${CURDIR_RESULTS}"
+  fi
+}
+
+function check_os() {
+  if [[ x"${OS}" == "xcentos7" ]]; then
+    echo "Target has to be specified."
+    echo "CentOS7 is not supported."
+    echo "e.g. make betka TARGET=fedora VERSIONS=XY"
+    exit 1
+  fi
+}
+
+function load_configuration() {
+  if [[ x"${OS}" == "xfedora" ]]; then
+    curl -L https://raw.githubusercontent.com/sclorg/container-workflow-tool/master/cwt_generator.config > cwt_config
+  else
+    curl -L https://url.corp.redhat.com/rhcwt_config > cwt_config
+  fi
+  # shellcheck disable=SC1091
+  source cwt_config
+}
+
+function check_os_and_parameters() {
+  if [[ x"${OS}" == "xfedora" ]]; then
+    echo "For generating dist-git sources to Fedora ${CWT_DOCKER_IMAGE} is used."
+    echo "'make betka' will try to sync main branch in Fedora land."
+    echo "If you want to change it then specify it by parameter DOWNSTREAM_BRANCH, like"
+    echo "'make betka TARGET=fedora DOWNSTREAM_BRANCH=f32'"
+  else
+    if [[ x"${CWT_DOCKER_IMAGE}" == "x" ]]; then
+      echo "Docker image for generating dist-git sources for RHEL has to be specified."
+      echo "Ask pkubat@redhat.com, hhorak@redhat.com or phracek@redhat.com for the name."
+      exit 1
+    fi
+  fi
+
+  if [[ x"${dir}" == "x" ]]; then
+    echo "VERSIONS has to be specified."
+    echo "Like: make betka TARGET=rhel7 VERSIONS=12 for s2i-nodejs-container."
+    exit 1
+  fi
+
+}
+
+function convert_branch_to_cwt_tool() {
+  local cwt_config
+  if [ x"${OS}" == x"rhel7" ]; then
+    # Example branch name rhscl-3.6-rhel-7 converted to rhscl360
+    IFS='-' read -ra branch_list <<<"${DOWNSTREAM_BRANCH}"
+    cwt_config="rhel7.yaml:${branch_list[0]}${branch_list[1]//.}0"
+  else
+    # Example branch name rhel-8.3.0 -> converted to rhel8.3
+    IFS='-' read -ra branch_list <<<"${DOWNSTREAM_BRANCH}"
+    IFS='.' read -ra release_list <<< "${branch_list[1]}"
+    cwt_config="rhel8.yaml:${branch_list[0]}${release_list[0]}.${release_list[1]}"
+  fi
+  echo "${cwt_config}"
+}
+
+function switch_to_branch() {
+  echo "Cloning downstream repo '${CLONE_URL}/${DOWNSTREAM_NAME} to ${RESULTS_DIR}."
+  git clone "${CLONE_URL}/${DOWNSTREAM_NAME}" "${RESULTS_DIR}"
+  pushd "${RESULTS_DIR}" >/dev/null
+  git checkout "${DOWNSTREAM_BRANCH}"
+  popd >/dev/null
+}
+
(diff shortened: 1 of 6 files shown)


pkubatrh:
For `DOWNSTREAM_NAME` - the mapping is actually already part of cwt and can be dumped via `cwt utils listupstream` command, eg.:
```
$ ./cwt utils listupstream | grep s2i-base-container
s2i-core s2i-core s2i-base-container https://github.com/sclorg/s2i-base-container.git
s2i-base s2i-base s2i-base-container https://github.com/sclorg/s2i-base-container.git
```

Would need some parsing and comparison to the `VERSION` being generated, WDYT?

phracek:
Thanks for the comment.
> For `DOWNSTREAM_NAME` - the mapping is actually already part of cwt and can be dumped via `cwt utils listupstream` command, eg.:
> 
> ```
> $ ./cwt utils listupstream | grep s2i-base-container
> s2i-core s2i-core s2i-base-container https://github.com/sclorg/s2i-base-container.git
> s2i-base s2i-base s2i-base-container https://github.com/sclorg/s2i-base-container.git
> ```
> 
What about if downstream has more names like rh-nodejs10, rh-nodejs12, etc?
> Would need some parsing and comparison to the `VERSION` being generated, WDYT?
This makes sense to me. Do you think, that separate PR is needed? Or add it into this one?
> 
> Edit: hmm the upstream path (= `VERSION`) is actually missing from that output, but could be added at the end.

This also expects that `cwt` or `rhcwt` are installed on the system. What about using DOCKER_IMAGE with a parameter like CWT_COMMAND and if it is present then do not run source generation but CWT_COMMAND.

pkubatrh:
>What about if downstream has more names like rh-nodejs10, rh-nodejs12, etc?

Hence why I added the mention of the possibility of `VERSION` being added to the output of `cwt`

>This also expects that cwt or rhcwt are installed on the system. 

Are we not already using a container image with `cwt` installed inside?

phracek:
> > What about if downstream has more names like rh-nodejs10, rh-nodejs12, etc?
> 
> Hence why I added the mention of the possibility of `VERSION` being added to the output of `cwt`
> 
> > This also expects that cwt or rhcwt are installed on the system.
> 
> Are we not already using a container image with `cwt` installed inside?
Yes, but by default, it generates source and not printing the output.

But If I am right then we can use the existing image and grep for the proper version https://github.com/sclorg/container-workflow-tool/blob/master/files/run.sh#L25
But before that, I have to clone the downstream repository and I do not have the name.

Workflow:
1. We are in the current container directory
2. create a tmp directory on PC
3. pull cwt container image
4. clone downstream directory - BRANCH and NAME is mandatory.
5. generate changes from upstream to downstream

pkubatrh:
The workflow I am suggesting (expecting `VERSION` in output as suggested in previous comments) is:

Workflow:

1. We are in the current container directory
2. **Get current git repository name and `VERSION` `make betka` is run for**
3. create a tmp directory on PC
4. pull cwt container image
5. **grep downstream name `cwt utils listupstream | grep $VERSION`**
5. clone downstream directory - BRANCH and NAME is mandatory.
6. generate changes from upstream to downstream

pkubatrh:
Atlernatively we could modify cwt's `--do-image` so it could in some way use upstream name and `$VERSION` for identifying the specific image, and then use it in the betka image by default.

phracek:
I have addressed all issues CWT_IMAGE is used for calling command `cwt utils listupstream`. Thanks for the great point.
The results are stored in the current directory into folder `results-{VERSIONS}` dir. 

In case of VERSIONS is not specified by `make betka` then `make betka` is called over all versions.
E.g. nodejs downstream changes for all versions will be stored in directories:
- results-10
- results-12
- results-14

The last point also from a usage point of view. Does it make sense to rewrite it into Python?
WDYT?

Also test would be covered well.

phracek:
[test]

phracek:
[test]

phracek:
CentOS CI failed on docker_squash.

```bash
14:32:16 Removing intermediate container 1f049f917b78
14:32:16 Successfully built 5a70dfb4dc59
14:32:16 Squashing the image '5a70dfb4dc59' from 'quay.io/centos7/s2i-core-centos7' layer.
14:32:16 CRITICAL:root:please install 'docker_squash' for Python 3.6.12
14:32:16 make[2]: Leaving directory `/root/sources/tests/failures/check'
```

I will investigated a little bit more. But this does not block this pull request.

phracek:
[test]

phracek:
Let's test centos7

[test]

phracek:
Let's test centos7

[test]

betka.sh
```diff
@@ -86,6 +87,35 @@ function pull_cwt_image() {
...
+  local cwt_output
+  if [[ x"${OS}" == "xfedora" ]]; then
+    command="cwt"
+  else
+    command="rhcwt"
+  fi
+  cwt_output="${command}_output"
+  # Run CWT tool in order to get downstream name for specific version
+  if ! docker run -it --rm "${CWT_DOCKER_IMAGE}" "${command}" \
+      --config="${DOWNSTREAM_BRANCH}".yaml utils listupstream > "${cwt_output}"; then
```
pkubatrh: Will `--config="${DOWNSTREAM_BRANCH}".yaml` work for rhcwt? rhcw has rhel7.yaml and rhel8.yaml configs, but the branches are eg. `rhscl-3.6-rhel-7` and `rhel-8.3.0`

betka.sh
```diff
@@ -45,27 +43,23 @@ function load_configuration() {
 function check_os_and_parameters() {
   if [[ x"${OS}" == "xfedora" ]]; then
     if [[ x"${DOWNSTREAM_BRANCH}" == "x" ]]; then
-      DOWNSTREAM_BRANCH="main"
+      DOWNSTREAM_BRANCH="default"
```
pkubatrh, line 46: Ah I see, the `DOWNSTREAM_BRANCH` variable is now set to `default`. this should probably get renamed to something like `CWT_CONFIG` then

pkubatrh: How about making the `DOWNSTREAM_BRANCH` variable required as it was in the previous version and based on its content decide which config to use?
Eg. DOWNSTREAM_BRANCH=f34 -> f34.yaml, DOWNSTREAM_BRANCH=rhel-8.3.0 -> rhel8.yaml:rhel8.3

phracek: I agree with renaming to CWT_CONFIG and for Fedora's case, we can use your proposal.
For the RHEL world, the user should know what he is doing and only check if the configuration is valid, like
`rhel8.yaml:rhel8.3` double dot is present and `rhel7` or `rhel8`.yaml file exists.

@pkubatrh Does it make sense to you?

pkubatrh: >For the RHEL world, the user should know what he is doing and only check if the configuration is valid, like
rhel8.yaml:rhel8.3 double dot is present and rhel7 or rhel8.yaml file exists.

I would expect the user to be a packager and know what branch they want to sync with. I would not necessarily expect the user to know what cwt config set to use. If we could transform the branch name to config name, it would be better from the user experience point of view.

betka.sh
phracek: Hopefully, I have addressed this issue by function `convert_branch_to_cwt_tool`

betka.sh
phracek: The function `convert_branch_to_cwt_tool` should do the magic

phracek:
@pkubatrh WDYT? Does it make sense to you? I have tested locally and it works properly.

phracek:
[test]

phracek:
[test]

phracek:
New design PR i here: https://github.com/sclorg/container-common-scripts/pull/200

PR #198, opened 2021-06-01
Pull image before tests execution.

phracek:
Fail in case of error.

Required-by: s2i-python-container#449

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -839,10 +839,11 @@ function ct_os_test_template_app_func() {
       for i_t in ${other_images} ; do
         echo "${i_t}"
         IFS='|' read -ra image_tag_a <<< "${i_t}"
-        if [[ "$(docker images -q "$image_name" 2>/dev/null)" != "" ]]; then
+        if [[ "$(docker images -q "$image_name" 2>/dev/null)" == "" ]]; then
           echo "ERROR: Image $image_name is not pulled yet."
+          docker images
           echo "Add to the beginning of scripts run-openshift-remote-cluster and run-openshift row"
-          echo "'ct_pull_image $image_name'."
+          echo "'ct_pull_image $image_name true'."
           exit 1
         fi
 
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -839,7 +839,13 @@ function ct_os_test_template_app_func() {
       for i_t in ${other_images} ; do
         echo "${i_t}"
         IFS='|' read -ra image_tag_a <<< "${i_t}"
-        docker pull "${image_tag_a[0]}"
+        if [[ "$(docker images -q "$image_name" 2>/dev/null)" != "" ]]; then
+          echo "ERROR: Image $image_name is not pulled yet."
+          echo "Add to the beginning of scripts run-openshift-remote-cluster and run-openshift row"
+          echo "'ct_pull_image $image_name'."
+          exit 1
+        fi
+
         if [ "${CT_EXTERNAL_REGISTRY:-false}" == 'true' ] ; then
           ct_os_import_image_ocp4 "${image_tag_a[0]}" "${image_tag_a[1]}"
         else
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -83,7 +83,7 @@ function ct_pull_image() {
     ((loop++)) || :
     echo "Pulling image $image_name failed."
     if [ "$loop" -gt "$loops" ]; then
-      echo "It happened $loops times. Giving up."
+      echo "Pulling of image $image_name failed $loops times in a row. Giving up."
       echo "!!! ERROR with pulling image $image_name !!!!"
       return 1
     fi
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -56,12 +56,14 @@ function ct_enable_cleanup() {
 # -------------
 # Function pull an image before tests execution
 # Argument: image_name - string containing the public name of the image to pull
+# Argument: exit - in case "true" is defined and pull failed, then script has to exit with 1 and no tests are executed
+# Argument: loops - how many times to pull image in case of failure
 # Function returns either 0 in case of pull was successful
 # Or the test suite exit with 1 in case of pull error
 function ct_pull_image() {
   local image_name="$1"; shift
+  local exit=${1:-"false"}; shift
   local loops=${1:-10}; shift
-  local exit=${1:-"false"}
   local loop=0
 
   # Let's try to pull image.
@@ -70,11 +72,7 @@ function ct_pull_image() {
   # Check if the image is available locally and try to pull it if it is not
   if [[ "$(docker images -q "$image_name" 2>/dev/null)" != "" ]]; then
     echo "The image $image_name is already pulled."
-    if [[ x"$exit" == "xfalse" ]]; then
-      return 0
-    else
-      exit 0
-    fi
+    return 0
   fi
 
   # Try pulling the image to see if it is accessible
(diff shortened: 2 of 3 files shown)


phracek:
[test]

test-lib.sh
```diff
@@ -52,6 +52,42 @@ function ct_enable_cleanup() {
...
+
+  # Try pulling the image to see if it is accessible
+  # WORKAROUND: Since Fedora registry sometimes fails randomly, let's try it more times
+  while ! docker pull "$image_name"; do
+    ((loop++)) || :
+    echo "Pulling image $image_name failed."
+    if [ "$loop" -gt "$loops" ]; then
+      echo "It happened $loops times. Giving up."
+      echo "!!! ERROR with pulling image $image_name !!!!"
+      return 1
```
phracek, line 83: This return value can be configurable and by default `return 1`. Otherwise, `exit 1` in case of checking image availability before tests execution.

The function itself can be used in all `container-common-scripts`.

test-lib.sh
```diff
@@ -765,20 +801,6 @@ EOF
...
-ct_check_image_availability() {
-  local public_image_name=$1;
-
-  # Try pulling the image to see if it is accessible
-  if ! docker pull "$public_image_name" &>/dev/null; then
-    echo "$public_image_name could not be downloaded via 'docker'"
-    return 1
-  fi
-}
-
```
pkubatrh, line 781: Can we keep the function and just call `ct_pull_image` inside? I would rather not remove the function completely in case it is used in some of the repositories

test-lib.sh
```diff
@@ -52,6 +52,42 @@ function ct_enable_cleanup() {
...
+# Argument: image_name - string containing the public name of the image to pull
+# Function returns either 0 in case of pull was successful
+# Or the test suite exit with 1 in case of pull error
+function ct_pull_image() {
+  local image_name="$1"
+  local loops=10
+  local loop=0
+
+  # Let's try to pull image.
+  echo "-> Pulling image $image_name before tests execution."
```
pkubatrh, line 67: I would remove the `before tests execution.` part from the output. It might confuse the reader in case the function is actually not used before executing tests.

test-lib.sh
```diff
@@ -52,6 +52,42 @@ function ct_enable_cleanup() {
...
 
+# ct_pull_image
+# -------------
+# Function pull an image before tests execution
+# Argument: image_name - string containing the public name of the image to pull
+# Function returns either 0 in case of pull was successful
+# Or the test suite exit with 1 in case of pull error
+function ct_pull_image() {
+  local image_name="$1"
+  local loops=10
```
pkubatrh, line 63: How about making it possible to set the amount of loops via function argument

test-lib-openshift.sh
```diff
@@ -779,7 +779,6 @@ function ct_os_test_template_app_func() {
       for i_t in ${other_images} ; do
         echo "${i_t}"
         IFS='|' read -ra image_tag_a <<< "${i_t}"
-        docker pull "${image_tag_a[0]}"
```
pkubatrh, line 782: Just wondering if this change is expected as it does not seem to be related to the rest of the changes

test-lib.sh
phracek: Sure. I will do it.

test-lib.sh
phracek: I will do it so, that If the parameter is not specified, then used default.

test-lib-openshift.sh
phracek: Good point. I wanted to remove docker pull, because of `docker pull` or especially pulling image has to be done at the beginning like here: https://github.com/sclorg/s2i-python-container/pull/449/files#diff-0b78990f6549655ad2c4497f3e0b771c74d42fc70837c7e1222728ae2cd566a4R20

But maybe we can add there a check if an image does not exist, then failed and say user to add ct_pull_image at the beginning of tests.

phracek: See me proposal if it make sense to you.

phracek:
[test]

phracek:
Failed because of docker hub rate limit.

[test]

phracek:
[test]

phracek:
[test]

phracek:
[test]

phracek:
@pkubatrh Can you please look at this pull request once more? It looks like it help us with testing on OpenShift 3. See error here: http://artifacts.dev.testing-farm.io/54a6352b-9b53-4bdd-9f1f-2200aef34e89/

phracek:
[test]

phracek:
Rebased

[test]

phracek:
PR is blocked by https://github.com/sclorg/container-common-scripts/pull/203

phracek:
Rebased

[test]

phracek:
[test]

phracek:
[test]

test-lib-openshift.sh
pkubatrh: Makes sense, thanks!

test-lib.sh
```diff
@@ -52,6 +52,47 @@ function ct_enable_cleanup() {
   trap ct_cleanup EXIT SIGINT
 }
 
+# ct_pull_image
+# -------------
+# Function pull an image before tests execution
+# Argument: image_name - string containing the public name of the image to pull
```
pkubatrh, line 58: Can you add docstrings for the new arguments as well?

test-lib.sh
```diff
@@ -52,6 +52,47 @@ function ct_enable_cleanup() {
...
+  local exit=${1:-"false"}
+  local loop=0
+
+  # Let's try to pull image.
+  echo "-> Pulling image $image_name ..."
+  # Sometimes in Fedora case it fails with HTTP 50X
+  # Check if the image is available locally and try to pull it if it is not
+  if [[ "$(docker images -q "$image_name" 2>/dev/null)" != "" ]]; then
+    echo "The image $image_name is already pulled."
+    if [[ x"$exit" == "xfalse" ]]; then
```
pkubatrh, line 73: I do not see any calls to `ct_pull_image` with `exit` set. Do you already have some use case in mind for future calls where this can be useful to have?

test-lib.sh
```diff
@@ -52,6 +52,47 @@ function ct_enable_cleanup() {
...
+    fi
+  fi
+
+  # Try pulling the image to see if it is accessible
+  # WORKAROUND: Since Fedora registry sometimes fails randomly, let's try it more times
+  while ! docker pull "$image_name"; do
+    ((loop++)) || :
+    echo "Pulling image $image_name failed."
+    if [ "$loop" -gt "$loops" ]; then
+      echo "It happened $loops times. Giving up."
```
pkubatrh, line 86: I owuld suggest a bit more verbose error message here.

```suggestion
      echo "Pulling of image $image_name failed $loops times in a row. Giving up."
```

test-lib.sh
phracek: This is a great catch. I have a mistake in the script. I will update the function.

test-lib-openshift.sh
```diff
@@ -825,7 +825,13 @@ function ct_os_test_template_app_func() {
...
+        if [[ "$(docker images -q "$image_name" 2>/dev/null)" != "" ]]; then
```
phracek, line 828: In case the image is already pulled, then this condition is wrong.

It should be `if [[ "$(docker images -q "$image_name" 2>/dev/null)" == "" ]]; then`

phracek:
[test]

pkubatrh:
Seems like the failing tests are because one of the images used in tests cannot be built since a Dockerfile is not available: https://github.com/sclorg/container-common-scripts/tree/master/tests/failures/check/v0

phracek:
[test]

phracek:
[test]

test-lib.sh
```diff
@@ -185,7 +228,7 @@ function ct_assert_container_creation_fails() {
 function ct_create_container() {
   local cid_file="$CID_FILE_DIR/$1" ; shift
   # create container with a cidfile in a directory for cleanup
-  # shellcheck disable=SC2086
+  # shellcheck disable=SC2086,SC2153
```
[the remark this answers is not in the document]
phracek: The reason is `^---------^ SC2153: Possible misspelling: IMAGE_NAME may not be assigned, but image_name is.`

phracek:
[test]

phracek:
[test]

PR #200, opened 2021-06-14
Add support 'make betka'

phracek:
This PR adds new functionality to generate downstream sources based on the upstream sources.

What is covered by this pull request:
- [x] centos7 is not supported and checked in the scripts. There is no plan to support it.
- [x] fedora is supported. If DOWNSTREAM_BRANCH is not specified, then `main` is default. `master` does not exist at all. DOWNSTREAM_NAME has to be specified.
- [x] rhel is supported. DOWNSTREAM_NAME has to be specified because we don't know what to generate. DOCKER_IMAGE has to be specified. Like `rh-nodejs10`. DOWNSTREAM_BRANCH has to be specified, like `rhel-8.4.0` or `rhscl-3.6-rhel7`
- [x] tests are covered also for `make betka`
- [x] Documentation updated

There is a PR https://github.com/sclorg/container-workflow-tool/pull/28 which fixes correct Dockerfile tag generation.

### DIFF
--- a/betka.py
+++ b/betka.py
@@ -26,12 +26,12 @@ import os
 import subprocess
 import sys
 import shutil
-
 from typing import List
 from pathlib import Path
 from contextlib import contextmanager
 from os import getenv
 from tempfile import TemporaryDirectory
+from collections import namedtuple
 
 
 def run_cmd(cmd, return_output=False, ignore_error=False, shell=False, **kwargs):
@@ -102,18 +102,32 @@ class BetkaGenerator(object):
 
     def check_requirements(self) -> bool:
         if self.os_env == "centos7":
-            msg = f"Target has to be specified.\n"\
-                  "CentOS7 is not supported.\n"\
-                  "e.g. make betka TARGET=fedora VERSIONS=XY\n"
+            msg = """
+'centos7' target is not supported.
+Target has to be e.g. rhel7, fedora or rhel8.
+e.g. make betka TARGET=fedora VERSIONS=XY
+            """
             print(msg)
             return False
         if self.cwt_docker_image == "":
-            msg = f"Docker image for generating dist-git sources for RHEL has to be specified.\n"\
-                  "Ask pkubat@redhat.com, hhorak@redhat.com or phracek@redhat.com for the name."
+            msg = """
+Docker image for generating dist-git sources for RHEL has to be specified.
+Ask pkubat@redhat.com, hhorak@redhat.com or phracek@redhat.com for the name.
+            """
             print(msg)
             return False
         if self.os_env == "fedora":
+            msg = f"""
+For generating dist-git sources to Fedora {self.cwt_docker_image} is used.
+'make betka' will try to sync main branch in Fedora land.
+            """
+            print(msg)
             if not self.downstream_branch:
+                msg = """
+If you want to change it then specify it by parameter DOWNSTREAM_BRANCH,
+like 'make betka TARGET=fedora DOWNSTREAM_BRANCH=f32'
+                """
+                print(msg)
                 self.cwt_config = "default.yaml"
             else:
                 # Examples config for F34 is f34.yaml
@@ -122,20 +136,24 @@ class BetkaGenerator(object):
                 self.cwt_config = f"{self.downstream_branch}.yaml"
         else:
             if not self.downstream_branch:
-                msg = f"DOWNSTREAM_BRANCH is not specified.\n" \
-                      f"Examples:\n" \
-                      f"rhel-8.3.0 for RHEL8\n" \
-                      f"rhscl-3.6-rhel7 for RHEL7\n" \
-                      f"For more details ask pkubat@redhat.com or phracek@redhat.com\n"
+                msg = """
+DOWNSTREAM_BRANCH is not specified.
+Examples:
+rhel-8.3.0 for RHEL8
+rhscl-3.6-rhel7 for RHEL7
+For more details ask pkubat@redhat.com or phracek@redhat.com
+                """
                 print(msg)
                 return False
             self.cwt_command = "rhcwt"
-        git_config = Path(pathlib.Path.home()) / ".gitconfig"
+        git_config = Path.home() / ".gitconfig"
         if not git_config.exists():
-            msg = f"File {git_config} is mandatory for using CWT tool." \
-                  f"Create it by commands:" \
-                  f"git config --global user.mail <your mail>" \
-                  f"git config --global user.name <your name>"
+            msg = f"""
+File {git_config} is mandatory for using CWT tool.
+Create it by commands:
+git config --global user.mail <your mail>
+git config --global user.name <your name>
+            """
             print(msg)
             return False
         return True
@@ -143,7 +161,7 @@ class BetkaGenerator(object):
     def delete_generated_dirs(self, ver: str):
         results_dir: Path = self.cur_dir / f"results-{ver}"
         if results_dir.exists():
-            print(f"Results dir {results_dir} already exists. Delete it.")
+            print(f"results-{ver} dir was previously generated. Delete it to generate new sources.")
             shutil.rmtree(results_dir)
 
     def get_valid_images(self, ver) -> List[str]:
@@ -153,44 +171,32 @@ class BetkaGenerator(object):
               f" --config={self.cwt_config} utils listupstream'"
         docker_output = run_cmd(cmd, return_output=True, shell=True)
         valid_images = []
-        for x in docker_output.split('\n'):
+        for line in docker_output.split('\n'):
             # Fields are:
             # python-27 python-27 s2i-python-container https://github.com/sclorg/s2i-python-container.git 2.7 rhel-8.4.0
-            if self.upstream_image_name in x and ver == x.split(' ')[4]:
-                valid_images.append(x)
+            if self.upstream_image_name in line and ver == line.split(" ")[4]:
+                valid_images.append(line)
         print(f"Valid images {valid_images}")
         return valid_images
 
     def convert_branch_to_cwt_tool(self):
         fields = self.downstream_branch.split('-')
         if self.os_env == "rhel7":
-            self.cwt_config = f"rhel7.yaml:{fields[0]}{fields[1].replace('.','')}"
+            self.cwt_config = f"rhel7.yaml:{fields[0]}{fields[1].replace('.','')}0"
         else:
             release_fields = fields[1].split('.')
             self.cwt_config = f"rhel8.yaml:{fields[0]}{release_fields[0]}.{release_fields[1]}"
 
-    def check_parameters(self):
-        if self.os_env == "fedora":
-            msg = f"For generating dist-git sources to Fedora {self.cwt_docker_image} is used." \
-                  f"'make betka' will try to sync main branch in Fedora land." \
-                  f"If you want to change it then specify it by parameter DOWNSTREAM_BRANCH," \
-                  f"like 'make betka TARGET=fedora DOWNSTREAM_BRANCH=f32'"
-            print(msg)
-        if self.os_env == "centos7":
-            return 1
-
     def clone_and_switch_to_branch(self, downstream_name: str, branch_name: str):
-        cmd = f"git clone {self.clone_url}/{downstream_name} {self.betka_tmp_dir.name}/results"
+        cmd = f"git clone --branch {branch_name} {self.clone_url}/{downstream_name} {self.betka_tmp_dir.name}/results"
         run_cmd(cmd, shell=True)
-        with cwd(f"{self.betka_tmp_dir.name}/results"):
-            cmd = f"git checkout {branch_name}"
-            run_cmd(cmd, shell=True)
 
     def copy_upstream_sources(self):
         p = Path(f"{self.betka_tmp_dir.name}/{self.upstream_image_name}")
-        if not p.exists():
-            print(f"Upstream sources are copied to {p}")
-            shutil.copytree(self.cur_dir, f"{p}", symlinks=True)
+        if p.exists():
+            shutil.rmtree(p)
+        print(f"Upstream sources are copied to {p}")
+        shutil.copytree(self.cur_dir, f"{p}", symlinks=True)
 
     def generate_sources(self, downstream_name: str):
         cmd = f"docker run -it --rm -v {Path.home()}/.gitconfig:/root/.gitconfig:ro,Z " \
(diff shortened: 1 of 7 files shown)


phracek:
[test]

frenzymadness (commented):
I don't have enough time to finish the review now so sending what I have.

In general, what is your motivation to use type annotations if there is no (implemented) way to check them?
For example, you are using them to annotate obvious types like `self.cur_dir: Path = Path.cwd()` but some of the methods like `check_parameters` are not annotated at all and some of them have annotated arguments but not the return value like `clone_and_switch_to_branch`.
What about using mypy to check the types if you want to use them?

betka.py
```diff
@@ -0,0 +1,233 @@
...
+        if self.os_env != "fedora":
+            self.convert_branch_to_cwt_tool()
+        cmd = f"docker run -it --rm {self.cwt_docker_image} bash -c '{self.cwt_command}" \
+              f" --config={self.cwt_config} utils listupstream'"
+        docker_output = run_cmd(cmd, return_output=True, shell=True)
+        valid_images = []
+        for x in docker_output.split('\n'):
+            # Fields are:
+            # python-27 python-27 s2i-python-container https://github.com/sclorg/s2i-python-container.git 2.7 rhel-8.4.0
+            if self.upstream_image_name in x and ver == x.split(' ')[4]:
```
frenzymadness, line 160: Also, it might make sense to use the same quotation marks everywhere.

betka.py
```diff
@@ -0,0 +1,233 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import os
+import pathlib
+import subprocess
+import sys
+import shutil
+
```
frenzymadness, line 30: This is just a nitpick. Imports in Python are usually organized like this:
```
imports from stdlib

imports from third-party packages

imports from the same project
```
So, in this specific case, there should be no empty line.

frenzymadness (commented):
Two more comments from a Pythonista.

betka.py
```diff
@@ -0,0 +1,233 @@
...
+                  f"git config --global user.mail <your mail>" \
+                  f"git config --global user.name <your name>"
+            print(msg)
+            return False
+        return True
+
+    def delete_generated_dirs(self, ver: str):
+        results_dir: Path = self.cur_dir / f"results-{ver}"
+        if results_dir.exists():
+            print(f"Results dir {results_dir} already exists. Delete it.")
```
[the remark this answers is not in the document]
phracek: sure

betka.py
```diff
@@ -0,0 +1,233 @@
...
+            run_cmd(cmd, shell=True)
+
+    def copy_upstream_sources(self):
+        p = Path(f"{self.betka_tmp_dir.name}/{self.upstream_image_name}")
+        if not p.exists():
+            print(f"Upstream sources are copied to {p}")
+            shutil.copytree(self.cur_dir, f"{p}", symlinks=True)
+
+    def generate_sources(self, downstream_name: str):
+        cmd = f"docker run -it --rm -v {pathlib.Path.home()}/.gitconfig:/root/.gitconfig:ro,Z " \
```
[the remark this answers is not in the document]
phracek: Good catch.

betka.py
```diff
@@ -0,0 +1,233 @@
...
+        for ver in self.versions_env.split(' '):
+            with cwd(str(self.cur_dir / ver)):
+                self.delete_generated_dirs(ver)
+                valid_images = self.get_valid_images(ver)
+                for img in valid_images:
+                    self.betka_tmp_dir = TemporaryDirectory()
+                    img_fields = img.split()
+                    self.ups_sources = f"{self.betka_tmp_dir.name}/{self.upstream_image_name}"
+                    self.copy_upstream_sources()
+                    self.clone_and_switch_to_branch(downstream_name=img_fields[0], branch_name=img_fields[5])
```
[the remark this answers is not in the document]
phracek: Fixed by the commit https://github.com/sclorg/container-common-scripts/pull/200/commits/edaadfc2051a3d223077e09c717f4acd61084a0f

phracek:
@frenzymadness Thanks for the deep review.

phracek:
[test]

betka.py
```diff
@@ -0,0 +1,244 @@
...
+        else:
+            release_fields = fields[1].split('.')
+            self.cwt_config = f"rhel8.yaml:{fields[0]}{release_fields[0]}.{release_fields[1]}"
+
+    def clone_and_switch_to_branch(self, downstream_name: str, branch_name: str):
+        cmd = f"git clone --branch {branch_name} {self.clone_url}/{downstream_name} {self.betka_tmp_dir.name}/results"
+        run_cmd(cmd, shell=True)
+
+    def copy_upstream_sources(self):
+        p = Path(f"{self.betka_tmp_dir.name}/{self.upstream_image_name}")
```
frenzymadness, line 195: Might be more readable:
```suggestion
        p = Path(self.betka_tmp_dir.name) / self.upstream_image_name
```

betka.py
```diff
@@ -0,0 +1,244 @@
...
+        if p.exists():
+            shutil.rmtree(p)
+        print(f"Upstream sources are copied to {p}")
+        shutil.copytree(self.cur_dir, f"{p}", symlinks=True)
+
+    def generate_sources(self, downstream_name: str):
+        cmd = f"docker run -it --rm -v {Path.home()}/.gitconfig:/root/.gitconfig:ro,Z " \
```
frenzymadness, line 202: Multiline string?

betka.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def copy_generated_source(self, ver: str):
+        shutil.copytree(self.ups_sources, self.cur_dir / f"results-{ver}", symlinks=True)
+
+    def convert_sources(self):
+        generated_sources = []
+        if not self.versions_env:
+            return 1
+        for ver in self.versions_env.split():
+            with cwd(str(self.cur_dir / ver)):
```
frenzymadness, line 216: `os.chdir` accepts Path objects, no need to use string here.
```suggestion
            with cwd(self.cur_dir / ver):
```

betka.py
```diff
@@ -0,0 +1,244 @@
...
+                self.delete_generated_dirs(ver)
+                valid_images = self.get_valid_images(ver)
+                # ['rh-nodejs14 nodejs-14 s2i-nodejs-container https://github.com/sclorg/s2i-nodejs-container.git 14 rhscl-3.6-rhel-7']
+                ImageInfo = namedtuple(
```
frenzymadness, line 220: ImageInfo is something like class so it should be defined outside of this method on the top level of the module.

betka.py
```diff
@@ -0,0 +1,244 @@
...
+        yield
+    finally:
+        os.chdir(curdir)
+
+
+class BetkaGenerator(object):
+
+    def __init__(self):
+        self.betka_tmp_dir = None
+        self.os_env: str = getenv("OS")
```
frenzymadness, line 92: I'd use `os.getenv` like you already do for the rest of the functions from this module.
```suggestion
        self.os_env: str = os.getenv("OS")
```
Then, you can remove `from os import getenv`

betka.py
```diff
@@ -0,0 +1,244 @@
...
+                    self.clone_and_switch_to_branch(
+                        downstream_name=img_info.downstream_name, branch_name=img_info.branch
+                    )
+                    self.generate_sources(downstream_name=img_info.downstream_name)
+                    self.copy_generated_source(ver=img_info.ver)
+                    generated_sources.append(f"results-{ver}")
+                    self.betka_tmp_dir.cleanup()
+        if generated_sources:
+            print("Dist-git sources generated by 'make betka' are stored in this/these directories:")
+            print('\n'.join(generated_sources))
```
frenzymadness, line 237: Nitpick: mixed quotation marks.

frenzymadness (commented):
A few more comments from me.

In general:
* all the tests are negative, is there a plan for some positive tests?
* what about the type annotations? Is there a plan to check them? If not, they might not be up-to-date and we'll never know about it.
* the tool generates a lot of output, is this just for development and debugging?

But most important is: what is the result I should expect? I've tried to run this for Python container `make betka TARGET=fedora VERSIONS=3.9 DOWNSTREAM_BRANCH=f34` and I get 177 lines of output and a new folder appeared in my cwd - `results-3.9` - that seems correct but it contains basically a copy of the main repository folder with the following content: `2.7  3.4  3.6  3.8  auto_targets.mk  cwt_config  imagestreams  Makefile     README.md  src
3.3  3.5  3.7  3.9  common           examples    LICENSE       manifest.sh  specs      test`. Am I doing something wrong?

phracek:
> But most important is: what is the result I should expect? I've tried to run this for Python container `make betka TARGET=fedora VERSIONS=3.9 DOWNSTREAM_BRANCH=f34` and I get 177 lines of output and a new folder appeared in my cwd - `results-3.9` - that seems correct but it contains basically a copy of the main repository folder with the following content: `2.7 3.4 3.6 3.8 auto_targets.mk cwt_config imagestreams Makefile README.md src 3.3 3.5 3.7 3.9 common examples LICENSE manifest.sh specs test`. Am I doing something wrong?

The directory `results-3.9` should contain files for container mentioned in src.fedoraproject.org/container/python in branch f34. It looks like the generator did not generate files properly. I will look on it.

phracek:
>     * the tool generates a lot of output, is this just for development and debugging?

I will try to reduce the output from the user point of view. The tool itself is only for us. Or I will add logger with info and with debug options.

phracek:
>     * what about the type annotations? Is there a plan to check them? If not, they might not be up-to-date and we'll never know about it.

This would be fine to check them, Just let me know how and I will do it. :)

frenzymadness:
> > ```
> > * what about the type annotations? Is there a plan to check them? If not, they might not be up-to-date and we'll never know about it.
> > ```
> 
> This would be fine to check them, Just let me know how and I will do it. :)

The best tool for it is mypy. It's usually used as plugin for pytest but because our tests are in bash, we can use it as a standalone tool. Currently, there are no errors because a lot of type annotations are missing but when I use a strict mode which disallows untyped definitions, the output look like this:
```
mypy --pretty --strict betka.py
betka.py:37: error: Function is missing a type annotation
    def run_cmd(cmd, return_output=False, ignore_error=False, shell=False, **kwargs):
    ^
betka.py:73: error: Function is missing a type annotation
    def cwd(target):
    ^
betka.py:90: error: Function is missing a return type annotation
        def __init__(self):
        ^
betka.py:90: note: Use "-> None" if function does not return a value
betka.py:92: error: Incompatible types in assignment (expression has type "Optional[str]",
variable has type "str")
            self.os_env: str = os.getenv("OS")
                               ^
betka.py:93: error: Incompatible types in assignment (expression has type "Optional[str]",
variable has type "str")
            self.versions_env: str = getenv("VERSIONS")
                                     ^
betka.py:94: error: Incompatible types in assignment (expression has type "Optional[str]",
variable has type "str")
            self.cwt_docker_image: str = getenv("CWT_DOCKER_IMAGE")
                                         ^
betka.py:95: error: Incompatible types in assignment (expression has type "Optional[str]",
variable has type "str")
            self.downstream_branch: str = getenv("DOWNSTREAM_BRANCH")
                                          ^
betka.py:96: error: Incompatible types in assignment (expression has type "Optional[str]",
variable has type "str")
            self.clone_url: str = getenv("CLONE_URL")
                                  ^
betka.py:161: error: Function is missing a return type annotation
        def delete_generated_dirs(self, ver: str):
        ^
betka.py:167: error: Function is missing a type annotation for one or more arguments
        def get_valid_images(self, ver) -> List[str]:
        ^
betka.py:169: error: Call to untyped function "convert_branch_to_cwt_tool" in typed context
                self.convert_branch_to_cwt_tool()
                ^
betka.py:172: error: Call to untyped function "run_cmd" in typed context
            docker_output = run_cmd(cmd, return_output=True, shell=True)
                            ^
betka.py:182: error: Function is missing a return type annotation
        def convert_branch_to_cwt_tool(self):
        ^
betka.py:182: note: Use "-> None" if function does not return a value
betka.py:190: error: Function is missing a return type annotation
        def clone_and_switch_to_branch(self, downstream_name: str, branch_name: str):
        ^
betka.py:191: error: Item "None" of "Optional[Any]" has no attribute "name"
            cmd = f"git clone --branch {branch_name} {self.clone_url}/{downstream_name} ...
                  ^
betka.py:192: error: Call to untyped function "run_cmd" in typed context
            run_cmd(cmd, shell=True)
            ^
betka.py:194: error: Function is missing a return type annotation
        def copy_upstream_sources(self):
        ^
betka.py:194: note: Use "-> None" if function does not return a value
betka.py:195: error: Item "None" of "Optional[Any]" has no attribute "name"
            p = Path(self.betka_tmp_dir.name) / self.upstream_image_name
                     ^
betka.py:201: error: Function is missing a return type annotation
        def generate_sources(self, downstream_name: str):
        ^
betka.py:202: error: Item "None" of "Optional[Any]" has no attribute "name"
            cmd = f"docker run -it --rm -v {Path.home()}/.gitconfig:/root/.gitconfig:ro,...
                  ^
betka.py:206: error: Call to untyped function "run_cmd" in typed context
            run_cmd(cmd, shell=True)
            ^
betka.py:208: error: Function is missing a return type annotation
        def copy_generated_source(self, ver: str):
        ^
betka.py:211: error: Function is missing a return type annotation
        def convert_sources(self):
        ^
betka.py:227: error: Call to untyped function "copy_upstream_sources" in typed context
                        self.copy_upstream_sources()
                        ^
betka.py:241: error: Call to untyped function "BetkaGenerator" in typed context
        bg = BetkaGenerator()
             ^
betka.py:244: error: Call to untyped function "convert_sources" in typed context
        sys.exit(bg.convert_sources())
                 ^
Found 26 errors in 1 file (checked 1 source file)
```

betka.py
```diff
@@ -0,0 +1,244 @@
...
+    def check_requirements(self) -> bool:
+        if self.os_env == "centos7":
+            msg = """
+'centos7' target is not supported.
+Target has to be e.g. rhel7, fedora or rhel8.
+e.g. make betka TARGET=fedora VERSIONS=XY
+            """
+            print(msg)
+            return False
+        if self.cwt_docker_image == "":
```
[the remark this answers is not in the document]
phracek: > But most important is: what is the result I should expect? I've tried to run this for Python container `make betka TARGET=fedora VERSIONS=3.9 DOWNSTREAM_BRANCH=f34` and I get 177 lines of output and a new folder appeared in my cwd - `results-3.9` - that seems correct but it contains basically a copy of the main repository folder with the following content: `2.7 3.4 3.6 3.8 auto_targets.mk cwt_config imagestreams Makefile README.md src 3.3 3.5 3.7 3.9 common examples LICENSE manifest.sh specs test`. Am I doing something wrong?

There was a bug in the code.
Now dist-git generation should work properly.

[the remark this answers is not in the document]
phracek: Fixed

betka.py
phracek: Fixed

betka.py
phracek: Fixed

phracek:
@frenzymadness mypy issues were fixed. Also multiline strings.
Big bug with source code generation was fixed. The correct path is here https://github.com/sclorg/container-common-scripts/pull/200/files#diff-34dd7322bd4adc55dc3a2e0cb2c1576d271c37a866c83e198edca04e4790f66aR218

phracek:
[test]

frenzymadness (approved):
It seems to me that it works well. I've tested it on s2i-python and s2i-base containers. mypy checker is also happy. Do you plan to add mypy to tests?

phracek:
In the next round. I would like to close this as soon as possible. But thanks for the review.

https://github.com/sclorg/betka/blob/master/.pre-commit-config.yaml

phracek:
Issue filed https://github.com/sclorg/container-common-scripts/issues/202

PR #201, opened 2021-06-28
add s2i multistage build function

pkubatrh:
Required-by: s2i-nodejs-container#269

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -765,6 +765,83 @@ EOF
     )
 }
 
+# ct_s2i_multistage_build APP_PATH SRC_IMAGE DST_IMAGE SEC_IMAGE [S2I_ARGS]
+# ----------------------------
+# Create a new s2i app image from local sources in a similar way as source-to-image would have used.
+# Argument: APP_PATH - local path to the app sources to be used in the test
+# Argument: SRC_IMAGE - image to be used as a base for the s2i build process
+# Argument: SEC_IMAGE - image to be used as the base for the result of the build process
+# Argument: DST_IMAGE - image name to be used during the tagging of the s2i build result
+# Argument: S2I_ARGS - Additional list of source-to-image arguments.
+#                      Only used to check for environment variable definitions.
+ct_s2i_multistage_build() {
+
+  local app_path=$1; shift
+  local src_image=$1; shift
+  local sec_image=$1; shift
+  local dst_image=$1; shift
+  local s2i_args=$*;
+  local local_app="app-src"
+  local user_id=
+  local mount_options=
+
+
+  # Run the entire thing inside a subshell so that we do not leak shell options outside of the function
+  (
+  # Error out if any part of the build fails
+  set -e
+
+  user=$(docker inspect -f "{{.Config.User}}" "$src_image")
+  # Default to root if no user is set by the image
+  user=${user:-0}
+  # run the user through the image in case it is non-numeric or does not exist
+  # NOTE: The '-eq' test is used to check if $user is numeric as it will fail if $user is not an integer
+  if ! [ "$user" -eq "$user" ] 2>/dev/null && ! user_id=$(docker run --rm "$src_image" bash -c "id -u $user 2>/dev/null"); then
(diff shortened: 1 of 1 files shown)


pkubatrh:
[test]

test-lib.sh
```diff
@@ -765,6 +765,83 @@ EOF
...
+  fi
+
+  # Use /tmp to not pollute cwd
+  tmpdir=$(mktemp -d)
+  df_name=$(mktemp -p "$tmpdir" Dockerfile.XXXX)
+  cd "$tmpdir"
+
+  # Strip file:// from APP_PATH and copy its contents into current context
+  mkdir -p "$local_app"
+  cp -r "${app_path/file:\/\//}/." "$local_app"
```
frenzymadness, line 814: What about support for git repositories?

test-lib.sh
```diff
@@ -765,6 +765,83 @@ EOF
...
+
+  cat <<EOF >"$df_name"
```
frenzymadness, line 816: Is it a good idea to hardcode the Dockerfile here? With all I learned during preps of Python minimal container, I can imagine that Dockerfile can be provided to this function and just slightly updated as `ct_test_app_dockerfile` does now. Mainly because minimal containers usually need to install some additional packages and do some other hack and there is no way how to do it now with hardcoded dockerfile.

frenzymadness (commented):
I'm using the mulistage build in the latest Python minimal container so my feedback is based on that experience. I'd say that the function for multistage build can work in a similar way as `ct_test_app_dockerfile` just for two separated images.

test-lib.sh
pkubatrh: We can add an argument for passing an existing Dockerfile and modifying it as necessary

test-lib.sh
pkubatrh: We can definitely do that. Feels like something useful in general, so maybe create as a library function (if there is none available yet, have to check) and pull it in?

test-lib.sh
```diff
@@ -900,6 +900,31 @@ ct_show_resources()
...
 }
 
+# ct_clone_git_repository
+# -----------------------------
+# Argument: app_url - git URI pointing to a repository, supports "@" to indicate a different branch
+# Argument: app_dir - name of the directory to clone the repository into
+ct_clone_git_repository()
+{
+  local app_url=$1; shift
+  local app_dir=$1
```
phracek, line 915: I would add there a check if app_dir is not empty. For sure.

test-lib.sh
```diff
@@ -760,8 +760,7 @@ EOF
     mount_options=$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*' || true)
 
     # Run the build and tag the result
-    # shellcheck disable=SC2086
-    docker build $mount_options -f "$df_name" --no-cache=true -t "$dst_image" .
+    docker build ${mount_options:+"$mount_options"} -f "$df_name" --no-cache=true -t "$dst_image" .
```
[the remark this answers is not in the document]
phracek: Just use `"$mount_points"`

test-lib.sh
```diff
@@ -809,9 +809,15 @@ ct_s2i_multistage_build() {
   df_name=$(mktemp -p "$tmpdir" Dockerfile.XXXX)
   cd "$tmpdir"
 
-  # Strip file:// from APP_PATH and copy its contents into current context
-  mkdir -p "$local_app"
-  cp -r "${app_path/file:\/\//}/." "$local_app"
+  # If the app_path looks like an URL, clone it
+  # Otherwise let's try to handle it as a local path
+  if [[ "$app_path" == "https"* ]] ; then
+    ct_clone_git_repository "$app_path" "$local_app"
```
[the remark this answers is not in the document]
phracek: Sorry, by git clone operation, the dir will exist

pkubatrh:
Rebased and added changes based on review
[test]

pkubatrh:
Quick hack to see if this was the reason behind the failure in CI. Will follow up with a proper fix after tests are through.

pkubatrh:
[test]

PR #203, opened 2021-07-26
Add support for mypy

phracek:
This commit only adds support for mypy and fixes errors caught by black and mypy


phracek:
@frenzymadness PTAL

phracek:
#203

frenzymadness:
I like it! To make it 100% perfect, I'd like to include the pre-commit tests in the CI. Is it possible to run pre-commit checks in the CI and fail in any case they would change anything? This way, we can use it locally and also check that everybody will obey the same rules.

phracek:
make sense to me. I will try a way how to add it into `.travis.yml` or somewhere else.

frenzymadness:
I believe that the best way is to try it and see the result. It might be also possible to include the shellcheck in the pre-commit config.

phracek:
[test]

phracek:
@frenzymadness PTAL. I have added the test into .travis.yaml and Makefile. It should be now checked always.

phracek:
[test]

phracek:
[test]

Makefile
```diff
@@ -32,6 +32,11 @@ test: check
 shellcheck:
 	./run-shellcheck.sh `git ls-files *.sh`
 
+pre-commit-check:
+	pip3 install pre-commit
```
frenzymadness, line 36: I'd rather not install anything this way. Imagine you'd run `make pre-commit-check` under the root and the command installs pre-commit and all its 12 dependencies into your system site-packages - this might be a problem in case of any conflict with your system. I'd honestly just assume that the command is available and let it show an error when it isn't.

phracek: and what about if pre-commit does not exist in CI? This means, that pre-commit checker will not work.

frenzymadness: Yes, but you can install it in the same way you do it for shellcheck: https://github.com/sclorg/container-common-scripts/blob/82c7f58e9a293aaaa5fcf3f45ae51b7678646d28/.github/workflows/sanity-tests.yaml#L36

phracek:
[test]

phracek:
[test]

phracek:
[test]

phracek:
[test]

phracek:
[test]

phracek:
@frenzymadness PR was updated and ready for merge. PTAL and let me know if everything is OK.

frenzymadness (approved):
Looks good to me!

PR #207, opened 2021-08-18
Support local application in ct_test_app_dockerfile

frenzymadness:
If $app_url contains `.git`, it's cloned from the URI. If not and
it exists as a folder locally, it's just copied via cp to $app_dir.

I'm testing this in Python container when I need to test something "from dockerfile" but with a local test application we already have in the same repository.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -827,7 +827,7 @@ ct_show_resources()
 # -----------------------------
 # Argument: dockerfile - path to a Dockerfile that will be used for building an image
 #                        (must work with an application directory called 'app-src')
-# Argument: app_url - git URI with a testing application, supports "@" to indicate a different branch
+# Argument: app_url - git or local URI with a testing application, supports "@" to indicate a different branch
 # Argument: body_regexp - PCRE regular expression that must match the response body
 # Argument: app_dir - name of the application directory that is used in the Dockerfile
 # Argument: port - Optional port number (default: 8080)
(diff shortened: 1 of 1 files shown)


frenzymadness:
[test]

frenzymadness:
The error does not seem to be related and the same failure happens in the two latest PRs.

frenzymadness:
[test]

test-lib.sh
```diff
@@ -865,20 +865,25 @@ ct_test_app_dockerfile() {
   echo "Using this Dockerfile:"
   cat Dockerfile
 
-  # If app_url contains @, the string after @ is considered
-  # as a name of a branch to clone instead of the main/master branch
-  IFS='@' read -ra git_url_parts <<< "${app_url}"
+  if  [[ $app_url == *".git"* ]]; then
```
torsava, line 868: Are we ssure all git URI end with ".git"? I'm a bit sceptical. Why not switch it around, first test `[ -d "$app_url" ]`, and treat everything else as git URI as before? A git URI will never be an existing local directory.

torsava:
Looks good to me!

frenzymadness:
Testing the new version in Python PR.

test-lib.sh
frenzymadness: You are right, implemented.

frenzymadness:
Both use cases seem to work well, thanks for the review!

PR #208, opened 2021-08-20
Allow distgen to be skipped for a specific versions

frenzymadness:
In Python container, I need to maintain the latest minimal container manually instead of via distgen (until I find a way how to do it in distgen) and because distgen runs also in the CI, I need a way to skip it for a specified version.

Successfully tested in https://github.com/sclorg/s2i-python-container/pull/467


torsava:
I have thoroughly tested the bash regex, it works fine under any circumstances. The rest looks good as well, the idea makes sense and it's already tested with the Python minimal container.

LGTM.

PR #212, opened 2021-10-04
Add container-common-script tests for all TARGETs

phracek:

### DIFF
--- a/tests/remote-containers.sh
+++ b/tests/remote-containers.sh
@@ -8,7 +8,7 @@ for image in ${TESTED_IMAGES}; do
     IMAGES[$image]=master
 done
 
-OS=centos7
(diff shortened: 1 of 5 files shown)


tests/remote-containers.sh
```diff
@@ -8,7 +8,7 @@ for image in ${TESTED_IMAGES}; do
     IMAGES[$image]=master
 done
 
-OS=centos7
+test -n "${OS-}" || false 'make sure $OS is defined'
```
hhorak, line 11: Good idea to run for all, if it works.

phracek: Let's try it.

pkubatrh:
Glad to see we will be running tests for all OS variations, thanks for looking at this!

PR #213, opened 2021-10-13
Do not install pre-commit hooks automatically

frenzymadness:
We have agreed on today's meeting that `make pre-commit-check` should not install pre-commit hooks into a local dev git repository. Installing them is an unexpected and possibly unwanted side-effect that is not needed for a one-time check run.

Instead, I've implemented a simple check whether they are already installed or not, and if not, there is a note about how to install them to run the check automatically before every commit.


pkubatrh:
lgtm, thanks!

PR #214, opened 2021-10-18
Support testing in CVP

phracek:
This pull request adds support for testing our containers in CVP pipeline.

The new variable is defined called `CVP=1`.
The test alone is executed from test/test-openshift.yaml playbook if file test/run-openshift-remote-cluster exists.

We will NOT create a project. This is already done by CVP.

The parameters to the script are:
```
VERSION={{ environment[0]['VERSION'] }} IMAGE_NAME={{ image_name }} OS={{ environment[0]['OS'] }} CVP=1 bash test/run-openshift-remote-cluster  | tee {{ cvp_artifacts_dir }}/{{ image_name }}.log
```

We do NOT build the image itself. It is provided and already uploaded into OCP4 cluster by CVP. Therefore it is specified as a parameter into `test/run-openshift-remote-cluster` script. The same as VERSION and OS.

### DIFF
--- a/betka.sh
+++ b/betka.sh
@@ -8,7 +8,8 @@
 set -ex
 
 function load_configuration() {
-  if [[ x"${OS}" == "xfedora" ]]; then
+  # shellcheck disable=SC2268
+  if [[ x"$OS" == "xfedora" ]]; then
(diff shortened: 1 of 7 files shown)


phracek:
[test]

phracek:
[test]

PR #218, opened 2021-11-29
Check if files generated by dist-gen are generated properly

phracek:
This new function checks if files generated by dist-gen are committed into each pull request properly.
The function is called only if the `manifest.sh` file is present in the root directory of a container.

The command which checks status of generated files is:
```
git diff --exit-code -- ':!common'
```
the common submodule is not checked at all.

Nowadays we have three containers that use dist-gen generated sources:
* varnish-container
* postgresql-container
* s2i-python-container


phracek:
[test]

test-lib.sh
```diff
@@ -1069,4 +1069,33 @@ ct_test_app_dockerfile() {
...
+    return 0
+  fi
+
+  echo "Generating all files by 'make generate-all'"
+  make generate-all
+  # We do not want to check container-common-script directory
+  if git diff --exit-code -- ':!common'; then
+    echo "Pull request does not contain properly generated files."
+    echo "Please run these command in order to update pull request:"
+    echo "'make generate-clean'"
```
frenzymadness, line 1093: Are you sure about `generate-clean` we don't have it in Python container.

frenzymadness (commented):
No idea why tests failed, the log is truncated unfortunately.

test-lib.sh
phracek: Updated with 'make clean-version', In my mind, I have 'make generate-clean' for some reason.

phracek:
[test]

phracek:
Tests are failing because of pino in s2i-nodejs-container. See some comments in issue https://github.com/sclorg/s2i-nodejs-container/pull/292

phracek:
[test]

pkubatrh:
I do not think it is necessary to have it defined as a function now right?

phracek:
> I do not think it is necessary to have it defined as a function now right?

You are right, it is not necessary. My idea was to have it in the function for the future time. But it can be removed.

phracek:
[test]

phracek:
> Also be aware that you still have `Required-by: postgresql-container#428` in the first commit and that PR is still using the function, so the tests will fail.

Removed. Thanks for point.

phracek:
@pkubatrh PTAL one more time.

phracek:
[test]

test.sh
```diff
@@ -8,6 +8,23 @@
 
 set -e
 
+if [[ -f "./manifest.sh" ]]; then
+  echo "Files are generated by dist-gen. Check if everything generated by distgen properly."
+  echo "Generating all files by 'make generate-all' ..."
+  make generate-all
+  echo "Generated files are generated."
+  # common submodule is not updated at all. No argument to git diff is needed.
```
pkubatrh, line 16: Wondering about this comment - if I understand it correctly then the `git diff` command would show a diff for the common submodule if it was updated right? We do update the common submodule for PRs made against container-common-scripts that use the `Required-by` keyword in the commit message. Would this work with them as well?

phracek: I don't think so. We do not do git pull in a submodule. We don't care about the submodule. This would be another test to test if the submodule is updated or not.
But I have found out the command which can be used for it.

`git diff --ignore-submodules=all`

phracek: Example:
```bash
$ git diff --ignore-submodules --exit-code
$ echo $?
0
$ git diff --exit-code
diff --git a/common b/common
index ce7b9d1..77d1ea2 160000
--- a/common
+++ b/common
@@ -1 +1 @@
-Subproject commit ce7b9d1aa48cf65bb8071254e4e5396de410904b
+Subproject commit 77d1ea2246524f5849bc088e58d967739f1cb681
$ echo $?
1
$
```

Issue #195, opened 2021-05-04
Start and stop OpenShift 3 properly


phracek:
The RHSCL CI is blocked when OpenShift 3.X does not start up.

We have to investigate how to start OpenShift 3. properly so tests are not blocked.

See.

phracek:
```bash
10:05:53 Checking if image openshift/origin-control-plane:v3.11 is available ...
10:05:53 Starting OpenShift using openshift/origin-control-plane:v3.11 ...
10:05:53 I0505 09:05:53.765101   25620 config.go:40] Running "create-master-config"
10:05:53 I0505 09:05:58.221202   25620 config.go:46] Running "create-node-config"
10:05:58 I0505 09:05:59.996691   25620 flags.go:30] Running "create-kubelet-flags"
10:06:00 I0505 09:06:01.017157   25620 run_kubelet.go:49] Running "start-kubelet"
10:06:01 I0505 09:06:01.505673   25620 run_self_hosted.go:181] Waiting for the kube-apiserver to be ready ...
10:06:01 I0505 09:06:26.543949   25620 interface.go:26] Installing "kube-proxy" ...
10:06:26 I0505 09:06:26.545632   25620 interface.go:26] Installing "kube-dns" ...
10:06:26 I0505 09:06:26.545663   25620 interface.go:26] Installing "openshift-service-cert-signer-operator" ...
10:06:26 I0505 09:06:26.545690   25620 interface.go:26] Installing "openshift-apiserver" ...
10:06:26 I0505 09:06:26.545750   25620 apply_template.go:81] Installing "openshift-apiserver"
10:06:26 I0505 09:06:26.545746   25620 apply_template.go:81] Installing "kube-proxy"
10:06:26 I0505 09:06:26.545831   25620 apply_template.go:81] Installing "kube-dns"
10:06:26 I0505 09:06:26.545878   25620 apply_template.go:81] Installing "openshift-service-cert-signer-operator"
10:06:26 I0505 09:06:29.801824   25620 interface.go:41] Finished installing "kube-proxy" "kube-dns" "openshift-service-cert-signer-operator" "openshift-apiserver"
10:06:29 Error: timed out waiting for the condition
10:11:29 error: Missing or incomplete configuration info.  Please login or point to an existing, complete config file:
10:11:30 
10:11:30   1. Via the command-line flag --config
10:11:30   2. Via the KUBECONFIG environment variable
10:11:30   3. In your home directory as ~/.kube/config
```

[...]

PR #220, opened 2021-12-01
Fixes: #195 turn of set +e while 'oc cluster down'

phracek:
Turn of 'set +e' while oc cluster down and
switch user to system:admin

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -530,8 +530,9 @@ function ct_os_cluster_up() {
 # Shuts down the local OpenShift cluster using 'oc cluster down'
 function ct_os_cluster_down() {
   if [ ${OS_CLUSTER_STARTED_BY_TEST:-0} -eq 1 ] ; then
-    echo "Cluster started by the test, shutting down."
+    echo "Switching user to system:admin before cluster is going down."
     oc login -u system:admin
+    echo "Cluster started by the test, shutting down."
     oc cluster down
   else
     echo "Cluster not started by the test, shutting down skipped."
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -531,6 +531,7 @@ function ct_os_cluster_up() {
 function ct_os_cluster_down() {
   if [ ${OS_CLUSTER_STARTED_BY_TEST:-0} -eq 1 ] ; then
     echo "Cluster started by the test, shutting down."
+    oc login -u system:admin
     oc cluster down
   else
     echo "Cluster not started by the test, shutting down skipped."


phracek:
The question is: will this work together with https://github.com/sclorg/s2i-python-container/blob/master/test/run-openshift#L18? From my bash knowledge point of view it should work.

pkubatrh:
The `oc login` addition makes sense to me if it is needed. But I do not think we want to ignore potential errors. What if the change does not help? Would we know if it does not fail the tests?

zmiklank:
@pkubatrh Wouldn't the potential errors be still present in the logs, as "set +e" only disables exit on error?

pkubatrh:
@zmiklank Yes, but do you ever check the logs for tests runs that did not fail?

phracek:
@pkubatrh @zmiklank Well, basically, `ct_os_cluster_down` is called at the end of each OpenShift 3 test suite.

See here:
https://github.com/sclorg/s2i-php-container/blob/master/test/run-openshift-local-cluster#L100
https://github.com/sclorg/postgresql-container/blob/master/test/run-openshift-local-cluster#L465

In most cases, our OpenShift 3 tests are failing for this reason: https://github.com/sclorg/container-common-scripts/issues/195#issuecomment-896638997

What I am missing in the function is something like the log, which will tell us more. Something like "Cluster is going down" and then call `oc cluster down`.
More logs can be added by this pull request. I will add it.

```bash
No resources found.
Error from server (Forbidden): pods is forbidden: User "developer" cannot list pods in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): replicationcontrollers is forbidden: User "developer" cannot list replicationcontrollers in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): services is forbidden: User "developer" cannot list services in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): daemonsets.apps is forbidden: User "developer" cannot list daemonsets.apps in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): deployments.apps is forbidden: User "developer" cannot list deployments.apps in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): replicasets.apps is forbidden: User "developer" cannot list replicasets.apps in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): statefulsets.apps is forbidden: User "developer" cannot list statefulsets.apps in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): horizontalpodautoscalers.autoscaling is forbidden: User "developer" cannot list horizontalpodautoscalers.autoscaling in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): jobs.batch is forbidden: User "developer" cannot list jobs.batch in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): cronjobs.batch is forbidden: User "developer" cannot list cronjobs.batch in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): deploymentconfigs.apps.openshift.io is forbidden: User "developer" cannot list deploymentconfigs.apps.openshift.io in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): buildconfigs.build.openshift.io is forbidden: User "developer" cannot list buildconfigs.build.openshift.io in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): builds.build.openshift.io is forbidden: User "developer" cannot list builds.build.openshift.io in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): imagestreams.image.openshift.io is forbidden: User "developer" cannot list imagestreams.image.openshift.io in the namespace "sclorg-test-15517": no RBAC policy matched
Error from server (Forbidden): routes.route.openshift.io is forbidden: User "developer" cannot list routes.route.openshift.io in the namespace "sclorg-test-15517": no RBAC policy matched
OpenShift tests for quay.io/centos7/mariadb-103-centos7:1 failed.
make: *** [test-openshift] Error 1
```

phracek:
> @zmiklank Yes, but do you ever check the logs for test runs that did not fail?

I don't follow this. If the test is failed, before oc_cluster_down, this function is called https://github.com/sclorg/s2i-php-container/blob/master/test/run-openshift-local-cluster#L21 or this one https://github.com/sclorg/postgresql-container/blob/master/test/run-openshift-local-cluster#L19.
Only if all tests are passing the function oc_cluster_down is called.

At the beginning of each OpenShift 3 tests `set -eo` is called like here:
https://github.com/sclorg/s2i-php-container/blob/master/test/run-openshift-local-cluster#L19
https://github.com/sclorg/postgresql-container/blob/master/test/run-openshift-local-cluster#L17

pkubatrh:
@phracek I understand what the problem is, but what do we want to fix with this PR? Do we want to fix the underlying problem (adding `oc login` so that `oc cluster down` does not fail), or do we just not care if `oc cluster down` fails (hence the `set +e` so that the test suite does not end in failure)?
The PR currently tries to do both and it seems to me that this contradicts each other - we should either ignore the failure if we do not care, or make sure it does not happen.

pkubatrh:
If we do both, then it implies we care about both. But we will never know the next time `oc cluster down` fails again for some different reason, because the failure is ignored.

phracek:
[test]

phracek:
@pkubatrh I have change this PR only for logging as admin. WDYT?

pkubatrh:
lgtm, guess the only way to test it out now is to merge it. :)

PR #222, opened 2021-12-09
Fix ct_pull_image. The fourth parameter is not used at all

phracek:
`ct_pull_image <image> <true|false> <count>`
count is optional.

Fourth parameter is not used at all.
When the bash is trying to shift, then it failed with exit 1.
```bash
VERSIONS="2.7 3.6 3.8 3.9 3.9-minimal" BASE_IMAGE_NAME="python" SKIP_SQUASH=0 UPDATE_BASE= OS=rhel7 CLEAN_AFTER= DOCKER_BUILD_CONTEXT=.. OPENSHIFT_NAMESPACES="" CUSTOM_REPO="" REGISTRY="""" TEST_OPENSHIFT_4=true /usr/bin/env bash common/test.sh
make: *** [test-openshift-4] Error 1
```

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -61,9 +61,9 @@ function ct_enable_cleanup() {
 # Function returns either 0 in case of pull was successful
 # Or the test suite exit with 1 in case of pull error
 function ct_pull_image() {
-  local image_name="$1"; shift
-  local exit=${1:-"false"}; shift
-  local loops=${1:-10}; shift
+  local image_name="$1"; [[ $# -gt 0 ]] && shift
+  local exit_variable=${1:-"false"}; [[ $# -gt 0 ]] && shift
+  local loops=${1:-10}
   local loop=0
 
   # Let's try to pull image.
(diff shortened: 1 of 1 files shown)


phracek:
[test]

phracek:
Sorry wrong.

test-lib.sh
```diff
@@ -63,7 +63,7 @@ function ct_enable_cleanup() {
 function ct_pull_image() {
   local image_name="$1"; shift
```
zmiklank, line 64:
```suggestion
  local image_name="$1"; [[ $# -gt 0 ]] && shift
```
For checking if there is some variable left to shift.

phracek:
[test]

phracek:
As soon as this pull request is reviewed and approved, we can merge it.

zmiklank (approved):

zmiklank:
Great, thanks.
I would also maybe squash the commits before merging, but it is not really necessary (IMO).

phracek:
Squashed.

PR #224, opened 2021-12-13
Add support for building CentOS 9 stream

phracek:
This pull request adds support for building SCLorg images in the CentOS9 stream.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -243,6 +243,8 @@ for dir in ${dirs}; do
     docker_build_with_version Dockerfile.centos6
   elif [ "$OS" == "centos8" ] || [ "$OS" == "centos8-candidate" ]; then
     docker_build_with_version Dockerfile.centos8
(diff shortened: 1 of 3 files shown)


hhorak:
I'm not sure whether to introduce `stream9` as there is already more commonly used term `c9s`, maybe we can use `c9s` for the use cases where we previously used `centos8`? It means for the Dockerfile suffix and `$OS` value. Just to make it clear, as `stream9` might be misinterpreted.

Otherwise it LGTM.

phracek:
> I'm not sure whether to introduce `stream9` as there is already more commonly used term `c9s`, maybe we can use `c9s` for the use cases where we previously used `centos8`? It means for the Dockerfile suffix and `$OS` value. Just to make it clear, as `stream9` might be misinterpreted.
> 
> Otherwise it LGTM.

Ok, I will change it to c9s.

phracek:
Let's use c9s as a Dockefile.c9s for all containers.

PR #226, opened 2022-01-03
Print each testcase result at the end of openshift testsuite

zmiklank:
Related to https://github.com/sclorg/s2i-nodejs-container/pull/310.
If the variable that should contain the test results does not exist, nothing will be printed.


phracek (approved):
LGTM.

PR #227, opened 2022-01-04
Add supporting building containers on MacOS

phracek:
This pull request adds support for building SCLORG container on MacOS by `docker` command.

The Linux systems are not touched by this change.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -8,7 +8,13 @@
 
 set -e
 
-script_name=$(readlink -f "$0")
+UNAME=$(uname)
+
+if [[ $UNAME == "Darwin" ]]; then
+  script_name=$(greadlink -f "$0")
+else
+  script_name=$(readlink -f "$0")
+fi
 script_dir=$(dirname "$script_name")
 
 OS=${1-$OS}
@@ -173,10 +179,27 @@ function docker_build_with_version {
 
   pull_image "$dockerfile"
 
-  # shellcheck disable=SC2016
-  parse_output 'docker build '"$BUILD_OPTIONS"' -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}"' \
-               "tail -n 1 | awk '/Successfully built|(^--> )?(Using cache )?[a-fA-F0-9]+$/{print \$NF}'" \
-               IMAGE_ID
+  if [[ $UNAME == "Darwin" ]]; then
+    # docker output looks like
+    #9 sha256:6d65c6e054efd48cb168a83bb2c94578a3f998270512c42d5a40bf0d55af1461
+    #9 CACHED
+
+    #10 exporting to image
+    #10 sha256:e8c613e07b0b7ff33893b694f7759a10d42e180f2b4dc349fb57dc6b71dcab00
+    #10 exporting layers done
+    #10 writing image sha256:60575b179c79de97797d3790b0238507354c443aca4f52d40a59fc89c78f2222 done
+    #10 DONE 0.0s
+    # shellcheck disable=SC2086
+    docker_output=$(docker build $BUILD_OPTIONS --progress plain -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}" 2>&1)
+    # Print docker build output for the another analysis in case of failure
+    echo "${docker_output}"
+    IMAGE_ID=$(echo "$docker_output" | tail -n 4 | awk '/^#[0-9]+ writing image sha256:[a-fA-F0-9]+ done$/{print $(NF-1)}')
+  else
+    # shellcheck disable=SC2016
+    parse_output 'docker build '"$BUILD_OPTIONS"' -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}"' \
+                 "tail -n 1 | awk '/Successfully built|(^--> )?(Using cache )?[a-fA-F0-9]+$/{print \$NF}'" \
+                 IMAGE_ID
+  fi
   clean_image
   echo "$IMAGE_ID" > .image-id.raw
 


phracek:
[test]

phracek:
[test]

build.sh
```diff
@@ -173,10 +179,27 @@ function docker_build_with_version {
...
+    #10 exporting to image
+    #10 sha256:e8c613e07b0b7ff33893b694f7759a10d42e180f2b4dc349fb57dc6b71dcab00
+    #10 exporting layers done
+    #10 writing image sha256:60575b179c79de97797d3790b0238507354c443aca4f52d40a59fc89c78f2222 done
+    #10 DONE 0.0s
+    # shellcheck disable=SC2086
+    docker_output=$(docker build $BUILD_OPTIONS --progress plain -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}" 2>&1)
+    # Print docker build output for the another analysis in case of failure
+    echo "${docker_output}"
+    IMAGE_ID=$(echo "$docker_output" | tail -n 4 | awk '/^#[0-9]+ writing image sha256:[a-fA-F0-9]+ done$/{print $(NF-1)}')
```
hhorak, line 196: Is the problem here only the different output or the `parse_output` implementation specifically? If it's just about a different output, it can still be implemented via `parse_output`, like this:
```
parse_output 'docker build '"$BUILD_OPTIONS"' --progress plain -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}"' \
                 "tail -n 4 | awk '/^#[0-9]+ writing image sha256:[a-fA-F0-9]+ done$/{print $(NF-1)}'" \
                 IMAGE_ID
```
And the debug output `echo "${docker_output}"` might be added to the `parse_output` function for all arches.

hhorak: If the problem is with how parse_output is implemented, it should be mentioned in the comment.

phracek: I would like to get rid off `parse_output`.

ljavorsk:
I've tested if the fix works.

The "make tag ..." is fixed by this PR.
However, the "make test ..." still fails, but I couldn't find the exact line where it crashes.
Maybe you could?

The error from "make test ...":
VERSIONS="10.5" SKIP_SQUASH=1 UPDATE_BASE= OS=centos7 CLEAN_AFTER= DOCKER_BUILD_CONTEXT=.. OPENSHIFT_NAMESPACES="" CUSTOM_REPO="" REGISTRY=""quay.io/"" TEST_MODE=true /usr/bin/env bash common/test.sh
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
make: *** [test] Error 1

phracek:
I have found the problem already.
There are several binaries which are called, like `mktemp -d` (https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L726), `mktemp -p` (https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L727) and functions called here `ct_show_resources` (https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L936)

Switch this PR to draft, till also test-lib.sh is not converted.
I would prefer to have a function that will do the aliases for MacOS.

build.sh
phracek: Make sense to me, I will try to use parse_output.

phracek: I have tried to use `parse_output` but it does not work to my for some unknown reason.
I have set -ex and the output is:

```bash
++ _parse_output_inner
++ set -o pipefail
++ case $stream in
++ eval 'tail -n 7 | awk '\''/#[0-9]+ writing image sha256:[a-fA-F0-9]+ done$/{print $(NF-1)}'\'''
++ eval 'docker build  --label io.openshift.builder-version="42b35fb" --progress plain -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}"'
+++ docker build --label io.openshift.builder-version=42b35fb --progress plain -f Dockerfile .
+++ tail -n 7
+++ awk '/#[0-9]+ writing image sha256:[a-fA-F0-9]+ done$/{print $(NF-1)}'
++ tee /dev/fd/63
+++ cat -
#1 [internal] load build definition from Dockerfile
#1 sha256:c98668414a0fe70e3e58c87d16e747c7b9039098831edbeb4b05c47362ab890b
#1 transferring dockerfile: 37B done
#1 DONE 0.0s

[snip]

#9 [5/5] RUN sed -i -f /opt/app-root/nginxconf.sed /etc/opt/rh/rh-nginx120/nginx/nginx.conf &&     chmod a+rwx /etc/opt/rh/rh-nginx120/nginx/nginx.conf &&     mkdir -p /opt/app-root/etc/nginx.d/ &&     mkdir -p /opt/app-root/etc/nginx.default.d/ &&     mkdir -p /opt/app-root/src/nginx-start/ &&     mkdir -p /usr/share/container-scripts/nginx/nginx-start &&     mkdir -p /var/opt/rh/rh-nginx120/log/nginx &&     mkdir -p /opt/app-root/etc/perl &&     ln -s /var/opt/rh/rh-nginx120/log/nginx /var/log/nginx &&     ln -s /etc/opt/rh/rh-nginx120/nginx /etc/nginx &&     ln -s /opt/rh/rh-nginx120/root/usr/share/nginx /usr/share/nginx &&     chmod -R a+rwx /opt/app-root/etc &&     chmod -R a+rwx /var/opt/rh/rh-nginx120 &&     chmod -R a+rwx /usr/share/container-scripts/nginx/nginx-start &&     chown -R 1001:0 /opt/app-root &&     chown -R 1001:0 /var/opt/rh/rh-nginx120 &&     chown -R 1001:0 /usr/share/container-scripts/nginx/nginx-start &&     chmod -R a+rwx /var/run &&     chown -R 1001:0 /var/run &&     rpm-file-permissions
#9 sha256:4575eb8a8351d7f578f2bacbd709c5aa3fdb91de6fc5246ae49c6ab40ad011c0
#9 CACHED

#10 exporting to image
#10 sha256:e8c613e07b0b7ff33893b694f7759a10d42e180f2b4dc349fb57dc6b71dcab00
#10 exporting layers done
#10 writing image sha256:c49e6e29c71e848e84e6c8ed010eeeabf96803b6655552d160b3df8ddd96cb3d done
#10 DONE 0.0s

Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
++ exit 0
+ raw_output=
+ rc=0
+ eval 'IMAGE_ID=$raw_output'
++ IMAGE_ID=
+ exit 0
+ echo 'IMAGE_ID: '
IMAGE_ID:
```

phracek:
The easiest solution is to allocate a machine somewhere else and build, tag and test containers there. Or move the `container-common-scripts` to pytest.

phracek:
A better way would be to convert container-common-scripts into pytest. Closing this. We can allocate a machine via a testing farm and therefore it does not make sense to support it.

PR #229, opened 2022-01-07
Add function for running all test in $TEST_SET.

zmiklank:
The function could be then called to run every application container test suite.

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -5,17 +5,16 @@ source "$(dirname "${BASH_SOURCE[0]}")"/test-lib.sh
 
 # Set of functions for testing docker images in OpenShift using 'oc' command
 
-# A variable containing the overall test result; must be changed to 0 in the end
-# of the testing script:
-#   OS_TESTSUITE_RESULT=0
+# A variable containing the overall test result
+#   TESTSUITE_RESULT=0
 # And the following trap must be set, in the beginning of the test script:
 #   trap ct_os_cleanup EXIT SIGINT
-OS_TESTSUITE_RESULT=1
+TESTSUITE_RESULT=0
 OS_CLUSTER_STARTED_BY_TEST=0
 
 function ct_os_cleanup() {
-  echo "${test_short_summary:-}"
-  if [ $OS_TESTSUITE_RESULT -eq 0 ] ; then
+  echo "${TEST_SUMMARY:-}"
+  if [ $TESTSUITE_RESULT -eq 0 ] ; then
     # shellcheck disable=SC2153
     echo "OpenShift tests for ${IMAGE_NAME} succeeded."
   else
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -1054,6 +1054,20 @@ ct_test_app_dockerfile() {
   return $ret
 }
 
+# ct_check_testcase_result
+# -----------------------------
+# Check if testcase ended in success or error
+# Argument: result - testcase result value
+# Uses: $TESTCASE_RESULT - result of the testcase
+# Uses: $IMAGE_NAME - name of the image being tested
+ct_check_testcase_result() {
+  local result="$1"
+  if [[ "$result" != "0" ]]; then
+    echo "Test for image '${IMAGE_NAME}' FAILED (exit code: ${result})"
+    TESTCASE_RESULT=1
+  fi
+  return $result
+}
 
 # ct_run_tests_from_testset
 # -----------------------------
@@ -1066,16 +1080,16 @@ ct_test_app_dockerfile() {
 ct_run_tests_from_testset() {
   local app_name="$1"
   for test_case in $TEST_SET; do
+    TESTCASE_RESULT=0
     info "Running test $test_case ... "
     $test_case
-    local test_result=$?
+    ct_check_testcase_result $?
     local test_msg
-    if [ $test_result -eq 0 ]; then
+    if [ $TESTCASE_RESULT -eq 0 ]; then
       test_msg="[PASSED]"
     else
       test_msg="[FAILED]"
       TESTSUITE_RESULT=1
-      echo "Test '${test_case}' for image '${IMAGE_NAME}' FAILED (exit code: ${test_result})"
     fi
     printf -v TEST_SUMMARY "%s %s for '%s' %s\n" "${TEST_SUMMARY}" "${test_msg}" "${app_name}" "$test_case"
   done;
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -489,21 +489,6 @@ ct_path_foreach ()
 }
 
 
-# ct_run_test_list
-# --------------------
-# Execute the tests specified by TEST_LIST
-# Uses: $TEST_LIST - list of test names
-function ct_run_test_list() {
-  for test_case in $TEST_LIST; do
-    : "Running test $test_case"
-    # shellcheck source=/dev/null
-    [ -f "test/$test_case" ] && source "test/$test_case"
-    # shellcheck source=/dev/null
-    [ -f "../test/$test_case" ] && source "../test/$test_case"
-    $test_case
-  done;
-}
-
 # ct_gen_self_signed_cert_pem
 # ---------------------------
 # Generates a self-signed PEM certificate pair into specified directory.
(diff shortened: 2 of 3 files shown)


zmiklank:
Also, @phracek, is the `ct_run_test_list` function from `test-lib.sh` still valid? Seems, it does similar work.

phracek:
> ct_run_test_list

It could be enhanced. I totally agree to have only one function.
It is not used at all. See https://github.com/search?q=org%3Asclorg+ct_run_test_list&type=code

zmiklank:
I removed the `ct_run_test_list` function.

test-lib.sh
```diff
@@ -1069,4 +1054,34 @@ ct_test_app_dockerfile() {
...
+# Argument: app_name - application name to log
+# Uses: $TEST_SET - set of test cases to run
+# Uses: $TEST_SUMMARY - variable for storing test results
+# Uses: $IMAGE_NAME - name of the image being tested
+ct_run_tests_from_testset() {
+  local app_name="$1"
+  local testsuite_result=0
+  for test_case in $TEST_SET; do
+    info "Running test $test_case ... "
+    $test_case
```
phracek, line 1071: Move the calling function after variable local definition.

Also add after calling `$test_case`
```bash
check_result
```
like it is used here https://github.com/sclorg/s2i-perl-container/blob/master/test/run#L369

test-lib.sh
```diff
@@ -1069,4 +1054,34 @@ ct_test_app_dockerfile() {
...
+# -----------------------------
+# Runs all tests in $TEST_SET, prints result to
+# the $TEST_SUMMARY variable
+# Argument: app_name - application name to log
+# Uses: $TEST_SET - set of test cases to run
+# Uses: $TEST_SUMMARY - variable for storing test results
+# Uses: $IMAGE_NAME - name of the image being tested
+ct_run_tests_from_testset() {
+  local app_name="$1"
+  local testsuite_result=0
```
phracek, line 1068: I would prefer to set it as a main variable at the beginning of each tests
```bash
TESTSUITE_RESULT=0
```bash

https://github.com/sclorg/s2i-perl-container/blob/master/test/run#L15

File: test-lib.sh [hunk 3]
```diff
@@ -1069,4 +1054,34 @@ ct_test_app_dockerfile() {
...
+    if [ $test_result -eq 0 ]; then
+      test_msg="[PASSED]"
+    else
+      test_msg="[FAILED]"
+      testsuite_result=1
+      echo "Test '${test_case}' for image '${IMAGE_NAME}' FAILED (exit code: ${test_result})"
+    fi
+    printf -v TEST_SUMMARY "%s %s for '%s' %s\n" "${TEST_SUMMARY}" "${test_msg}" "${app_name}" "$test_case"
+  done;
+  return $testsuite_result
```
phracek, line 1084: This is not needed. I would prefer to use global variable.

test-lib.sh
```diff
@@ -1069,4 +1054,34 @@ ct_test_app_dockerfile() {
...
+    info "Running test $test_case ... "
+    $test_case
+    local test_result=$?
+    local test_msg
+    local output_format="%s %s $1\n"
+    if [ $test_result -eq 0 ]; then
+      test_msg="[PASSED]"
+    else
+      test_msg="[FAILED]"
+      testsuite_result=1
```
phracek, line 1079: This could be variable like here: https://github.com/sclorg/s2i-perl-container/blob/master/test/run#L15

And then we can unify also standard docker tests.

test-lib.sh
```diff
@@ -1069,4 +1054,34 @@ ct_test_app_dockerfile() {
...
+# Uses: $IMAGE_NAME - name of the image being tested
+ct_run_tests_from_testset() {
+  local app_name="$1"
+  local testsuite_result=0
+  for test_case in $TEST_SET; do
+    info "Running test $test_case ... "
+    $test_case
+    local test_result=$?
+    local test_msg
+    local output_format="%s %s $1\n"
```
phracek, line 1074: output_format is not used, so let's remove it.

test-lib-openshift.sh
```diff
@@ -14,7 +14,7 @@ OS_TESTSUITE_RESULT=1
 OS_CLUSTER_STARTED_BY_TEST=0
 
 function ct_os_cleanup() {
-  echo "${test_short_summary:-}"
+  echo "${TEST_SUMMARY:-}"
   if [ $OS_TESTSUITE_RESULT -eq 0 ] ; then
```
phracek, line 18:
```suggestion
  if [ TESTSUITE_RESULT -eq 0 ] ; then
```

Like here: https://github.com/sclorg/s2i-perl-container/blob/master/test/run#L15 or
https://github.com/sclorg/s2i-perl-container/blob/master/test/run#L405

phracek (changes requested):
Thanks for the PR!. Please fix issues or my recommendation.

test-lib.sh
zmiklank: Well, the only line which is needed from the `check_result` is the logging one, and IMHO clearer solution of its placement, is where it currently is in this PR. 
```
      test_msg="[FAILED]"
      testsuite_result=1
      echo "Test '${test_case}' for image '${IMAGE_NAME}' FAILED (exit code: ${test_result})"
```
Moreover, the `check_result` isn't owned by container-common-scripts, but is kind of specific to every container  (even though its implementation is quite similar through the sclorg codebase, it is not always the same).
The main function of `check_result` is IMHO to set the `$TESTCASE_RESULT` __global__ variable to 1, if the ret_val is non zero. However, in my opinion, this variable shouldn't be global, but should be in the local scope of `ct_run_tests_from_testset` function, as the only result, which really interests us, is the result [0/1] of the whole test suite.
What do you think?

test-lib.sh
zmiklank: Sorry, I've missed that. Resolved.

test-lib.sh
zmiklank: If the variable is local, and its value is returned by this function, then the docker tests can use TESTSUITE_RESULT and openshift tests could use the OS_TESTSUITE_RESULT variable, as it is currently designed.

However, the truth is, that this would have to be implemented differently, for a case when the user calls SIGINT during the run of the test suite, because then [OS_]TESTSUITE_RESULT wouldn't be touched by the  `ct_run_tests_from_testset` function, and thus would return incoherent result. This could be fixed by setting the [OS_]TESTSUITE_RESULT to 1 in the beginning of the `run*` scripts.

Why do you prefer the global variable?

test-lib.sh
phracek: The I would have an idea. Move check_result into `container-common-scripts` as well.
Let's make the whole functionality in `container-common-scripts` repo. The rest of containers will share it and we will remove duplicates. Good catch and approach in the discussion.
the chec_result as soon as after `$test_case` is used in case of `test_case` itself does not check in the function check_result. We need the return value.

test-lib.sh
phracek: What about if we run several ct_run_tests_from_testset. Not only the one. Then we have to if the function `ct_run_tests_from_testset` returns [0/1]. The global does this work.

zmiklank: Well, this could be IMHO handled in the main `run*` script. It would be readable easier.

zmiklank:
I updated the `ct_run_tests_from_testset` function, so it works with global `TESTSUITE_RESULT`.

test-lib.sh
zmiklank: Ok, I will move the check result to the `test-lib.sh`.

zmiklank:
I added a `_ct_check_testcase_result` function to `test-lib.sh`.

test-lib.sh
```diff
@@ -1069,4 +1054,45 @@ ct_test_app_dockerfile() {
   return $ret
 }
 
+# _ct_check_testcase_result
+# -----------------------------
+# Check if testcase ended in success or error
+# Argument: result - testcase result value
+# Uses: $TESTCASE_RESULT - result of the testcase
+# Uses: $IMAGE_NAME - name of the image being tested
+_ct_check_testcase_result() {
```
phracek, line 1063:
```suggestion
ct_check_testcase_result() {
```
As far as I know, we do not use `_` prefix in functions.

zmiklank: Ok.

phracek (approved):
Thanks for the update. LGTM.
Let's try it in our containers.

phracek:
[test]

zmiklank:
[test]

zmiklank:
[test]

PR #235, opened 2022-03-03
change the info log function to echo, as it is not available in the test-lib.sh

zmiklank:
This fixes the following error(if you have info command installed in your system):
`info: No menu item 'Running test test_perl_imagestream ... ' in node '(dir)Top'`


phracek (approved):
LGTM. This change does not have an effect on running tests.

PR #238, opened 2022-03-22
Improve log debug

hhorak:
By including timestamp, test duration, and a bit more separators in the test log, it should be easier to debug tests and read longer logs.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -49,6 +49,24 @@ function ct_cleanup() {
   done
   rmdir "$CID_FILE_DIR"
   : "Done."
+
+  ct_show_results
+  exit $TESTCASE_RESULT
+}
+
+# ct_show_results
+# ---------------
+# Prints results of all test cases that are stored into TEST_SUMMARY variable.
+# Uses: $TEST_SUMMARY - text info about test-cases
+# Uses: $TESTCASE_RESULT - overall result of all tests
+function ct_show_results() {
+  echo "$TEST_SUMMARY"
+
+  if [ $TESTSUITE_RESULT -eq 0 ] ; then
+    echo "Tests for ${IMAGE_NAME} succeeded."
+  else
+    echo "Tests for ${IMAGE_NAME} failed."
+  fi
 }
 
 # ct_enable_cleanup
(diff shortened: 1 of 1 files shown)


hhorak:
[test]

hhorak:
[test]

hhorak:
Another example of usage: https://github.com/sclorg/s2i-perl-container/pull/230

hhorak:
[test]

test-lib.sh
```diff
@@ -1087,21 +1114,64 @@ ct_check_testcase_result() {
...
+# --------------
+# Returns timestamp in seconds
+function ct_timestamp_s() {
+  date '+%s'
+}
+
+# ct_timestamp_pretty
+# -----------------
+# Returns timestamp readable to a human
+function ct_timestamp_pretty() {
```
phracek, line 1161: It would be nice to have an example in a comment. What is a readable timestamp?

test-lib.sh
```diff
@@ -1087,21 +1114,64 @@ ct_check_testcase_result() {
...
+# ct_timestamp_diff
+# -----------------
+# Computes a time diff between two timestamps
+# Argument: start_date - Beginning (in seconds)
+# Argument: final_date - End (in seconds)
+# Returns: Time difference in format HH:MM:SS
+function ct_timestamp_diff() {
+  local start_date=$1
+  local final_date=$2
+  date -u -d "0 $final_date seconds - $start_date seconds" +"%H:%M:%S"
```
phracek, line 1174: Can you please add to the comment an example of what is input, like format, and was is output?

phracek (changes requested):
Does it make sense to have a test case for time functions?

test-lib.sh
hhorak: Good idea, will add an example.

test-lib.sh
hhorak: Output is described there: "Returns: Time difference in format HH:MM:SS"

hhorak: I've added some more info what exactly "seconds" means.

hhorak:
[test]

phracek (approved):
LGTM. Thanks. Please squash the commits before MERGE.

PR #240, opened 2022-03-24
force image removal during cleanup

pkubatrh:
podman will complain if we try to remove an image by ID when it has multiple tags
We are always using mulitple tags, so we need to force the removal


phracek (approved):
Thanks for this catch. LGTM. Would it be possible to add a note, as to why the row was added? It is only nitpicking ;) Otherwise feel free to merge it.

phracek:
[test]

pkubatrh:
The reason for the change was explained in the commit message. I do not think we need to add info directly into the code as a comment. Rebasing for tests

[test]

pkubatrh:
[test]

PR #242, opened 2022-03-28
Add a function for getting uncompressed image size and show the size

hhorak:
This might help when playing with container size changes. We can later add some check to see what is the difference between the current size and the last image available, but this is not implemented yet.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -1223,12 +1223,14 @@ ct_get_image_size_uncompresseed() {
 # ct_get_image_size_compresseed
 # -------------------------------
 # Shows compressed image size in MB
-# This might not be entirely same as podman, but should be close enough.
+# This is a slight hack, that counts compressed size based on the compressed
+# content. It might not be entirely same as what docker pull shows, but should
+# be close enough.
 # Argument: image_name - image locally available
 ct_get_image_size_compresseed() {
   local image_name=$1
   local size_bytes
-  size_bytes=$(podman save "${image_name}" | gzip - | wc --bytes)
+  size_bytes=$(docker save "${image_name}" | gzip - | wc --bytes)
   echo "$(( size_bytes / 1024 / 1024 ))MB"
 }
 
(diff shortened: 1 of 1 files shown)


hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

pkubatrh:
Could we also add compressed size as well?
Might be good idea to list the image size at the end of the tests instead of at the start, so that we do not have to look through the log to find the info.

hhorak:
@pkubatrh I've added the size to the end, that seems like a better place. I'm not sure though what's the best way to get the compressed size -- the only way I see when working with a local image is exporting to a tar and then gzip it, which does not seem to be what we actually need. Any suggestions?

hhorak:
We discussed today on a meeting, that for approximate compressed size, we could likely export an image and gzip it.

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

test-lib.sh
```diff
@@ -1208,8 +1209,20 @@ function ct_timestamp_diff() {
...
+
+# ct_get_image_size_compresseed
+# -------------------------------
+# Shows compressed image size in MB
+# This might not be entirely same as podman, but should be close enough.
+# Argument: image_name - image locally available
+ct_get_image_size_compresseed() {
+  local image_name=$1
+  local size_bytes
+  size_bytes=$(podman save "${image_name}" | gzip - | wc --bytes)
```
zmiklank, line 1225: This would fail when using docker. It would IMO be sufficient to use `docker` instead fo `podman`.

zmiklank (changes requested):
Thanks for the PR!

See my comment. Beside that this LGTM.

test-lib.sh
hhorak: Aj, this is obviously a bug, thanks. Will fix it finally.

hhorak:
[test]

hhorak:
All feedback addressed, CI passes (except el7 that makes problems) but there is no risk of regression, so I believe we can merge now.

PR #244, opened 2022-03-28
Update tag.sh with proper postfix

phracek:


phracek:
[test]

phracek:
[test]

phracek:
This is not needed. It can be done via ENV variable.

PR #245, opened 2022-03-29
Add ct_get_certificate_timestamp and ct_get_certificate_age_s functions

hhorak:
These functions are expected to be used for checking whether a certificate was just
generated during the container assembing (e.g age of the certificate is less than a minute).

The certificates generation on assembling the container with the application is important
for not having the container images with certificates being part of the image.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -1410,6 +1410,55 @@ function ct_timestamp_diff() {
   date -u -d "0 $final_date seconds - $start_date seconds" +"%H:%M:%S"
 }
 
+# ct_get_certificate_timestamp
+# ----------------------------
+# Looks into a running container into a specified file (certificate) and extracts
+# a notBefore date.
+# Argument: container - ID of a running container
+# Argument: path - path to the certificate inside the running container
+# Returns: timestamp (seconds since Unix era) for the certificate generation
+function ct_get_certificate_timestamp() {
+  local container=$1
+  local path=$2
+  date '+%s' --date="$(docker exec "$container" bash -c "cat $path" | openssl x509  -startdate -noout | grep notBefore | sed -e 's/notBefore=//')"
+}
+
+# ct_get_certificate_age_s
+# ------------------------
+# Looks into a running container into a specified file and retuns age of the certificate
+# Argument: container - ID of a running container
+# Argument: path - path inside the running container
+# Returns: age of the certificate in seconds
+function ct_get_certificate_age_s() {
+  local container=$1
+  local path=$2
+  local now
+  local cert_timestamp
+  now=$(date '+%s')
+  cert_timestamp=$(ct_get_certificate_timestamp "$container" "$path")
+  echo $(( now - cert_timestamp ))
+}
+
+# ct_get_image_age_s
+# ------------------
+# Retuns age of a given image in seconds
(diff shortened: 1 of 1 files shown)


hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

test-lib.sh
```diff
@@ -1186,4 +1186,58 @@ function ct_timestamp_diff() {
   date -u -d "0 $final_date seconds - $start_date seconds" +"%H:%M:%S"
 }
 
+# ct_get_certificate_timestamp
+# ----------------------------
+# Looks into a running container into a specified file (certificate) and extracts
+# a notBefore date.
+# Argument: container - ID of a running container
+# Argument: path - path inside the running container
```
zmiklank, line 1194: Is this path to **a certificate** inside the dockerfile?

test-lib.sh
```diff
@@ -1186,4 +1186,58 @@ function ct_timestamp_diff() {
...
+# Returns: timestamp (seconds since Unix era) for the certificate generation
+function ct_get_certificate_timestamp() {
+  local container=$1
+  local path=$2
+set -x
+which openssl >&2
```
zmiklank, line 1200: Is this formatting intentional?

test-lib.sh
```diff
@@ -1195,7 +1195,12 @@ function ct_timestamp_diff() {
 function ct_get_certificate_timestamp() {
   local container=$1
   local path=$2
-  date '+%s' --date="$(docker exec -ti "$container" bash -c "cat $path" | openssl x509  -startdate -noout | grep notBefore | sed -e 's/notBefore=//')"
+set -x
```
zmiklank, line 1199: It would not be harmful IMO, if the `set -x` stays here even after debugging of this PR is finished. But then would be better to append closing `set +x` at the end of this function.

zmiklank (changes requested):
I am not sure, if this PR is still in queue to be reviewed, nonetheless, here is my review :).

I would also suggest squashing the commits before merging/rebasing.

Else LGTM.

pkubatrh:
@hhorak Can you take a look at the comments?

test-lib.sh
hhorak: This looks like a not-finished PR, will take a look and fix the formatting.

test-lib.sh
hhorak: This was likely an artifact left from debugging. I'd rather not handle with `-x` within some tests. I'll remove this.

test-lib.sh
hhorak: yes, will add this to the comment

hhorak:
This PR is tested in https://github.com/sclorg/httpd-container/pull/140, so not intended to be merged before it's working in CI in that PR.

hhorak:
[test]

hhorak:
Rebased and https://github.com/sclorg/httpd-container/pull/140 testing it right now, let's see whether tests there are green.

zmiklank (approved):
LGTM. Thanks for the great comments in the code!

PR #246, opened 2022-03-31
Add support for building CentOS Stream 8

phracek:

### DIFF
--- a/tag.sh
+++ b/tag.sh
@@ -19,11 +19,9 @@ for dir in ${VERSIONS}; do
   date_and_hash="${commit_date}-$(git rev-parse --short HEAD)"
 
   full_reg_name="$REGISTRY$name"
-  echo "-> Tagging image '$IMAGE_ID' as '$full_reg_name:$version' and '$full_reg_name:latest' and '$full_reg_name:$date_and_hash'"
-  if [[ $OS == "c9s" ]] || [[ $OS == "fedora" ]]; then
-    echo "-> Tagging image '$IMAGE_ID' as '$full_reg_name:$OS'"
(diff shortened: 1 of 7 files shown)


phracek:
@zmiklank @hhorak @pkubatrh  I would be glad for review and feedback

phracek:
[test]

phracek:
[test]

phracek:
[test]

tag.sh
```diff
@@ -19,11 +19,9 @@ for dir in ${VERSIONS}; do
   date_and_hash="${commit_date}-$(git rev-parse --short HEAD)"
 
   full_reg_name="$REGISTRY$name"
-  echo "-> Tagging image '$IMAGE_ID' as '$full_reg_name:$version' and '$full_reg_name:latest' and '$full_reg_name:$date_and_hash'"
-  if [[ $OS == "c9s" ]] || [[ $OS == "fedora" ]]; then
-    echo "-> Tagging image '$IMAGE_ID' as '$full_reg_name:$OS'"
-    docker tag "$IMAGE_ID" "$full_reg_name:$OS"
-  fi
+  echo "-> Tagging image '$IMAGE_ID' as '$full_reg_name:$version' and '$full_reg_name:latest' and '$full_reg_name:$OS' and '$full_reg_name:$date_and_hash'"
```
zmiklank, line 22: Could you please explain this change?
I would like to double-check because it affects also other distributions.

phracek: It does not make sense to tag image as `:$OS` only for c9s, c8s, and fedora. Let's do it for all images. But only c8s, c9s and fedora use it.

PR #247, opened 2022-04-06
Update black with click dependencies

phracek:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,9 +4,10 @@
 
 repos:
   - repo: https://github.com/ambv/black
-    rev: 19.3b0
+    rev: 22.1.0
     hooks:
       - id: black
+        additional_dependencies: ['click==8.0.4']
     language_version: python3.6
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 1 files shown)


phracek:
[test]

phracek:
[test]

test-lib.sh
```diff
@@ -36,7 +36,8 @@ function ct_cleanup() {
     docker stop "$container"
 
     # Container has not been removed by `docker stop` and still exists
-    if [ $( docker ps -a -f id=$container | wc -l ) -eq 2 ]; then
+    # shellcheck disable=SC2046
```
pkubatrh, line 39: Why are we disabling the check instead of using double quotes as expected?

phracek: I don't like double quotes twice. :( But I can change it.

phracek: Like here: https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L79

phracek:
[test]

phracek:
[test]

phracek:
The Sanity tests are passing. This does not affect tests. Merging it.

PR #252, opened 2022-05-11
Deploy image into internal OpenShift registry.

phracek:
We do not need an external registry at all.

By setting the variable CT_OCP4_TEST we use OpenShift 4 internal registry.


test-lib-remote-openshift.sh
```diff
@@ -98,16 +98,11 @@ function ct_os_import_image_ocp4() {
   local image_name="${1}"; shift
   local imagestream=${1:-$image_name:latest}
   local namespace
+  local image_name_no_namespace=${image_name##*/}
+  local image_tagged="${image_name_no_namespace%:*}:${VERSION}"
```
hhorak, line 102: I don't understand why those variables are added, they seem to be not used in the function.

test-lib-openshift.sh
```diff
@@ -424,16 +444,39 @@ function ct_os_docker_login() {
...
+      return 0
+    fi
+    return 1
+  fi
+  if [ "${os_version}" == "v4" ]; then
+    # Variable OCP4_ADDRESS is set in function ct_os_docker_login_v4
+    if ct_os_docker_login_v4; then
+      output_name="$OCP4_ADDRESS/$namespace/$image_tagged"
+      docker tag "${image_tagged}" "${output_name}"
+      docker push --tls-verify=false "${output_name}"
```
hhorak, line 475: These 3 commands are almost the same as in case of v3, so it's possible to have this piece of code only once and only define `$address`, `$namespace` and `$image_tagged` differently for v3 and v4 if needed. It's not a large piece of code though, so I don't have a big problem with keeping it as it is.

hhorak:
LGTM, with just 2 small nitpicks to consider, nothing from that is a blocker for merging though.

test-lib-openshift.sh
phracek: Thanks for the review. I have addressed your comment. Please review it again.

phracek:
[test]

hhorak:
Cool, sounds good, thanks. LGTM now.

PR #253, opened 2022-05-17
add support for unstable tests to common

zmiklank:
This PR proposes the first step to use the unstable flag for unstable tests in the containers.
See more in https://github.com/sclorg/s2i-python-container/pull/510

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -1115,6 +1115,8 @@ ct_check_testcase_result() {
 # Uses: $TEST_SET - set of test cases to run
 # Uses: $TEST_SUMMARY - variable for storing test results
 # Uses: $IMAGE_NAME - name of the image being tested
+# Uses: $UNSTABLE_TESTS - set of tests, whose result can be ignored
+# Uses: $IGNORE_UNSTABLE_TESTS - flag to ignore unstable tests
 ct_run_tests_from_testset() {
   local app_name="${1:-appnamenotset}"
   local time_beg_pretty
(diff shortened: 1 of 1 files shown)


test-lib.sh
```diff
@@ -1088,10 +1088,15 @@ ct_check_testcase_result() {
...
 # Uses: $TEST_SUMMARY - variable for storing test results
 # Uses: $IMAGE_NAME - name of the image being tested
+# Uses: $UNSTABLE_TESTS - set of tests, whose result can be ignored
 ct_run_tests_from_testset() {
   local app_name="$1"
+  local is_unstable=0
   for test_case in $TEST_SET; do
     TESTCASE_RESULT=0
+    if [[ " ${UNSTABLE_TESTS[@]} " =~ " ${test_case} " ]]; then
+      is_unstable=1
```
frenzymadness, line 1098: We are in a for loop here and there is no way to set `is_unstable` back to zero so I'm afraid that one unstable test would cause all latter tests to be also considered unstable. Or not?

zmiklank: Right, sorry, fixed.

frenzymadness:
Do you plan to open a PR against Python container so we can test them together?

zmiklank:
Yes, it is in progress, I am now testing it locally. When the change is ready I will create the PR against s2i-python.

zmiklank:
[test-all]

phracek (approved):

PR #254, opened 2022-05-19
Add support for building RHEL9 images

phracek:

### DIFF
--- a/betka.py
+++ b/betka.py
@@ -119,7 +119,7 @@ class BetkaGenerator(object):
         if self.os_env == "centos7":
             msg = """
 'centos7' target is not supported.
-Target has to be e.g. rhel7, fedora or rhel8.
+Target has to be e.g. rhel7, rhel8, rhel9 or fedora.
 e.g. make betka TARGET=fedora VERSIONS=XY
             """
             logging.info(msg)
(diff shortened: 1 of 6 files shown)


phracek:
[test]

phracek:
[test]

betka.py
```diff
@@ -197,11 +197,16 @@ def convert_branch_to_cwt_tool(self) -> Any:
         fields = self.downstream_branch.split("-")
         if self.os_env == "rhel7":
             self.cwt_config = f"rhel7.yaml:{fields[0]}{fields[1].replace('.','')}0"
-        else:
+        elif self.os_env == "rhel8":
             release_fields = fields[1].split(".")
             self.cwt_config = (
                 f"rhel8.yaml:{fields[0]}{release_fields[0]}.{release_fields[1]}"
             )
+        else:
```
pkubatrh, line 205: Just a nitpick, but I would actually use `elif` here as well to explicitly check whether rhel9 was used rather than expecting no other choice, as well as using the `else` branch to fail in case something other than rhel{7,8,9} was passed as a target

pkubatrh:
otherwise lgtm

phracek:
[test]

PR #255, opened 2022-05-19
Print status on cleanup for containers

zmiklank:
This adds support for https://github.com/sclorg/s2i-python-container/pull/515.

This PR proposes printing test results on `ct_cleanup`, so the container tests can also use `ct_run_tests_from_testset`.
This PR should not affect containers that yet do not support the `ct_run_tests_from_testset` function, as in that case nothing is printed.

This PR also proposes to exit the shell at the end of the cleanup, so that the user can terminate the test suite manually when `ct_enable_cleanup` is called before.

If the reviewer thinks, this should be separated into two commits I understand and will do it.


zmiklank:
[test-all]

phracek (approved):
LGTM. Thanks!. Good work

zmiklank:
I will close this, as this issue is already addressed in https://github.com/sclorg/container-common-scripts/pull/238.

PR #257, opened 2022-05-27
Shellcheck fixes in the last PR merged

hhorak:
@zmiklank Pls review, whether the changes do not have an unintentional side-effect (I didn't test the functionality itself, only wanted to make the scripts shellcheck clean again). There is a handy `make shellcheck` call that helps keeping the scripts less error-prone.


zmiklank (approved):
Thanks for the fixes. I did not spot any side effects.

PR #259, opened 2022-06-02
Fix implementation of unstable tests

frenzymadness:
A whole application can be unstable (like pipenv-test-app) not a specific test case (like test_application).

Cc: @zmiklank

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -29,6 +29,7 @@ UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
 # unexpectedly. Removes the cid_files and CID_FILE_DIR as well.
 # Uses: $CID_FILE_DIR - path to directory containing cid_files
 # Uses: $EXPECTED_EXIT_CODE - expected container exit code
+# Uses: $TESTSUITE_RESULT - overall result of all tests
 function ct_cleanup() {
   ct_show_resources
   for cid_file in "$CID_FILE_DIR"/* ; do
@@ -55,14 +56,14 @@ function ct_cleanup() {
   : "Done."
 
   ct_show_results
-  exit "${TESTCASE_RESULT:-0}"
+  exit "${TESTSUITE_RESULT:-0}"
 }
 
 # ct_show_results
 # ---------------
 # Prints results of all test cases that are stored into TEST_SUMMARY variable.
 # Uses: $TEST_SUMMARY - text info about test-cases
-# Uses: $TESTCASE_RESULT - overall result of all tests
+# Uses: $TESTSUITE_RESULT - overall result of all tests
 function ct_show_results() {
   echo
   echo "==============================================="
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -18,6 +18,10 @@
 EXPECTED_EXIT_CODE=0
 export TESTSUITE_RESULT=0
 
+# define UNSTABLE_TESTS if not already defined, as this variable
+# is not mandatory for containers
+UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
+
 # ct_cleanup
 # --------------------
 # Cleans up containers used during tests. Stops and removes all containers
@@ -1134,7 +1138,7 @@ ct_run_tests_from_testset() {
   for test_case in $TEST_SET; do
     TESTCASE_RESULT=0
     # shellcheck disable=SC2076
-    if [[ " ${UNSTABLE_TESTS[*]} " =~ " ${test_case} " ]]; then
+    if [[ " ${UNSTABLE_TESTS[*]} " =~ " ${app_name} " ]]; then
       is_unstable=1
     else
       is_unstable=0
@@ -1150,8 +1154,10 @@ ct_run_tests_from_testset() {
     if [ $TESTCASE_RESULT -eq 0 ]; then
       test_msg="[PASSED]"
     else
-      test_msg="[FAILED]"
-      if [ -z "$IGNORE_UNSTABLE_TESTS" ] || [ $is_unstable -eq 0 ]; then
+      if [ -n "${IGNORE_UNSTABLE_TESTS:-""}" ] && [ $is_unstable -eq 1 ]; then
+        test_msg="[FAILED][UNSTABLE-IGNORED]"
+      else
+        test_msg="[FAILED]"
         TESTSUITE_RESULT=1
       fi
     fi


zmiklank:
Oh, from all the conversations and the naming of variables I understood, that it is the tests, which can be unstable, not apps, sorry. As I look again at your PR now, I see, that I must have overlooked that.

However, as my point of view on this topic changed now a lot, I begin to think, if it was not useful to implement the unstable apps and the unstable tests too.

This should be technically easy, as the apps are never named the same as the tests.
It could look something like this:
```
if [[ " ${UNSTABLE_TESTS[*]} " =~ " ${test_case} " ]] ||  [[ " ${UNSTABLE_TESTS[*]} " =~ " ${app_name} " ]] then
```

However, I am not sure if this use case is needed or if it is important. Maybe this could be added just after some container would need that.

WDYT?

frenzymadness:
My point of view might be biased because I know all the details only from the Python container but I think ignoring the results of unstable tests is needed only for applications, not for specific tests of that application.

If any external influence might impact a test, I want to mark the whole app as unstable because it does not matter which test I run on top of the application.

Correct me, if I'm wrong but this was my initial intention.

phracek:
@frenzymadness But what about if UNSTABLE_TESTS is not defined: https://github.com/sclorg/s2i-nodejs-container/pull/337#issuecomment-1147171019

test-lib.sh
```diff
@@ -1134,7 +1134,7 @@ ct_run_tests_from_testset() {
   for test_case in $TEST_SET; do
     TESTCASE_RESULT=0
     # shellcheck disable=SC2076
-    if [[ " ${UNSTABLE_TESTS[*]} " =~ " ${test_case} " ]]; then
+    if [[ " ${UNSTABLE_TESTS[*]} " =~ " ${app_name} " ]]; then
```
phracek, line 1137: I guess we need to check also if UNSTABLE_TESTS is defined or not.

phracek (changes requested):
Check also availability for variable UNSTABLE_TESTS.

test-lib.sh
zmiklank: Hi, this is my fault, I pushed the fixing commit here, please check if it is correct, or if not just discard it.

test-lib.sh
```diff
@@ -16,6 +16,8 @@
 
 # may be redefined in the specific container testfile
 EXPECTED_EXIT_CODE=0
+UNSTABLE_TESTS=""
```
frenzymadness, line 19: What happens when I have `declare -a UNSTABLE_TESTS=(pipenv-test-app)` in my `test/run`. Does this line redefines the variable?

zmiklank: In this order it would work:
```
$ UNSTABLE_TESTS=""
$ declare -a UNSTABLE_TESTS=(pipenv-test-app)
$ echo $UNSTABLE_TESTS
pipenv-test-app
```
However, the `testlib.sh` must be included in the script before declaring the array and that is inconvenient. I have changed the implementation a little, thanks.

zmiklank:
[test-all]

zmiklank:
I have also proposed to let the user know, that the test result was ignored because of an unstable test.
Retest:
[test-all]

phracek:
@zmiklank @frenzymadness Please rebase on the top of master.

test-lib.sh
```diff
@@ -1150,10 +1154,11 @@ ct_run_tests_from_testset() {
     if [ $TESTCASE_RESULT -eq 0 ]; then
       test_msg="[PASSED]"
     else
-      test_msg="[FAILED]"
-      if [ -z "$IGNORE_UNSTABLE_TESTS" ] || [ $is_unstable -eq 0 ]; then
+      if [ -z "${IGNORE_UNSTABLE_TESTS:-""}" ] || [ $is_unstable -eq 0 ]; then
+        test_msg="[FAILED]"
         TESTSUITE_RESULT=1
       fi
+      test_msg="[FAILED][UNSTABLE-IGNORED]"
```
frenzymadness, line 1161: I like the idea but it seems to me that the new message is used even when `IGNORE_UNSTABLE_TESTS` is not defined. Actually, it does not matter what combination of the two variables you have, the new message is used every time a test fails.

frenzymadness:
> From my point of view, it LGTM. especially tag `[FAILED][UNSTABLE-IGNORED]`

I think the tag does not work as expected. See my inline comment.

phracek:
This issue blogs https://github.com/sclorg/mariadb-container/pull/178#issuecomment-1148287702

phracek:
> > From my point of view, it LGTM. especially tag `[FAILED][UNSTABLE-IGNORED]`
> 
> I think the tag does not work as expected. See my inline comment.

I see, it should be [PASSED][UNSTABLE-IGNORED]

test-lib.sh
zmiklank: right, it seems, it was like this from the beginning. please rereview.

phracek:
@frenzymadness Sorry, I am lost with this issue. What should be the real outcome?

If the test is `unstable` what should be a tag in the results [FAILED] or [PASSED]. Definitely tests suit should not failed with `1` but `0`

frenzymadness:
> @frenzymadness Sorry, I am lost with this issue. What should be the real outcome?
> 
> If the test is `unstable` what should be a tag in the results [FAILED] or [PASSED]. Definitely tests suit should not failed with `1` but `0`

The new tag `[UNSTABLE-IGNORED]` should be there only if the result is really ignored but now it's there even if the result is not ignored and testsuite fails. I'm gonna fix that and send a commit here.

frenzymadness:
I see that @zmiklank already changed that. Thank you. Let me test it.

phracek:
[test]

frenzymadness:
When I have `pipenv-test-app` (intentionally broken for the testing purposes) in the list of unstable tests and I do not specify `IGNORE_UNSTABLE_TESTS`, the result is:
```
…
 [FAILED] for 'pipenv-test-app' test_application (00:00:37)
 [FAILED] for 'pipenv-test-app' test_application_with_user (00:00:37)
 [FAILED] for 'pipenv-test-app' test_application_enable_init_wrapper (00:00:37)
…
Tests for quay.io/fedora/python3-39:0 failed.
```
And the exit code is 0. <-- this does not seem to be correct but it's probably fixed by https://github.com/sclorg/container-common-scripts/commit/b0bd592ffa3ee5da92296ea8c197aaba644c8468

When I specify `IGNORE_UNSTABLE_TESTS=1` on the command line like `IGNORE_UNSTABLE_TESTS=1 make test TARGET=fedora VERSIONS=3.9`, the result is:
```
 [FAILED][UNSTABLE-IGNORED] for 'pipenv-test-app' test_application (00:00:37)
 [FAILED][UNSTABLE-IGNORED] for 'pipenv-test-app' test_application_with_user (00:00:38)
 [FAILED][UNSTABLE-IGNORED] for 'pipenv-test-app' test_application_enable_init_wrapper (00:00:38)
…
Tests for quay.io/fedora/python3-39:0 failed.
```
And the exit code is 0 which is fine.

I'm gonna rebase this branch to the top of the master branch and do another round of testing. Stay tuned.

frenzymadness:
After a call with @phracek and @zmiklank we have realized that the exit code is a bug we have to fix and we're gonna do it here.

frenzymadness:
I can confirm that it works now. I've tested all four combinations of the two new variables.

frenzymadness:
@phracek wanna do the final check?

Also, don't forget to squash all the commits to one when merging.

zmiklank:
@frenzymadness great. I have squashed the commits into two commits, as the testsuite result fix has a little common with unstable tests.

zmiklank:
And @frenzymadness, thank you for the thorough review :)

PR #261, opened 2022-06-06
Use -s for --no-patch so it works on rhel7 correctly

hhorak:
Surprisingly, --no-patch option does not exist in git that is on RHEL-7. It still has the option -s that does the same, so let's use this one, to make it not print warning.


hhorak:
[test-all]

phracek (approved):
LGTM. Thanks for the fix.

PR #263, opened 2022-06-09
Proper calling docker login into Internal OpenShift registry

phracek:


phracek:
Container-common-scripts tests do not test openshift functionality. So testing is not needed https://github.com/sclorg/container-common-scripts/issues/264

phracek:
[test]

PR #266, opened 2022-06-14
do not squash rhel9 images

pkubatrh:


pkubatrh:
Even RHEL8 does not need it defined. SKIP_SQUASH=1 is default:
https://github.com/sclorg/container-common-scripts/blob/master/common.mk#L67

phracek (approved):
LGTM. Thanks for answer.

PR #267, opened 2022-06-15
Get rid off UPDATE_BASE variable

phracek:
UPDATE_BASE variable is not needed.
It is replaced by function `pull_image` here: https://github.com/sclorg/container-common-scripts/blob/master/build.sh#L81


phracek:
[test]

zmiklank (approved):
I agree with the removal. But please explain this a little also in the commit message.

PR #268, opened 2022-06-16
support for run_test_cb

zmiklank:
This change should be backward compatible. It is needed because of moving s2i-nodejs container tests to generic test function. However, I suppose there will be also other containers like this.
See some info in the commit message.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -1140,6 +1140,16 @@ ct_run_tests_from_testset() {
   git show -s
   echo
 
+  # The run_test_cb function can be defined in each of the containers,
+  # when they need specific way of running the tests.
+  # If, for the concrete container, is enough to run the test function
+  # when testing, then it can use run_test_cb defined below.
+  if [[ $(type -t run_test_cb) != function ]]; then
+    run_test_cb() {
+      $1
+    }
+  fi
+
   for test_case in $TEST_SET; do
     TESTCASE_RESULT=0
     # shellcheck disable=SC2076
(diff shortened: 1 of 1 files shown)


zmiklank:
[test-all]

zmiklank:
I have rebased on master.
[test-all]

phracek:
Sorry, I don't follow what means this change. In case `run_test_cb` is not function, then we call function run_test_cb?

Any examples for my understanding would be nice. In this case, I am a newbie;)

phracek (changes requested):
Sorry, I don't follow what means this change. In case run_test_cb is not function, then we call function run_test_cb?

Any examples for my understanding would be nice. In this case, I am a newbie;)

zmiklank:
Sure.

Some containers (in this case the s2i-nodejs one) have more complex test execution than ordinary container images. These can write their own function for this, as in [here](https://github.com/sclorg/s2i-nodejs-container/blob/949fb4526e4254c87dcc9391ff30133940a0b52b/test/run#L131).
If this is not needed, then the `run_test_cb` is defined directly in the `ct_run_tests_from_testset`, see here https://github.com/sclorg/container-common-scripts/blob/09e45fc6ab0c29f8cb2b61b0f8278a2f472aaa43/test-lib.sh#L1147
The `run_test_cb` is defined in `ct_run_tests_from_testset`, only if not defined before.

phracek:
@zmiklank This is not needed now, I guess. Right?

zmiklank:
Yes, I need to rework this.

test-lib.sh
```diff
@@ -1140,6 +1140,16 @@ ct_run_tests_from_testset() {
   git show -s
   echo
 
+  # The run_test_cb function can be defined in each of the containers,
+  # when they need specific way of running the tests.
+  # If, for the concrete container, is enough to run the test function
+  # when testing, then it can use run_test_cb defined below.
+  if [[ $(type -t run_test_cb) != function ]]; then
+    run_test_cb() {
+      $1
```
hhorak, line 1149: wouldn't `eval $1` be better, so we do not run a sub-shell unnecessarily?

PR #270, opened 2022-06-28
Separate container tests into alone CI test - part 2

phracek:
This pull request adds support for real testing containers by container-common-scripts.

The command for running tests is:
```bash
TESTED_IMAGE=postgresql-container make test OS=<OS>
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 SHELL := /usr/bin/env bash
 
 all:
-	@echo >&2 "Only 'make check' allowed"
+	@echo >&2 "Only 'make shellcheck', 'make test', or 'make test-openshift-4' are allowed"
 
-.PHONY: check test all check-failures
+.PHONY: test all check-failures check-squash check-latest-imagestream test test-openshift-4
 
 TEST_LIB_TESTS = \
 	path_foreach \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -50,7 +47,4 @@ check-latest-imagestream:
 check-betka:
 	cd tests && ./check_betka.sh
 
-check-remote-containers:
-	TESTED_IMAGES="$(TESTED_IMAGES)" tests/remote-containers.sh
-
-check: check-failures check-squash check-latest-imagestream check-remote-containers
+check: check-failures check-squash check-latest-imagestream
```
zmiklank, line 50: Empty `Makefile` targets caused us some problems before. Could you please remove the target entirely or add some content to it?

Makefile
```diff
@@ -28,6 +21,10 @@ test-lib-foreach:
 check-test-lib: $(TEST_LIB_TESTS)
 
 test: check
+	TESTED_SCENARIO=test tests/remote-containers.sh
+
+test-openshift-4: check
```
zmiklank, line 26: Shouldn't be this targer added to .PHONY ones? Probably also shellcheck,  test-lib-foreach  (this one probably should not exist at all), pre-commit-check, check-squash, check-latest-imagestream, check-betka targets should be marked as phony.

Makefile
phracek: Good point. I have refactored it a bit

Makefile
phracek: .PHONY added.

phracek:
rebased

[test]

PR #275, opened 2022-07-21
Check if files generated by dist-gen are generated properly

phracek:
This new function checks if files generated by dist-gen are committed into each pull request properly.
The function is called only if the manifest.sh file is present in the root directory of a container.

The command which checks the status of generated files is:

```
git diff --ignore-submodules=all --exit-code
```

the common submodule is not checked at all.

Nowadays we have three containers that use dist-gen generated sources:

    varnish-container
    postgresql-container
    s2i-python-container

The old PR has been closed because it had no reference https://github.com/sclorg/container-common-scripts/pull/218


phracek:
[test]

phracek:
[test]

phracek:
See also comment here: https://github.com/sclorg/container-common-scripts/pull/218/files/c9d77935c6b6211d9e0062436341c18088113bc7

phracek:
@pkubatrh Would you be willing to review it?

test.sh
```diff
@@ -8,6 +8,25 @@
 
 set -e
 
+
+if [[ -f "./manifest.sh" ]]; then
```
zmiklank, line 12: `manifest.sh` will no longer be used for python https://github.com/sclorg/s2i-python-container/pull/539. If the new solution works in python, the plan is to use it in all distgen-generated container repositories.
We can, of course, use this solution till then, but let's count with a fact, that this is temporary.

test.sh
```diff
@@ -8,6 +8,25 @@
...
+  if ! git diff --ignore-submodules=all --exit-code; then
+    git show -s
+    echo "Pull request does not contain properly generated files."
+    echo "Please run these command in order to update pull request:"
+    echo "'make clean-versions'"
+    echo "'make generate-all'"
+    exit 1
+  fi
+else
+  echo "This container does not use dist-gen. Skipping it."
```
zmiklank, line 27:
```suggestion
  echo "This container does not use dist-gen. Skipping a check for generating files by the dist-gen."
```

I would suggest to explain more what are we skipping here.

test.sh
```diff
@@ -8,6 +8,25 @@
...
+if [[ -f "./manifest.sh" ]]; then
+  echo "Files are generated by dist-gen. Check if everything generated by distgen properly."
+  echo "Generating all files by 'make generate-all' ..."
+  make generate-all
+  echo "Files are generated by distgen."
+  # common submodule is not updated at all. No argument to git diff is needed.
+  if ! git diff --ignore-submodules=all --exit-code; then
+    git show -s
+    echo "Pull request does not contain properly generated files."
+    echo "Please run these command in order to update pull request:"
```
zmiklank, line 21: What about adding a information, that the distgen-generated files should be added in a separate commit?

zmiklank (changes requested):
Would this work in python container, where files for minimal container are generated manually, and not by the `manifest.sh` rules?

zmiklank:
> Would this work in python container, where files for minimal container are generated manually, and not by the `manifest.sh` rules?

I tried that and proceed with the steps hinted in this PR:
```
'make clean-versions'
'make generate-all'
```
and I deleted the whole 3.9-minimal directory.

test.sh
phracek: Ok, make sense to me, but we have to rewrite it also for `varnish-container` and `postgresql-container`.

test.sh
phracek: Make sense to me.

test.sh
phracek: Or we can check if `manifest.sh` or `manifest.yaml` exists.

zmiklank: There can also be files named like [manifest-minimal.yml](https://github.com/sclorg/s2i-python-container/pull/539/files#diff-4e820cda54f76392ca641e6eb4b84a36ae0d5c50befa3444f12fc08a3ac70805).

zmiklank: But well, yes, we could check for `manifest*.yaml`

zmiklank: Also the Makefile rule is called `generate-new` here and not `generate-all`.

zmiklank: I would wait with these changes until the https://github.com/sclorg/s2i-python-container/pull/539 is merged.

frenzymadness:
I see two ways forward here:
* Implement the new generator everywhere - That'd mean taking the PR from Python container repo, implementing it in the common scripts and switching all the containers using distgen at once. This is kinda risky but allows this PR to work with all of the containers at once.
* Ignore Python for now - Finish this PR and ignore Python container for now (don't detect `manifest.yml`). We'll test the new generator for some time only for Python container and when we think it's ready, we can then move it to the common scripts and change the detection mechanism here as well.

What do you think and prefer?

zmiklank:
> I see two ways forward here:
> 
>     * Implement the new generator everywhere - That'd mean taking the PR from Python container repo, implementing it in the common scripts and switching all the containers using distgen at once. This is kinda risky but allows this PR to work with all of the containers at once.
> 
>     * Ignore Python for now - Finish this PR and ignore Python container for now (don't detect `manifest.yml`). We'll test the new generator for some time only for Python container and when we think it's ready, we can then move it to the common scripts and change the detection mechanism here as well.
> 
> 
> What do you think and prefer?

I agree and prefer 2).

test.sh
phracek: > Also the Makefile rule is called `generate-new` here and not `generate-all`.

Sorry I don't see `generate-new` I see only `generate-all` https://github.com/sclorg/container-common-scripts/blob/master/common.mk#L147

phracek:
Rebased

[test]

test.sh
zmiklank: I meant in the new proposal from Lumir.

frenzymadness: We agreed to ignore generate-new and manifest.yml for now.

PR #276, opened 2022-07-26
add script for updating common/ in container repositories

zmiklank:
The script would pull specified SCLORG container image git repositories,
commits there an update of common/ submodule to the latest origin/master
commit or a specified commit. Then the changes would be pushed to the
repositories with the user's ssh key.

The script should meet the following rules:
 - run by the user, not a bot
 - be easy to configure and run
    - in most cases should be sufficient to run `make push-to-containers`
 - user is able to choose which container repositories to update
    - with editing array with containers
 - doublecheck if the commit is OK to push
 
 WDYT?


phracek:
@zmiklank One nitpick ;) Why is it not written in Python? Tests are missing and it would be easier to write them in python. We need some GitHub token or credentials for pushing PR, right? Do you count on testing them?

zmiklank:
@phracek, thanks for the review.

I've chosen shell, as all of the container-common-scripts are written in shell.
I can of course add some tests, as you suggested.

We do not need token on credentials, as this script is meant to be run from localhost, using the credentials of the user.

Makefile
```diff
@@ -46,3 +46,7 @@ check-latest-imagestream:
 
 check-betka:
 	cd tests && ./check_betka.sh
+
+push-to-containers:
+	@echo "THIS COULD BE DANGEROUS, WILL PUSH TO ALL SCLORG CONTAINER REPOSITORIES"
+	./push_as_submodule.sh
```
phracek, line 52: Do we really want to name it push_as_submodule?
I would propose something like `push_sclorg_repos`.

Or is this script closely related to only common? I guess we would like to use it also e.g. Update GitHub actions ;)

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+  local reason=${2:-"unknown"}
+  if [[ $ret -ne 0 ]]; then
+    echo "ERR: Exiting for reason: $reason" 2>&1 && exit 42
+  fi
+}
+
+function clone_repo() {
+  local image=$1
+  [[ -z $image ]] && exit_on_err 1 "image name for cloning repository is missing"
+  local container="${image}-container"
```
phracek, line 60: This is not needed. Let's define it directly in `all_sclorg_images`.

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+  local url="git@github.com:sclorg/${container}.git"
+  git clone "$url" "$image"
```
phracek, line 62: `$image` should be full repository name and not short one.

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+remote=origin
+commit_msg="Update common submodule to current latest $remote/$branch commit"
+
+#----------------------------------
+
+function exit_on_err() {
+  local ret=$1
+  local reason=${2:-"unknown"}
+  if [[ $ret -ne 0 ]]; then
+    echo "ERR: Exiting for reason: $reason" 2>&1 && exit 42
```
phracek, line 53: Why exit 42? it should be enough `exit 1`.
Do you have any reason for it?

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+hash=$1; [[ -z $hash ]] && hash=$(get_remote_branch_hash)
+cur_dir=$PWD
+tmp_dir=$(mktemp -d) && cd "$tmp_dir" || exit 1
+updated_containers=""
+trap cleanup SIGINT EXIT
+
+echo "Updating common/ submodule in following repositories: ${all_sclorg_images[*]}"
+for image in "${all_sclorg_images[@]}"; do
+  clone_repo "$image"
+  cd "$image" || exit 1
```
phracek, line 143: In the case of `exit 1` we should switch back to `PWD`. Any idea why to stay in `cd ..` . It should be `popd`

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+
+echo "This script is going to pull from and push to multiple git repositories."
+echo "Therefore it is recommended to use ssh-agent."
+echo "You can pass SHA of commit to update the submodules to as a parameter."
+echo "If it is not passed, current HEAD of $remote/$branch is used."
+echo "Press Enter to continue." ; read -r
+
+hash=$1; [[ -z $hash ]] && hash=$(get_remote_branch_hash)
+cur_dir=$PWD
+tmp_dir=$(mktemp -d) && cd "$tmp_dir" || exit 1
```
phracek, line 136: I would use `pushd` instead of `cd`

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+updated_containers=""
+trap cleanup SIGINT EXIT
+
+echo "Updating common/ submodule in following repositories: ${all_sclorg_images[*]}"
+for image in "${all_sclorg_images[@]}"; do
+  clone_repo "$image"
+  cd "$image" || exit 1
+  # following steps count with a fact, that $PWD==$image
+  update_submodule "$hash"
```
phracek, line 145: In case of error, do we want to really stay in `cd "$image"` repository?

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+  git fetch -a
+  git checkout "$sha"
+  exit_on_err $? "checkout common on $sha failed"
+  cd ..
+  git add common/
+}
+
+function commit_change() {
+  local commit_msg=$1
+  [[ -z "$commit_msg" ]] && exit_on_err 1 "commit message is missing"
```
phracek, line 83: This is not needed, because `commit_msg` is defined at the beginning.

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+function show_diff() {
+  local image=$1
+  local answer
+  git diff $branch $remote/$branch
+  exit_on_err $? "diff failed"
+  echo "Are you sure, you want to push the above changes to $remote/$branch of"
+  echo "git@github.com:sclorg/${image}-container.git?"
+  echo "Y/n"
+  read -r answer
+  while [[ $answer != Y && $answer != n ]]; do
```
phracek, line 96: Some explanation message in case of `y` or `N` would be nice. Or use both variants

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+
+echo "Updating common/ submodule in following repositories: ${all_sclorg_images[*]}"
+for image in "${all_sclorg_images[@]}"; do
+  clone_repo "$image"
+  cd "$image" || exit 1
+  # following steps count with a fact, that $PWD==$image
+  update_submodule "$hash"
+  commit_change "$commit_msg"
+  push_change "$image" && updated_containers="$updated_containers $image"
+  cd ..
```
phracek, line 148: I would prefer to use `cd $PWD` here.

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+  read -r answer
+  while [[ $answer != Y && $answer != n ]]; do
+    read -r answer
+  done
+  [[ "$answer" == "n" ]] && return 1
+  return 0
+}
+
+function push_change() {
+  local image=$1
```
phracek, line 104: Can you please add there some comments, what the function does?
`show_diff` shows submodule or container repository?

phracek: I have missed it. The container repository. Ignore these comments.

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
+function get_remote_branch_hash() {
+  git fetch "$remote"
+  git rev-parse "$remote"/"$branch"
+  exit_on_err $?
+}
+
+function cleanup() {
+  echo "Updated container images are: $updated_containers"
```
phracek, line 118: Would it be possible to remember all pull requests and show them at the end?
Only RFE.

push_as_submodule.sh
```diff
@@ -0,0 +1,149 @@
...
+  cd "$image" || return 1
+  git submodule update --init
+  exit_on_err $? "submodule init failed"
+  cd ..
+}
+
+function update_submodule() {
+  local sha=$1
+  [[ -z "$sha" ]] && exit_on_err 1 "commit SHA is missing for updating submodule"
+  cd common/ || return 1
```
phracek, line 73: the return 1 is not needed, because if `git submodule update --init` fails, then common directory does not exist. See `clone_repo` function.

phracek (changes requested):

push_as_submodule.sh
zmiklank: OK.

push_as_submodule.sh
zmiklank: Just a small [joke](https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker%27s_Guide_to_the_Galaxy). Will rewrite it.

push_as_submodule.sh
zmiklank: OK

Makefile
zmiklank: Yes. I can not imagine any other situation (than updating submodules), that will require pushing the same patch to more container repositories. The repositories are too different and the patch would not be applicable in the vast majority of cases.

push_as_submodule.sh
zmiklank: see explanation above.

push_as_submodule.sh
zmiklank: I did it like this, for the script to be more generic and expendable. (E.g., someone would like to pass the commit-msg from the command line in the future.)

If you do not agree with this argumentation I can, of course, rewrite the script.

push_as_submodule.sh
zmiklank: The explanation is on lines 91-93, do you think I should extend it?

push_as_submodule.sh
zmiklank: If we use `cd $PWD` would stay in `$image` dir. I would then suggest removing the `cd` command from here entirely.

push_as_submodule.sh
zmiklank: This will not create PRs. This will push changes directly to master. Is this a problem?

push_as_submodule.sh
zmiklank: This is only a correctness check. I think it is valid. Maybe someone would decide to rename `common` and then we would need this check.

Could you please explain your though more, maybe I misunderstood?

zmiklank:
@phracek thanks again, could you please look into the replies I wrote?

Also, I changed the container image name format from `s2i-base` to `s2i-base-container`, etc.

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

phracek:
[test-all]

phracek (approved):
Thanks for the PR! LGTM for now, let's get merge and is how it works in real life.

phracek:
@zmiklank Please rebase this pull request against master. Thank you.

phracek:
[test]

PR #277, opened 2022-07-26
make squash work for podman, not for docker in the testsuite

zmiklank:
NOTE: podman is able to run "podman build --squash $context", but docker is not. In order to squash image during the build with docker, its daemon needs to have experimental features enabled:
```
echo '{"experimental": true}' >> /etc/docker/daemon.json
```

Followup:
After some discussion within the team we decided, that it is not our goal to support squashing with docker, thus we will continue to test squash only with podman.
This results in removing docker-squash functionality and introducing `build --squash` option, which is to be enabled if podman is used, and if `SKIP_SQUASH` is set to 0


zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

build.sh
```diff
@@ -170,60 +164,22 @@ function docker_build_with_version {
 
   pull_image "$dockerfile"
 
+  is_podman() {
+    local _bin
+    _bin=$(docker --version | cut -d" " -f1)
```
pkubatrh, line 169: This will work with `podman-docker`, but not when a symlink is used (eg. `/usr/bin/docker -> podman`) on systems where podman-docker might not be available. Not sure if we want to make sure this case is covered as well, but mentioning anyway.

build.sh
```diff
@@ -170,60 +164,22 @@ function docker_build_with_version {
...
+    [[ $_bin == "podman" ]] && echo true
```
pkubatrh, line 170: Why do we echo true here?

build.sh
```diff
@@ -170,60 +164,22 @@ function docker_build_with_version {
...
+  }
+
+  if [[ "$SKIP_SQUASH" == 0 ]] && is_podman; then
+    BUILD_OPTIONS+=" --squash"
+  fi
   # shellcheck disable=SC2016
   parse_output 'docker build '"$BUILD_OPTIONS"' -f "$dockerfile" "${DOCKER_BUILD_CONTEXT}"' \
                "tail -n 1 | awk '/Successfully built|(^--> )?(Using cache )?[a-fA-F0-9]+$/{print \$NF}'" \
                IMAGE_ID
-  clean_image
```
pkubatrh, line 177: Why remove the `clean_image` call over here?

clean.sh
```diff
@@ -1,16 +1,12 @@
 #! /bin/sh
 
-set -e
```
pkubatrh, line 3: Since we are removing this, we might as well remove the `|| :` sections as well.

zmiklank: Good point, thanks. I will remove that.

build.sh
zmiklank: You are right, I will just return 1 if not podman.

zmiklank:
[test-all]

build.sh
zmiklank: I really did not expect this behavior. I will try to rethink this, so that also `/usr/bin/docker -> podman` is usable.

build.sh
zmiklank: A mistake.

zmiklank:
[test-all]

zmiklank:
I have reworked the PR based on @pkubatrh's review. Please re-review, if tests pass.

zmiklank:
Unfortunately, they seem to fail. I will debug that.

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
Testing is not possible if fast-forward merge is not possible. As something was merged to master before last run of tests, they are all invalid.
I have rebased on master and now am rerunning the tests.
[test-all]

phracek:
@zmiklank I would prefer before pushing execute `make shellcheck`.
Sanity test failures:

```bash
./run-shellcheck.sh `git ls-files *.sh`

In build.sh line 186:
  pushd "${dir}" > /dev/null
  ^------------------------^ SC2164: Use 'pushd ... || exit' or 'pushd ... || return' in case pushd fails.

Did you mean: 
make: *** [Makefile:28: shellcheck] Error 123
  pushd "${dir}" > /dev/null || exit


In build.sh line 205:
  popd > /dev/null
  ^--------------^ SC2164: Use 'popd ... || exit' or 'popd ... || return' in case popd fails.

Did you mean: 
  popd > /dev/null || exit


In tag.sh line 14:
  pushd "${dir}" > /dev/null
  ^------------------------^ SC2164: Use 'pushd ... || exit' or 'pushd ... || return' in case pushd fails.

Did you mean: 
  pushd "${dir}" > /dev/null || exit


In tag.sh line 29:
  popd > /dev/null
  ^--------------^ SC2164: Use 'popd ... || exit' or 'popd ... || return' in case popd fails.

Did you mean: 
  popd > /dev/null || exit
```

zmiklank:
@phracek, the error logs you have posted are not related to this PR. I did not add/adjust any of the mentioned lines.

This should be addressed elsewhere.

zmiklank:
It seems, that tests error were caused by the impossibility of ff merge. Removing debugging options and retesting:
[test-all]

zmiklank:
[test-all]

build.sh
zmiklank: I need to remove the `clean_image` call, because if we clean the built image we will not have an image to test on.

zmiklank:
This is stuck on the impossibility to build `s2i-core-container` because of an invalid reference to some s2i stuff in the image.
This needs to be handled separately.

zmiklank:
[test-all]

build.sh
```diff
@@ -170,60 +164,20 @@ function docker_build_with_version {
 
   pull_image "$dockerfile"
 
+  docker info 2>/dev/null | grep podman 1>/dev/null
+  is_podman=$?
+
+  if [[ "$SKIP_SQUASH" == 0 ]] && [[ $is_podman -eq 0 ]]; then
```
frenzymadness, line 170: One condition uses `==` and the second one uses `-eq`, not a real problem, just a nitpick.

tests/remote-containers.sh
```diff
@@ -43,16 +45,39 @@ analyse_commits ()
     done < <(git log --format=%B --reverse "$MERGE_INTO"..HEAD)
 }
 
-analyse_commits
+archive_common()
+{
+  local tar=$1
```
frenzymadness, line 50: What about something like archive_name or archive_path instead of tar?

tests/remote-containers.sh
```diff
@@ -43,16 +45,39 @@ analyse_commits ()
...
+  printf -v test_short_summary "${test_short_summary}${result_str} $TESTED_IMAGE\n"
+
+  echo "$test_short_summary"
+  if [ $TESTSUITE_RESULT -eq 0 ] ; then
+    echo "Tests for 'container-common-scripts' succeeded."
+  else
+    echo "Tests for 'container-common-scripts' failed."
+  fi
+
+  rm $common_tar
```
frenzymadness, line 66: $common_archive maybe?

frenzymadness (approved):
I like the solution. I've just proposed some possible naming improvements.

zmiklank:
> I like the solution. I've just proposed some possible naming improvements.

Thanks for reviewing, I have fixed the found issues in new commits. 
However, before merging I need to look at the tests again.

zmiklank:
[test-all]

phracek (changes requested):
The tests are failing and container repos are not executed:

```bash
db77ca1a51c0: Pull complete
Digest: sha256:bec155a4ff846d6c395a7ba1d9d4643017d281b2c5816e5207baf324f63df158
Status: Downloaded newer image for registry.access.redhat.com/rhscl/s2i-core-rhel7:latest
make[2]: Leaving directory `/tmp/sclorg/container-common-scripts/tests/failures/check'
make[2]: *** [v0] Error 1
make[1]: *** [build-serial] Error 2
make[1]: Leaving directory `/tmp/sclorg/container-common-scripts/tests/failures/check'
make: *** [check-failures] Error 2
```

Can you please fix it?

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

phracek:
[test-all]

phracek (approved):
Everything LGTM. Let's wait for tests

PR #287, opened 2022-09-01
Let's use only one OpenShift 4 client version 

phracek:
Let's use only one OpenShift 4 client version located in /usr/local/oc-v4/bin directory


zmiklank:
Why are we changing OC version from 4.4 to 4?

phracek:
> Why are we changing OC version from 4.4 to 4?

@zmiklank Thanks for this question. We would like to update OpenShift Client from 4.4 to 4.10 and therefore let's have only one directory. We really do not need a lot of `oc` versions.

zmiklank:
Ok, thanks for an explanation. LGTM.

zmiklank (approved):

PR #288, opened 2022-09-05
Add support for testing single_version

phracek:
Add support for testing single_version in case `make test VERSIONS=` is called
then no version is tested.

This pull request is needed by
https://github.com/sclorg/sclorg-testing-farm/pull/37

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -8,6 +8,12 @@
 
 set -e
 
+# This adds backwards compatibility if only single version needs to be testing
+# In CI we would like to test single version but VERSIONS= means, that nothing is tested
+# make test TARGET=<OS> VERSIONS=<something> ... checks single version for CLI
+# make test TARGET=<OS> SINGLE_VERSION=<something> ... checks single version from Testing Farm
+VERSIONS=${SINGLE_VERSION:-$VERSIONS}
+
 for dir in ${VERSIONS}; do
   [ ! -e "${dir}/.image-id" ] && echo "-> Image for version $dir not built, skipping tests." && continue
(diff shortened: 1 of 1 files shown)


phracek:
nginx-container PR for testing single versions https://github.com/sclorg/nginx-container/pull/209

zmiklank:
I'll take a look, thanks.

phracek:
[test]

test.sh
```diff
@@ -8,6 +8,12 @@
 
 set -e
 
+# This adds backwards compatibility if only single version needs to be testing
+# In CI we would like to test single version but VERSIONS= means, that nothing is tested
+# make test TARGET=<OS> VERSIONS=<something> ... checks single version for CLI
+# make test TARGET=<OS> SINGLE_VERSION=<something> ... checks single version from Testing Farm
+VERSIONS=${SINGLE_VERSION:-$VERSIONS}
```
zmiklank, line 15: Is this necessary?
Wouldn't it suffice to set just `VERSIONS=$VERSION` in the TMT plan? Then no changes would be needed in container-common-scripts.

phracek: Good point. But if the `VERSION` is not specified then `VERSIONS` will be empty and nothing is tested.
The VERSIONS are specified in Makefile and the TMT plan overrides it. And therefore we have a problem.

zmiklank (approved):

test.sh
zmiklank: Thanks. I did not catch it from the comment, but now I understand.

PR #291, opened 2022-09-08
Fix CVP pipeline issues.

phracek:


phracek:
[test]

zmiklank (commented):
Could you please explain a bit in a commit message, what is happening in the commit?

phracek:
@zmiklank Rebased and add more descriptive commits.

test-lib-openshift.sh
```diff
@@ -1250,9 +1250,11 @@ function ct_os_test_image_stream_template() {
     # Most functions depend on the set -e anyway at this point.
   fi
   ct_os_wait_pod_ready "${service_name}" 120
+  result=$?
```
zmiklank, line 1253: I would say, that checking the result is needed here, something like `if result !=0; return 1;`

phracek: We have to delete objects from the project.
return is defined here: https://github.com/sclorg/container-common-scripts/pull/291/files#diff-1a059cdb3760c3d0e33b1ef71241a1f88e1f4e8d7390d29bd5f86a44b090ec4aR1257

phracek:
[test]

zmiklank (approved):
lgtm

PR #292, opened 2022-09-12
Replace s2i-nodejs-container with nginx-container and s2i-perl-container

phracek:
Because s2i-nodejs-container is still failing on `pino` upstream problem,

let's get remove it with s2i-perl and nginx-containers which are stable.

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container-to-test: ["postgresql-container", "s2i-nodejs-container", "s2i-python-container"]
+        container-to-test: ["postgresql-container", "nginx", "s2i-perl-container", "s2i-python-container"]
         os: ["fedora", "centos7", "rhel7", "rhel8", "rhel9"]
 
     if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/container-tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container-to-test: ["postgresql-container", "s2i-nodejs-container", "s2i-python-container"]
+        container-to-test: ["postgresql-container", "nginx", "s2i-perl-container", "s2i-python-container"]
```
pkubatrh, line 15:
```suggestion
        container-to-test: ["postgresql-container", "nginx-container", "s2i-perl-container", "s2i-python-container"]
```
Missing container suffix?

zmiklank, line 15:
```suggestion
        container-to-test: ["postgresql-container", "nginx-container", "s2i-perl-container", "s2i-python-container"]
```

pkubatrh:
lgtm. Lets add node back once the pino issue is resolved.

.github/workflows/container-tests.yml
zmiklank: Looks like I am a few seconds late :).

phracek: Stupid error ;)

Issue #278, opened 2022-08-01
Do shellcheck on PR code


zmiklank:
- find out if https://github.com/redhat-plumbers-in-action/differential-shellcheck is usable for us.
- needs further brainstorming

zmiklank:
Shellcheck has already been integrated. One can run it by writing [test] or [test-all] to PR's comment.

PR #293, opened 2022-09-16
ci(lint): add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs **differential ShellCheck scans** on shell scripts changed via PR and reports results directly in PR. The beauty of differential scans is that you will get **reports only about newly added defects**. 

Since this repository has many shell scripts, I think you might find some value in having a shell linter. Differential ShellCheck action is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

> **Note**: I have also removed `.travis.yml`. It seems it is no longer used.

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,22 @@
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,0 @@
-language: shell
-os: linux
-script:
-  - make shellcheck
-  - make pre-commit-check


jamacku:
It might be helpful for contributors that would forget to run `make shellcheck` locally (as noted in [sclorg/container-common-scripts#277](https://github.com/sclorg/container-common-scripts/pull/277#issuecomment-1210497736)), and they would still get notification about newly introduced shellcheck defects. And when you get to review, defects might be already addressed.

zmiklank:
Thank you for the contribution. We already talked about integrating the differential shell check, see https://github.com/sclorg/container-common-scripts/issues/278 (but more people knew about it, not just me :)).

However, we already have a shellcheck running on demand with other tests on PR as part of SanityTest GitHub Action (see e.g., [here](https://github.com/sclorg/container-common-scripts/actions/runs/3045825552/jobs/4907867197)).

I see a big advantage of checking only PR changes when using differential-shellcheck.
This would be useful for example in the PR you mentioned above (https://github.com/sclorg/container-common-scripts/pull/277#issuecomment-1210497736).

I personally am for removing our existing shellcheck GitHub action and start using this one. One more think to consider, is that out existing sanity-check GitHub Action also executes a `pre-commit-check`, which should not be removed.

@phracek WDYT?

I will also try to run our tests, so that I can see how the results will be displayed, when the shellcheck results are already present.
[test-all]

phracek:
I totally agree with @zmiklank  From my point of view it looks good to me.
I would prefer to add this one check but let's not remove our spellcheck.

@jamacku Can you please rebase this pull request against master?

jamacku:
> @jamacku Can you please rebase this pull request against master?

Done :+1:

zmiklank (approved):

phracek:
[test-all]

phracek (approved):
Thanks. LGTM. Let's wait for the tests

jamacku:
@zmiklank, @phracek, would you be interested in `differential-shellcheck` GitHub Action in other repositories with shell scripts?

zmiklank:
@jamacku I am not aware of any other repository that would need shellcheck running, thanks.

PR #297, opened 2022-09-21
test-lib: add test for run_all_tests function

zmiklank:
The test tests if testsuite ends with error if any ran test failed and if it ends with success if all tests pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ TEST_LIB_TESTS = \
 	random_string \
 	test_npm \
 	image_availability \
+	run_all_tests\
 	public_image_name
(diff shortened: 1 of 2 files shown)


zmiklank:
[test-all]

zmiklank:
> Thanks for this great pull request. Would it be possible to add there also UNSTABLE tests?

I would rather do that in different testcase. Preferably also in other PR, when I find time to do it. But great idea, I will put it on my TODO list.

zmiklank:
[test-all]

zmiklank:
[test-all]

PR #299, opened 2022-09-30
add possibility for enabling debug mode in building and testing scripts

zmiklank:
This PR porposes an option to enable debug mode in the building, testing, tagging and cleaning scripts.


zmiklank:
[test-all]

phracek:
@zmiklank I agree with debug mode, but all GitHub actions have to be changed as well. What about adding into GitHubAction variable for it? https://github.com/sclorg/testing-farm-as-github-action/blob/main/action.yml#L128. Adding more and more variables and fixing GitHub actions can be somehow generalized. What about checking if the comment contains [something=value] and then parsing it and adding it to GitHub action?

zmiklank:
[test-all]

phracek (approved):
LGTM.

PR #303, opened 2022-11-07
test-lib: check if  $CID_FILE_DIR not empty in ct_cleanup

zmiklank:
Partially fixes https://github.com/sclorg/varnish-container/issues/111

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -32,30 +32,35 @@ UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
 # Uses: $TESTSUITE_RESULT - overall result of all tests
 function ct_cleanup() {
   ct_show_resources
+  ct_show_results
+
+  if [[ -z ${CID_FILE_DIR:-} ]]; then
+    echo "The \$CID_FILE_DIR is not set. Container cleaning is to be skipped."
+    exit "${TESTSUITE_RESULT:-0}"
+  fi;
+
+  echo "Examining CID files in $CID_FILE_DIR"
   for cid_file in "$CID_FILE_DIR"/* ; do
     [ -f "$cid_file" ] || continue
     local container
     container=$(cat "$cid_file")
 
-    : "Stopping and removing container $container..."
+    echo "Stopping and removing container $container..."
     docker stop "$container"
 
     # Container has not been removed by `docker stop` and still exists
     if [ "$( docker ps -a -f "id=$container" | wc -l )" -eq 2 ]; then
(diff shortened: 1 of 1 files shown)


zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

test-lib.sh
```diff
@@ -31,29 +31,38 @@ UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
...
-    : "Stopping and removing container $container..."
-    docker stop "$container"
-
-    # Container has not been removed by `docker stop` and still exists
-    if [ "$( docker ps -a -f "id=$container" | wc -l )" -eq 2 ]; then
-      exit_status=$(docker inspect -f '{{.State.ExitCode}}' "$container")
-      if [ "$exit_status" != "$EXPECTED_EXIT_CODE" ]; then
-        : "Dumping logs for $container"
-        docker logs "$container"
+  clean_containers() {
```
phracek, line 34: Is there any reason for subfunction?
What about starting with

```
  if [[ -z ${CID_FILE_DIR:-} ]]; then
    echo "The \$CID_FILE_DIR is not set. Container cleaning is to be skipped."
    return
  fi;
```
and run cleanup in case of CIF_FILE_DIR is set.

phracek (changes requested):
Thanks for fixing this issue. I have a comment, why subfunction is used? Does it make sense?

zmiklank:
[test-all]

phracek (approved):

phracek (approved):
LGTM now thanks.

test-lib.sh
```diff
@@ -32,30 +32,35 @@ UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
 # Uses: $TESTSUITE_RESULT - overall result of all tests
 function ct_cleanup() {
   ct_show_resources
+  ct_show_results
+
+  if [[ -z ${CID_FILE_DIR:-} ]]; then
```
phracek, line 37: Is this correct? I have missed that.

phracek (changes requested):

phracek (changes requested):
Please see my comment. Do not merge it yet. I guess $ should be there.

test-lib.sh
zmiklank: Why you think this is not correct? Should I use `test` instead of `[[`?

PR #305, opened 2022-11-10
add test for checking distgen generated files

zmiklank:
This test will not be run during the standard testsuite as it is intended only for container repositories with distgen generated files.
Another PRs for python, varnish and postgresql container will follow when this is merged.

This PR does not need to be tested because current testsuite does not cover the added code.


tests/check_distgen_generated_files.sh
```diff
@@ -0,0 +1,28 @@
...
+# This test is to be used only in container repositories in which the
+# container files are generated by the distgen tool.
+# This tests checks if files present in the repository are the same as
+# freshly generated files.
+# The commands to make freshly generated files are:
+#   'make clean-versions'
+#   'make generate-all'
+#
+# The resposibility to regenerate all the needed files lies with
+# the Pull Requests's author, who need to add them in a separate commit
```
frenzymadness, line 12: `Pull Requests' authors` or `Pull Request's author`

frenzymadness (changes requested):
I think it's very good. I've just found some typos. Do you want me to test it in Python container image?

zmiklank:
Thanks for the review, I have addressed your notes.

> Do you want me to test it in Python container image?

I have tested this manually already, so it does not seem to be necessary.

PR #306, opened 2022-11-10
clean all application images built during tests

zmiklank:
- introducing init function for ct tests

The changes in all containers which will be needed can be seen in this exemplary PR: https://github.com/sclorg/nginx-container/pull/224

Edit: added commit for preserving backwards compatibility.

### DIFF
--- a/clean.sh
+++ b/clean.sh
@@ -6,13 +6,18 @@ test -f auto_targets.mk && rm auto_targets.mk
 
 for version; do
   for id_file in .image-id .image-id-from; do
-    if [ -f "$version"/"$id_file" ]; then
-      image="$(cat "$version"/"$id_file")"
+    id_path="$version"/"$id_file"
+    if [ -f "$id_path" ]; then
+      image="$(cat "$id_path")"
       test -n "$image" || continue
+      docker inspect "$image" > /dev/null 2>&1 || continue
       containers="$(docker ps -q -a -f ancestor="$image")"
-      docker rm -f "$containers" 2>/dev/null
+      if [ -n "$containers" ]; then
+        docker stop "$containers"
+        docker rm -f "$containers"
+      fi
       docker rmi -f "$image"
-      rm -f "$version"/"$id_file"
+      rm -f "$id_path"
     fi
   done
 done
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -74,24 +74,45 @@ function ct_cleanup() {
     [ -f "$cid_file" ] || continue
     local container
     container=$(cat "$cid_file")
+    rm "$cid_file"
+
+    ct_container_exists "$container" || continue
 
     echo "Stopping and removing container $container..."
-    docker stop "$container"
-
-    # Container has not been removed by `docker stop` and still exists
-    if [ "$( docker ps -a -f "id=$container" | wc -l )" -eq 2 ]; then
-      exit_status=$(docker inspect -f '{{.State.ExitCode}}' "$container")
-      if [ "$exit_status" != "$EXPECTED_EXIT_CODE" ]; then
-        echo "Dumping logs for $container"
-        docker logs "$container"
-      fi
-      docker rm -v "$container"
+    if ct_container_running "$container"; then
+      docker stop "$container"
     fi
-    rm "$cid_file"
+
+    exit_status=$(docker inspect -f '{{.State.ExitCode}}' "$container")
+    if [ "$exit_status" != "$EXPECTED_EXIT_CODE" ]; then
+      echo "Dumping logs for $container"
+      docker logs "$container"
+    fi
+    docker rm -v "$container"
   done
 
   rmdir "$CID_FILE_DIR"
   exit "${TESTSUITE_RESULT:-0}"
+}
+
+# ct_container_running
+# --------------------
+# Return 0 if given container is in running state
+# Uses: $1 - container id to check
+function ct_container_running() {
+  local running
+  running="$(docker inspect -f '{{.State.Running}}' "$1")"
+  [ "$running" = "true" ] || return 1
+}
+
+# ct_container_exists
+# --------------------
+# Return 0 if given container exists
+# Uses: $1 - container id to check
+function ct_container_exists() {
+  local exists
+  exists="$(docker ps -q -f "id=$1")"
+  [ -n "$exists" ] || return 1
 }
 
 # ct_clean_app_images
@@ -100,14 +121,16 @@ function ct_cleanup() {
 # Uses: $APP_ID_FILE_DIR - path to directory containing cid_files
 function ct_clean_app_images() {
   local image
-  if [[ ! -d ${APP_ID_FILE_DIR:-} ]]; then
+  if [[ ! -d "${APP_ID_FILE_DIR:-}" ]]; then
     echo "The \$APP_ID_FILE_DIR=$APP_ID_FILE_DIR is not created. App cleaning is to be skipped."
     return 0
   fi;
   echo "Examining image ID files in \$APP_ID_FILE_DIR=$APP_ID_FILE_DIR"
   for file in "${APP_ID_FILE_DIR:?}"/*; do
     image="$(cat "$file")"
-    docker rm -f "$(docker ps -q -a -f ancestor="$image")" 2>/dev/null
+    docker inspect "$image" > /dev/null 2>&1 || continue
+    containers="$(docker ps -q -a -f ancestor="$image")"
+    [[ -z "$containers" ]] || docker rm -f "$containers" 2>/dev/null
     docker rmi -f "$image"
   done
   rm -fr "$APP_ID_FILE_DIR"
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -24,6 +24,17 @@ EXPECTED_EXIT_CODE=0
 # is not mandatory for containers
 UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
 
+# following lines are added only for preserving backwards compatibility
+# and can be removed, when PR #306 is fully integrated to all container
+# repositories.
+TESTSUITE_RESULT=0
+APP_ID_FILE_DIR="/tmp/APP_ID_FILE_DIR_$RANDOM"
+# when removing this remove also all these lines:
+# mkdir -p "${APP_ID_FILE_DIR:?}"
+# as the $APP_ID_FILE_DIR should be created only in the ct_init function
+# ---------------------------------------------------------------------
+
+
 # ct_init
 # --------------------
 # This function needs to be called before any container test starts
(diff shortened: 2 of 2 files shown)


zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

test-lib.sh
```diff
@@ -4,24 +4,46 @@
...
 
 # define UNSTABLE_TESTS if not already defined, as this variable
 # is not mandatory for containers
 UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
 
+# following lines are added only for preserving backwards compatibility
+# and can be removed, when PR #306 is fully integrated to all container
+# repositories.
+TESTSUITE_RESULT=0
+APP_ID_FILE_DIR="/tmp/APP_ID_FILE_DIR_$RANDOM"
```
phracek, line 24: Variable $RANDOM is specified where?

test-lib.sh
```diff
@@ -4,24 +4,46 @@
...
-# ctest_doc_content"
 
 # Container CI tests
 # abbreviated as "ct"
 
+# run ct_init before starting the actual testsuite
+
 # may be redefined in the specific container testfile
 EXPECTED_EXIT_CODE=0
-export TESTSUITE_RESULT=0
```
phracek, line 19: Why not exported? It should be used either for OpenShift 4 tests, right?

test-lib.sh
```diff
@@ -64,6 +87,25 @@ function ct_cleanup() {
...
+# Uses: $APP_ID_FILE_DIR - path to directory containing cid_files
+function ct_clean_app_images() {
+  local image
+  if [[ ! -d ${APP_ID_FILE_DIR:-} ]]; then
+    echo "The \$APP_ID_FILE_DIR=$APP_ID_FILE_DIR is not created. App cleaning is to be skipped."
+    return 0
+  fi;
+  echo "Examining image ID files in \$APP_ID_FILE_DIR=$APP_ID_FILE_DIR"
+  for file in "${APP_ID_FILE_DIR:?}"/*; do
+    image="$(cat "$file")"
```
phracek, line 102: Test if image is not empty. For sure

test-lib.sh
```diff
@@ -434,7 +476,10 @@ FROM $IMAGE_NAME
 RUN command -v $binary | grep "$binary_path"
 EOF
   # Build an image, looking for expected path in the output
-  if ! docker build -f "$tmpdir/Dockerfile" --no-cache "$tmpdir"; then
+  local id_file
+  mkdir -p "${APP_ID_FILE_DIR:?}"
+  id_file="${APP_ID_FILE_DIR:?}"/"$RANDOM"
+  if ! docker build -f "$tmpdir/Dockerfile" --no-cache "$tmpdir" --iidfile "$id_file"; then
```
phracek, line 482: Will it work with podman? docker looks like to be obsoleted soon.

test-lib.sh
```diff
@@ -837,7 +882,10 @@ EOF
     read -ra mount_options <<< "$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*' || true)"
 
     # Run the build and tag the result
-    docker build ${mount_options[@]+"${mount_options[@]}"} -f "$df_name" --no-cache=true -t "$dst_image" .
+    local id_file
+    mkdir -p "${APP_ID_FILE_DIR:?}"
+    id_file="${APP_ID_FILE_DIR:?}"/"$RANDOM"
+    docker build ${mount_options[@]+"${mount_options[@]}"} --iidfile="$id_file" -f "$df_name" --no-cache=true -t "$dst_image" .
```
phracek, line 888: Let's use podman directly.

phracek (changes requested):
Thanks for the PR. What about using podman directly? Some notes from my point of view.

test-lib.sh
zmiklank: I think using podman directly needs further discussion also also maybe a separate PR as there could be a lot of changes.

test-lib.sh
zmiklank: from bash manpage:

```
RANDOM Each time this parameter is referenced, it expands to a random integer between
              0 and 32767.  Assigning a value to RANDOM initializes (seeds) the sequence  of
              random  numbers.  If RANDOM is unset, it loses its special properties, even if
              it is subsequently reset.
```

test-lib.sh
zmiklank: It is now in the init function.

zmiklank:
@phracek just for your info - this is still a work in progress - the PR is marked as draft, as something is not working as it should.

zmiklank:
[test-all]

zmiklank:
[test-all]

test-lib.sh
hhorak: Why random and not `mktemp -d`?

test-lib.sh
hhorak: we use `docker` everywhere now, so I agree with Zuzka, to keep using it further and if change to podman, then everywhere.

test-lib.sh
zmiklank: This is because I did not really comprehend what sourcing a bash script means and I thought that it is not possible.
But meanwhile I learnt that it should be possible, so I will rewrite it to mktemp -d.

zmiklank:
[test-all]

test-lib.sh
hhorak: Hm, that's the detail I missed -- doing something on a filesystem when the script is just sourced is not perfect. At minimum it should be always done only once (so maybe checking whether the variable is defined already and not reset it when the directory is created?).

However, reading the comment, as this is just a temporary solution until the new approach is integrated everywhere, let's not engineer it here :)

zmiklank: > Hm, that's the detail I missed -- doing something on a filesystem when the script is just sourced is not perfect.

`mktemp` offers `--dry-run` option, which only prints the name of temporary directory, but does not create it. This option can be used here.

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

test-lib.sh
```diff
@@ -40,37 +62,67 @@ UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
...
+  ct_clean_app_images
 
   if [[ -z ${CID_FILE_DIR:-} ]]; then
     echo "The \$CID_FILE_DIR is not set. Container cleaning is to be skipped."
     exit "${TESTSUITE_RESULT:-0}"
   fi;
 
-  echo "Examining CID files in $CID_FILE_DIR"
+  # return 0 if container exists
+  function container_exists() {
```
phracek, line 73: Why function in function? Can it be moved outside?

test-lib.sh
```diff
@@ -40,37 +62,67 @@ UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
...
   exit "${TESTSUITE_RESULT:-0}"
 }
 
+# ct_clean_app_images
+# --------------------
+# Cleans up application images referenced by APP_ID_FILE_DIR
+# Uses: $APP_ID_FILE_DIR - path to directory containing cid_files
+function ct_clean_app_images() {
+  local image
+  if [[ ! -d ${APP_ID_FILE_DIR:-} ]]; then
```
phracek, line 111:
```suggestion
  if [[ ! -d "${APP_ID_FILE_DIR:-}" ]]; then
```

test-lib.sh
```diff
@@ -40,37 +62,67 @@ UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
...
+    echo "The \$APP_ID_FILE_DIR=$APP_ID_FILE_DIR is not created. App cleaning is to be skipped."
+    return 0
+  fi;
+  echo "Examining image ID files in \$APP_ID_FILE_DIR=$APP_ID_FILE_DIR"
+  for file in "${APP_ID_FILE_DIR:?}"/*; do
```
phracek, line 116: I don't follow why is there ':?'. Based on the https://devhints.io/bash it means show error message but message is missing and exit if $APP_ID_FILE_DIR is unset.
This is not needed. Variable is defined in ct_init.

test-lib.sh
```diff
@@ -849,7 +904,10 @@ EOF
     read -ra mount_options <<< "$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*' || true)"
 
     # Run the build and tag the result
-    docker build ${mount_options[@]+"${mount_options[@]}"} -f "$df_name" --no-cache=true -t "$dst_image" .
+    local id_file
+    mkdir -p "${APP_ID_FILE_DIR:?}"
```
phracek, line 908: The same here. I don't see a reason for ':?'

phracek (changes requested):

test-lib.sh
zmiklank: This is the output if message is not set:
```
$ echo ${APP_ID_FILE_DIR:?}
bash: APP_ID_FILE_DIR: parameter null or not set
```
I thought it could be good to have this check for cases when the `ct_init` hasn't been called for some reason.

test-lib.sh
zmiklank: Fixed in https://github.com/sclorg/container-common-scripts/pull/306/commits/988e2c479cd2b2a113d5b64a911fab7deadd3f74.

zmiklank:
[test-all]

clean.sh
```diff
@@ -9,8 +9,12 @@ for version; do
     if [ -f "$version"/"$id_file" ]; then
```
phracek, line 9: "$version"/"$id_file" is used three times. What about to use a new variable.

Small nitpicking.

test-lib.sh
```diff
@@ -40,37 +62,70 @@ UNSTABLE_TESTS="${UNSTABLE_TESTS:-""}"
...
     container=$(cat "$cid_file")
+    rm "$cid_file"
 
     echo "Stopping and removing container $container..."
+    ct_container_exists "$container" || continue
     docker stop "$container"
 
     # Container has not been removed by `docker stop` and still exists
-    if [ "$( docker ps -a -f "id=$container" | wc -l )" -eq 2 ]; then
+    if ct_container_exists "$container"; then
```
phracek, line 84: This condition is not needed. You are already checking it a couple of rows above `ct_container_exists "$container" || continue`

clean.sh
zmiklank: You are right.

zmiklank:
[test-all]

zmiklank:
[test-all]

phracek (approved):
LGTM. Thanks for fixes.

PR #308, opened 2022-11-15
More verbose run_all_tests testcase

zmiklank:
and  [test-lib: move image size info to final show results](https://github.com/sclorg/container-common-scripts/commit/0e2c74dd91f2f7d95592edeef3b210a3f2b01ade)

### DIFF
--- a/tests/test-lib/run_all_tests
+++ b/tests/test-lib/run_all_tests
@@ -20,17 +20,29 @@ TEST_SET_POS="foo bar"
 TEST_SET_NEG="foo_neg bar"
 UNSTABLE_TESTS=""
 IGNORE_UNSTABLE_TESTS=0
+ret_val=0
 
-# positive test
+echo "running positive TC that should pass"
 CID_FILE_DIR=$(mktemp -d)
 TEST_SUMMARY=""
 TEST_SET=${TEST_SET_POS} ct_run_tests_from_testset "should_pass" >> /dev/null
-(ct_cleanup >> /dev/null)
-test $? -eq 0
+(ct_cleanup >> /dev/null 2>&1)
+if test $? -eq 0 ; then
+  echo "TC has passed"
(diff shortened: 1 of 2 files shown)


zmiklank:
[test-all]

zmiklank:
[test-all]

tests/test-lib/run_all_tests
```diff
@@ -20,17 +20,29 @@ TEST_SET_POS="foo bar"
...
+ret_val=0
 
-# positive test
+echo "running positive TC that should pass"
 CID_FILE_DIR=$(mktemp -d)
 TEST_SUMMARY=""
 TEST_SET=${TEST_SET_POS} ct_run_tests_from_testset "should_pass" >> /dev/null
-(ct_cleanup >> /dev/null)
-test $? -eq 0
+(ct_cleanup >> /dev/null 2>&1)
```
phracek, line 29: Why in brackets? If it fails, then we have a problem somewhere else. We do not need to run it in different shell.

phracek: This can be moved after condition `if`

zmiklank: The subshell is there because the correct beviour of `ct_cleanup` is to exit a shell at its end. That is not usable in this test as we need to collect the result afterwards.
Thus the subshell is used. If the shell in subshell exits with 1 also the return from subshell is 1.

> This can be moved after condition if
What is the reason for moving this? What benefits would it bring? The `ct_run_tests_from_testset` does not return value based on test results, only the TESTSUITE_RESULT is set. Do you mean we should check for TESTSUITE_RESULT here?

zmiklank: @phracek?

phracek (approved):
Thanks for explanation. LGTM.

PR #309, opened 2022-11-24
do not source the OC library if already sourced

zmiklank:


phracek:
@zmiklank Make sense to me to not include the source file twice. Can you please do it also for the other files?

@hhorak @pkubatrh WDYT bash experts? Does it make sense also from your point of view?

zmiklank:
Yes, if we proceed with this proposal I will add the check also to other libs.

zmiklank:
[test-all]

hhorak:
lgtm

zmiklank:
[test-all]

zmiklank:
Fixes: https://github.com/sclorg/postgresql-container/issues/478

PR #310, opened 2022-11-24
Make sanity check action results clickable

zmiklank:
And exit with error if tests fail

### DIFF
--- a/.github/workflows/sanity-tests.yaml
+++ b/.github/workflows/sanity-tests.yaml
@@ -36,49 +36,40 @@ jobs:
           sudo apt-get update && sudo apt-get -y install curl jq make git python3 python3-pip
           pip3 install pre-commit
 
-      - name: Create status check to pending
-        id: pending
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          cat << EOF > pending.json
-          {
-            "sha": "${{ steps.sha.outputs.SHA }}",
-            "state": "pending",
-            "context": "Sanity tests"
-          }
-          EOF
-          echo "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/${{ steps.sha.outputs.SHA}} "
-          curl -X POST -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3+json" \
-            https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/${{ steps.sha.outputs.SHA }} \
-            --data @pending.json
+      - name: Set status to pending
+        uses: myrotvorets/set-commit-status-action@1.1.5
+        with:
+          context: "Sanity check"
+          sha: ${{ steps.sha.outputs.sha }}
 
       - name: Run Sanity pre-commit tests
         continue-on-error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/sanity-tests.yaml
```diff
@@ -36,49 +36,40 @@ jobs:
...
-          }
-          EOF
-          echo "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/${{ steps.sha.outputs.SHA}} "
-          curl -X POST -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3+json" \
-            https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/${{ steps.sha.outputs.SHA }} \
-            --data @pending.json
+      - name: Set status to pending
+        uses: myrotvorets/set-commit-status-action@1.1.5
+        with:
+          context: "Sanity check"
```
phracek, line 42: `status` is missing.

phracek: I see default is `pending`

.github/workflows/sanity-tests.yaml
```diff
@@ -36,49 +36,40 @@ jobs:
...
       - name: Run Sanity pre-commit tests
         continue-on-error: true
+        shell: bash
         id: pre_commit_tests
         run: |
           set +e
           make pre-commit-check
-          echo "exit_status=$?" >> $GITHUB_OUTPUT
+          exit_status=$?
+          if [[ "$exit_status" == 0 ]]; then
```
phracek, line 53: This will not work. It should be either [ "$exit_status" == "0" ] or [ $exit_status -eq 0 ]

.github/workflows/sanity-tests.yaml
```diff
@@ -36,49 +36,40 @@ jobs:
...
+          sha: ${{ steps.sha.outputs.sha }}
 
-          if [[ ${{ steps.pre_commit_tests.outputs.exit_status }} == 0 ]]; then
-            FINAL_STATE="success"
-          else
-            FINAL_STATE="failure"
+      - name: Exit on ERR
+        shell: bash
+        run: |
+          if [[ "${{ steps.pre_commit_tests.outputs.result }}" == failure ]]; then
```
phracek, line 71:
```suggestion
          if [[ "${{ steps.pre_commit_tests.outputs.result }}" == "failure" ]]; then
```

.github/workflows/sanity-tests.yaml
```diff
@@ -36,49 +36,40 @@ jobs:
...
+            echo "Sanity checks failed."
+            echo "See the reason in 'Run Sanity pre-commit tests' step."
+            exit 1
```
phracek, line 74: Exit 1 is not needed. I guess we have a status `failure`

phracek (changes requested):

.github/workflows/sanity-tests.yaml
zmiklank: I though that it is a little misleading if the status is failure, but the action itself ends with success.
Or do you think, that is the correct behaviour?

phracek (approved):

PR #315, opened 2022-12-13
do not use --iidfile parameter in docker build

zmiklank:


zmiklank:
[test]

zmiklank:
[test-all]

zmiklank:
[test-all]

test-lib.sh
```diff
@@ -514,8 +514,10 @@ EOF
   local id_file
   mkdir -p "${APP_ID_FILE_DIR:?}"
   id_file="${APP_ID_FILE_DIR:?}"/"$RANDOM"
-  if ! docker build -f "$tmpdir/Dockerfile" --no-cache "$tmpdir" --iidfile "$id_file"; then
-    echo "  ERROR: Failed to find $binary in Dockerfile!" >&2
+  docker build -f "$tmpdir/Dockerfile" -q --no-cache "$tmpdir" > "$id_file"
+  #shellcheck disable=SC2181
+  if [ "$?" -ne 0 ]; then
```
phracek, line 519:
```suggestion
  if [ $? -ne 0 ]; then
```

test-lib.sh
```diff
@@ -487,7 +488,6 @@ function ct_npm_works() {
 
   if [ -f "$cid_file" ]; then
       docker stop "$(cat "$cid_file")"
-      rm "$cid_file"
```
phracek, line 490: Why it is not removed?

zmiklank:
[test-all]

test-lib.sh
```diff
@@ -514,8 +514,10 @@ EOF
   local id_file
   mkdir -p "${APP_ID_FILE_DIR:?}"
   id_file="${APP_ID_FILE_DIR:?}"/"$RANDOM"
-  if ! docker build -f "$tmpdir/Dockerfile" --no-cache "$tmpdir" --iidfile "$id_file"; then
-    echo "  ERROR: Failed to find $binary in Dockerfile!" >&2
+  docker build -f "$tmpdir/Dockerfile" -q --no-cache "$tmpdir" > "$id_file"
```
phracek, line 517: In case $id_file is not checked why the output is redirected to $id_file?
Any reason for this?

phracek (changes requested):
Thanks for this PR! Some issues, that I have hit.

test-lib.sh
zmiklank: This is done in final `ct_cleanup` function.

test-lib.sh
zmiklank: What do you mean by "$id_file is not checked"?

zmiklank:
I also tested this on rhel7 machine with docker installed and it worked.

test-lib.sh
zmiklank: > Do we really need -q? What about the result of why the image build failed?

That goes to stderr. 
We need -q, because if it is not present then whole output of the build goes to "$id_file." Id did not find any other option for separating the image id, as the --id-file parameter can not be used.

phracek: We do not see the output of `docker build` and we can not analyze troubles.

zmiklank: And what if we check the return value and when it does not equal 0 we run the build again with all log going to stdout?

phracek (changes requested):
Description is missing in the pull request.

test-lib.sh
zmiklank: That could be possible only if there is some equivalent parameter to `--iidfile` in the docker which is available in rhel7.
However there is no equivalent. So I do not know how the suggested approach could be implemented.

Another possibility could be to write some build function, which will build image with `--iidfile` if possible, and `-q` parameter otherwise.

phracek: > And what if we check the return value and when it does not equal 0 we run the build again with all log going to stdout?

Do not run the same command twice.

phracek: > That could be possible only if there is some equivalent parameter to `--iidfile` in the docker which is available in rhel7. However there is no equivalent. So I do not know how the suggested approach could be implemented.
> 
> Another possibility could be to write some build function, which will build image with `--iidfile` if possible, and `-q` parameter otherwise.

What about this approach:
```bash

  DOCKER_BUILD_OPTIONS="" # Option for docker command
  docker info 2>/dev/null | grep podman 1>/dev/null && is_podman=1 || is_podman=0
  if [ $is_podman -eq 1 ]; then
    DOCKER_BUILD_OPTIONS='--iidfile="$id_file"'
  fi

```

Based on the documentation https://docs.docker.com/engine/reference/commandline/build/
docker build has an option --iidfile.

Does it make sense to for ask to docker maintainer to add this option also to RHEL7?

phracek (changes requested):
Please see my comments.

I really do not want to build a container twice, because some layers already exist.

test-lib.sh
zmiklank: It would mean that cleaning app images for docker will not work. Are you OK with that?

zmiklank: @phracek?

zmiklank: @hhorak @pkubatrh Please could you look at this thread and state your opinion about this? I would like to have this fixed.

In my opinion the best approach would be to build quietly and the repeat the build command for logging purposes if the build fails.

zmiklank:
> I really do not want to build a container twice, because some layers already exist.

 That is even better for us. We would need only the logs from the failing layer anyway.

test-lib.sh
hhorak: From the solutions discussed here I like see the `-q` more, also would like to avoid splitting behavior if possible. However, isn't this the same issue as we have in `build.sh` already? We might use the same (maybe simplified) approach: https://github.com/sclorg/container-common-scripts/blob/master/build.sh#L180

zmiklank: Good idea, thanks.

zmiklank:
[test]

zmiklank:
[test]

test-lib.sh
zmiklank: @hhorak this is now reworked. Please take a look.

phracek:
s2i-base-container is failing here for CentOS 7:
```bash
Testing 'cgroup limits' usage...
  The test is running as root, all tests for cgroup limits will be run
  Testing 'limited memory' usage...
  File "/usr/bin/cgroup-limits", line 180, in <module>
    "NUMBER_OF_CORES": get_number_of_cores()
  File "/usr/bin/cgroup-limits", line 97, in get_number_of_cores
    limits = [l for l in [_read_cpu_quota(), _read_cpuset_size(), _read_nproc()] if l]
  File "/usr/bin/cgroup-limits", line 172, in _read_nproc
    except FileNotFoundError:
NameError: global name 'FileNotFoundError' is not defined
MEMORY_LIMIT_IN_BYTES=
test/run: line 25: [: : integer expression expected
MEMORY_LIMIT_IN_BYTES not set to 536870912.
```

It should be fixed by #322

zmiklank:
> And for RHEL8 there are failing for reason:
> 
> ```shell
> Testing 'cgroup limits' usage...
>   The test is running as root, all tests for cgroup limits will be run
>   Testing 'limited memory' usage...
> Traceback (most recent call last):
>   File "/usr/bin/cgroup-limits", line 180, in <module>
>     "NUMBER_OF_CORES": get_number_of_cores()
>   File "/usr/bin/cgroup-limits", line 97, in get_number_of_cores
>     limits = [l for l in [_read_cpu_quota(), _read_cpuset_size(), _read_nproc()] if l]
>   File "/usr/bin/cgroup-limits", line 171, in _read_nproc
>     return int(subprocess.run('nproc', capture_output=True).stdout)
>   File "/usr/lib64/python3.6/subprocess.py", line 423, in run
>     with Popen(*popenargs, **kwargs) as process:
> TypeError: __init__() got an unexpected keyword argument 'capture_output'
> MEMORY_LIMIT_IN_BYTES=
> test/run: line 25: [: : integer expression expected
> MEMORY_LIMIT_IN_BYTES not set to 536870912.
> ```

I think this is, however, not caused or related to this PR.

test-lib.sh
```diff
@@ -95,6 +95,33 @@ function ct_cleanup() {
...
+ct_build_image_and_parse_id() {
+  local tmpdir
+  local log_file
+  local ret_val
+  local dockerfile
+  local command
+  log_file="$(mktemp)"
+  ret_val=1
+  [ -n "$1" ] && dockerfile="-f $1"
+  command="$(echo "docker build --no-cache $dockerfile $2" | tr -d "'")"
```
phracek, line 115: What is the propose of `tr -d "'"`?

phracek (changes requested):

test-lib.sh
zmiklank: Because when inserting parameters from a variable to the command they were eclosed with '' and the command then did not work.

PR #316, opened 2022-12-19
cgroups: return None if number of cores can not be obtained

zmiklank:
Fixes: https://github.com/sclorg/postgresql-container/issues/482

### DIFF
--- a/shared-scripts/core/usr/bin/cgroup-limits
+++ b/shared-scripts/core/usr/bin/cgroup-limits
@@ -70,10 +70,10 @@ def get_number_of_cores():
     Read number of CPU cores.
     """
 
-    limits = [_read_cpu_quota(), _read_cpuset_size()]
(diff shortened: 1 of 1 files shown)


zmiklank:
[test-all]

zmiklank:
@pkubat could you please look at this?

Another approach could be to return at least something, if either _read_cpu_quota() or _read_cpuset_size() is successful. Wdyt?

pkubatrh:
Since the `return` part already expects `None` in the list items, we should instead check if there are only `None` values in the list.

zmiklank:
Oh, you are right.

zmiklank:
Fixed.

pkubatrh:
[test-all]

pkubatrh:
LGTM if tests pass

zmiklank:
This change will not appear in the test results, because they affect the s2i-core-container -> other containers would need be to be rebuilt on top of it first, in order to propagate the change.

PR #237, opened 2022-03-21
Calculate available cores from cpu cgroups controller

yrro:
Fixes: https://github.com/sclorg/s2i-base-container/issues/180

PR #317, opened 2022-12-21
Make cgroup-limits script work for non-root containers

hhorak:
Resource limits detection was enhanced via:
https://github.com/sclorg/container-common-scripts/pull/237 While this is not directly connected, mentioning to connect both issues that touch resource limits detection.

Per podman-run(1) man page, on some systems, changing the resource limits may not be allowed for non-root users. For more details, see https://github.com/containers/podman/blob/main/troubleshooting.md#26-running-containers-with-resource-limits-fails-with-a-permissions-error.

This caused the cgroup-limits script to fail with python traceback:

```
$ podman run -ti --rm quay.io/fedora/s2i-core:37 /usr/bin/cgroup-limits Warning: Can't detect cpu quota from cgroups
Warning: Can't detect cpuset size from cgroups
  File /usr/bin/cgroup-limits, line 143, in <module>
    NUMBER_OF_CORES: get_number_of_cores()
                       ^^^^^^^^^^^^^^^^^^^^^
  File /usr/bin/cgroup-limits, line 76, in get_number_of_cores
    return min([l for l in limits if l])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: min() arg is an empty sequence
```

This change implements another way to see number of CPUs (without cgroups), using nproc tool.

This might work as the last resort when running a container as non-root, so we get the same output as when the same container is run as a root on the same machine.

Also tracked internally: https://issues.redhat.com/browse/RHELPLAN-141599

### DIFF
--- a/shared-scripts/core/usr/bin/cgroup-limits
+++ b/shared-scripts/core/usr/bin/cgroup-limits
@@ -146,7 +146,7 @@ def _read_cpuset_size():
     if line is None:
         # None of the files above exists when running podman as non-root,
         # so in that case, this warning is printed every-time
-        print("Warning: Can't detect cpuset size from cgroups",
+        print("Warning: Can't detect cpuset size from cgroups, will use nproc",
               file=sys.stderr)
         return None
 
@@ -168,8 +168,11 @@ def _read_nproc():
     configuration (per podman-run(1) man page).
     """
     try:
-        return int(subprocess.run('nproc', capture_output=True).stdout)
-    except FileNotFoundError:
+        stdout, stderr = subprocess.Popen('nproc', stdout=subprocess.PIPE).communicate()
+        return int(stdout)
+    except EnvironmentError as e:
+        if e.errno != errno.ENOENT:
+            raise
         return None
 
 
--- a/shared-scripts/core/usr/bin/cgroup-limits
+++ b/shared-scripts/core/usr/bin/cgroup-limits
@@ -33,11 +33,35 @@ Variables currently supported:
     NO_MEMORY_LIMIT
         Set to "true" if MEMORY_LIMIT_IN_BYTES is so high that the caller
         can act as if no memory limit was set. Undefined otherwise.
+
+Note about non-root containers:
+
+    Per podman-run(1) man page, on some systems, changing the resource limits
+    may not be allowed for non-root users. For more details, see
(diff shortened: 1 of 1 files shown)


hhorak:
[test]

phracek:
[test-all]

phracek (changes requested):
The pull request itself LGTM, but I have hit some issues on RHEL7 machine:

Python 2.7 traceback:
```bash
$ ./shared-scripts/core/usr/bin/cgroup-limits
  File "./shared-scripts/core/usr/bin/cgroup-limits", line 180, in <module>
    "NUMBER_OF_CORES": get_number_of_cores()
  File "./shared-scripts/core/usr/bin/cgroup-limits", line 97, in get_number_of_cores
    limits = [l for l in [_read_cpu_quota(), _read_cpuset_size(), _read_nproc()] if l]
  File "./shared-scripts/core/usr/bin/cgroup-limits", line 172, in _read_nproc
    except FileNotFoundError:
NameError: global name 'FileNotFoundError' is not defined
```

on RHEL7 by calling python-3.6:
```bash
$ python3.6 ./shared-scripts/core/usr/bin/cgroup-limits
  File "./shared-scripts/core/usr/bin/cgroup-limits", line 180, in <module>
    "NUMBER_OF_CORES": get_number_of_cores()
  File "./shared-scripts/core/usr/bin/cgroup-limits", line 97, in get_number_of_cores
    limits = [l for l in [_read_cpu_quota(), _read_cpuset_size(), _read_nproc()] if l]
  File "./shared-scripts/core/usr/bin/cgroup-limits", line 171, in _read_nproc
    return int(subprocess.run('nproc', capture_output=True).stdout)
  File "/usr/lib64/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'capture_output'
```

phracek (commented):
Python-3.6 does not have a parameter called `capture_output`: https://docs.python.org/3.6/library/subprocess.html

hhorak:
Thanks, I've re-implemented the `nproc` call to something that should work on both, Python 2 and 3.

hhorak:
[test-all]

hhorak:
[test-all]

phracek:
[test]

hhorak:
I don't understand why, but some commits from this PR are already merged in master. The remaining fix that fixes #319 is is now separately in #322, so let's close this PR that is not needed in its most part.

PR #318, opened 2023-01-05
Switch to project default at the beginning of each test.

phracek:
Switch to project default at the beginning of each test.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -1310,6 +1310,11 @@ ct_run_tests_from_testset() {
         TESTSUITE_RESULT=1
       fi
     fi
+    # As soon as test is finished
+    # switch the project from sclorg-test-<NUMBER> to default.
+    if [ "${CT_OCP4_TEST:-false}" == "true" ]; then
+      oc project default
+    fi
     time_diff=$(ct_timestamp_diff "$time_beg" "$time_end")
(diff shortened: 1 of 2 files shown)


phracek:
[test]

phracek:
[test]

phracek:
Rebased on the top of master

[test]

zmiklank:
I see that this change is for the openshift tests, so I will run also the openshift testsuite.
[test-openshift]

test-lib.sh
```diff
@@ -1310,6 +1310,11 @@ ct_run_tests_from_testset() {
         TESTSUITE_RESULT=1
       fi
     fi
+    # As soon as test is finished
+    # switch the project from sclorg-test-<NUMBER> to default.
+    if [ "${CT_OCP4_TEST:-false}" == "true" ]; then
```
zmiklank, line 1315: Handling OC stuff does not really seem to fit to this function.  It is not related to executing tests.
However, I understand, that adding this line to all of the containers is not really great solution too.

What about adding these lines to the openshift cleanup function `ct_os_cleanup`?

zmiklank (changes requested):

test-lib.sh
phracek: This function has to be called after each test and not after the whole test suite.

zmiklank: Oh, I see now, thanks for explanation. I am OK with this solution then.

zmiklank (approved):

Issue #326, opened 2023-02-09
regression in 'ct_clean_app_images'


phracek:
See TF logs: https://artifacts.dev.testing-farm.io/1aeb5586-979f-482e-92cd-b4e6ae053b91/

Tests itself PASSED:

It looks like we have a big regression here:

```bash
+ echo ===============================================
===============================================
+ echo 'Test cases results:'
Test cases results:
+ echo

+ echo ' [PASSED] for '\''postgresql-container_tests'\'' run_container_creation_tests (00:00:33)
 [PASSED] for '\''postgresql-container_tests'\'' run_general_tests (00:01:01)
 [PASSED] for '\''postgresql-container_tests'\'' run_change_password_test (00:00:06)
 [PASSED] for '\''postgresql-container_tests'\'' run_replication_test (00:00:08)
 [PASSED] for '\''postgresql-container_tests'\'' run_master_restart_test (00:00:12)
 [PASSED] for '\''postgresql-container_tests'\'' run_doc_test (00:00:01)
 [PASSED] for '\''postgresql-container_tests'\'' run_s2i_test (00:00:17)
 [PASSED] for '\''postgresql-container_tests'\'' run_test_cfg_hook (00:00:10)
 [PASSED] for '\''postgresql-container_tests'\'' run_s2i_bake_data_test (00:00:18)
 [PASSED] for '\''postgresql-container_tests'\'' run_s2i_enable_ssl_test (00:00:10)
 [PASSED] for '\''postgresql-container_tests'\'' run_upgrade_test (00:00:00)
 [PASSED] for '\''postgresql-container_tests'\'' run_migration_test (00:00:00)
 [PASSED] for '\''postgresql-container_tests'\'' run_pgaudit_test (00:00:03)
 [PASSED] for 'postgresql-container_tests' run_container_creation_tests (00:00:33)
 [PASSED] for 'postgresql-container_tests' run_general_tests (00:01:01)
 [PASSED] for 'postgresql-container_tests' run_change_password_test (00:00:06)
 [PASSED] for 'postgresql-container_tests' run_replication_test (00:00:08)
 [PASSED] for 'postgresql-container_tests' run_master_restart_test (00:00:12)
 [PASSED] for 'postgresql-container_tests' run_doc_test (00:00:01)
 [PASSED] for 'postgresql-container_tests' run_s2i_test (00:00:17)
 [PASSED] for 'postgresql-container_tests' run_test_cfg_hook (00:00:10)
[...]

PR #329, opened 2023-02-21
ignore rm of cidfile if it not exist

zmiklank:
apparently some versions of podman delete also the cidfile during podman rm and some not. That leads to situation that in some OSes (e.g., c9s) the cleaning function fails on podman rm, as the cid_file is removed at that point. This commit fixes this behaviour.

Fixes: https://github.com/sclorg/container-common-scripts/issues/326


zmiklank:
[test]

phracek (approved):
LGTM

PR #339, opened 2023-03-21
Add .build-args-OS file to let specify build options

hhorak:
Podman used to set some capabilities by default, but does not any more with version 4.4.0 and newer. These capabilities need to be set for being able to build micro images that install RPMs into an alternative path in the Dockerfile. Introducing .build-args-OS file means we will be able to specify some arguments to docker build command for each Dockerfile specifically, for example specify capabilities for the micro images.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -177,7 +177,9 @@ function docker_build_with_version {
   # let them be included in the build command.
   # The content of the .build-opts-<OS> will be used as part of the docker build command.
   if [ -f "$build_args_file" ] ; then
-      BUILD_OPTIONS+=" $(cat "$build_args_file")"
+      build_opts_from_file="$(cat "$build_args_file")"
+      echo '⚠️ Be aware, using additional build arguments: ' "${build_opts_from_file}"
+      BUILD_OPTIONS+=" ${build_opts_from_file}"
   fi
 
   pull_image "$dockerfile"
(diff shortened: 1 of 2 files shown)


frenzymadness (changes requested):
I've found just one typo.

frenzymadness:
If you don't mind me asking an offtopic question: why do micro images need to install RPMs to non-standard locations?

build.sh
```diff
@@ -170,6 +171,15 @@ function docker_build_with_version {
     fi
   fi
 
+  # If a specific Dockerfile requires some specific build options (like capabilities),
+  # let them be included in the build command.
+  # The content of the .build-opts-<OS> will be used as part of the docker build command.
+  if [ -f "$build_args_file" ] ; then
+      build_opts_from_file="$(cat "$build_args_file")"
```
phracek, line 178: This is a bit security issue. Who will check, that content is right? What about if there is security or vulnerability content and tests are going to be run? I am trying to be a bit paranoic. :)

phracek (changes requested):
What do you think about my comment? Does it make sense?

build.sh
frenzymadness: How is this file different from others? I mean, if you can put any malicious code to this file, you can put it also anywhere else in the codebase.

zmiklank (approved):
LGTM.

phracek:
@hhorak Please rebase it and we can merge it.

phracek (approved):

phracek:
phracek rebased.

[test]

Issue #286, opened 2022-08-30
Add trap on ERR if exit on error bash option is enabled


zmiklank:
This would simplify the debug mode.

From man trap:
```
  -e      Exit  immediately  if a pipeline (which may consist of a single simple
                      command), a list, or a compound command (see SHELL GRAMMAR above), ex‐
                      its  with  a  non-zero status. 
                      .
                      .
                      .
                       A  trap  on  ERR,  if set, is executed before the shell exits.
                      This option applies to the shell environment and each  subshell  envi‐
                      ronment  separately (see COMMAND EXECUTION ENVIRONMENT above), and may
                      cause subshells to exit before executing all the commands in the  sub‐
                      shell.
```
We would need to add also -E bash option:
```
    -E      If set, any trap on ERR is inherited by shell functions, command  sub‐
             stitutions,  and commands executed in a subshell environment.  The ERR
             trap is normally not inherited in such cases.
```

PR #341, opened 2023-03-30
add debugging trap on error to scripts with exit on err enabled

zmiklank:
Resolves: https://github.com/sclorg/container-common-scripts/issues/286

### DIFF
--- a/betka.sh
+++ b/betka.sh
@@ -5,7 +5,7 @@
 # The result directory will contain generated sources from upstream directories
 #
 
-set -eEx
+set -eE
 
 trap 'echo "errexit on line $LINENO, $0" >&2' ERR
 
--- a/betka.sh
+++ b/betka.sh
@@ -5,7 +5,9 @@
 # The result directory will contain generated sources from upstream directories
 #
 
-set -ex
+set -eEx
+
+trap 'echo "errexit on line $LINENO, $0" >&2' ERR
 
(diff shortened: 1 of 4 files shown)


zmiklank:
[test]

zmiklank:
test errors are not related to this PR. But at least we can see there, that trap on ERR works ;).

betka.sh
```diff
@@ -5,7 +5,9 @@
 # The result directory will contain generated sources from upstream directories
 #
 
-set -ex
+set -eEx
```
phracek, line 8: -x is not needed.

zmiklank: I can remove it. But the -x was present before this PR too.

zmiklank: It is removed in [12dc547](https://github.com/sclorg/container-common-scripts/pull/341/commits/12dc547d1bf8623b5835e16d323bd3ac36ca0724).

phracek (approved):
Everything LGTM. Thanks. Only nitpick. I would add there a note or comment what -E means. So the user should not Google it ;)

zmiklank:
> Everything LGTM. Thanks. Only nitpick. I would add there a note or comment what -E means. So the user should not Google it ;)

I know, but this info can be easily gathered from `man sh`. I am not sure if it is efficient to have comments for such things here. But that's subjective opinion, of course.

PR #342, opened 2023-05-15
Show resources only in non-debug mode

hhorak:
It may slow down debugging the tests, so let's not show resources whatsoever when in DEBUG mode


hhorak:
[test]

zmiklank:
Good idea. It is unfortunate that we still do not have a straightforward way do enable debug mode in CI :/.

PR #345, opened 2023-06-21
Don't print the error message on multispec exit code 2

befeleme:
For non-existing and not expected matrix combinations generator produced a loud message that looked like a genuine error.

```
[ERROR] distgen failed: Command '['dg', '--multispec', 'specs/multispec.yml',
'--template', 'src/Dockerfile.template', '--distro', 'centos-7-x86_64.yaml',
'--multispec-selector', 'version=3.11', '--output', '3.11/Dockerfile.centos7']'
returned non-zero exit status 2.
```

For exit code 2 the message is now supressed.
Related: https://github.com/sclorg/s2i-python-container/issues/594

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -37,7 +37,10 @@ def run_distgen(
     try:
         check_output(cmd)
     except CalledProcessError as e:
-        print("[ERROR] distgen failed:", e)
+        # Exit code 2 is a special code for non existing matrix combinations
+        # It's not an actual error, hence don't print out its error message
+        if e.returncode != 2:
+            print("[ERROR] distgen failed:", e)
 
 
 def get_version_distro_mapping(


zmiklank:
Thank you for the PR.

With this patch the log output on missing combination is as follows:
```
DGM	src/Dockerfile-minimal.template → 3.9-minimal/Dockerfile.centos7
dg: CRITICAL: This combination is not included in matrix section
```
which is I think informative enough - we know that something is missing and we know what exactly it is - so we can still take some action if the combination should not be missing.

zmiklank:
[test]

zmiklank (approved):

frenzymadness (approved):
I like it!

PR #349, opened 2023-08-21
Do not delete pvc

phracek:
Do not delete `pvc` in OpenShift 4 environment. It deletes also `pvc` in image-registry-operator.

Also switch back variable back to CT_SKIP_NEW_PROJECT=false.


phracek:
[test]

phracek:
[test]

PR #350, opened 2023-08-23
Build and tag the non-minimal variant as well

hhorak:
Without having the non-minimal image build and tagged, the image might be missing (sometimes it's not available in the repository yet, for new versions specifically), so let's build and tag it as well.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -207,6 +207,14 @@ if [ -n "${SINGLE_VERSION:-}" ]; then
     exit 0
   fi
   dirs=$SINGLE_VERSION
+  if [[ "${SINGLE_VERSION}" == *"minimal"* ]]; then
+    echo "Adding ${SINGLE_VERSION//-minimal/} because it might be needed for testing $SINGLE_VERSION."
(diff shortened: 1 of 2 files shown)


hhorak:
For example needed for https://github.com/sclorg/s2i-nodejs-container/pull/400

phracek (changes requested):
Good catch. Let's try it in CI, where it works.

What about 'micro' images? LIke in https://github.com/httpd-container/2.4-micro?

Before merging this pull request we need to analyse logs at least for s2i-nodejs-container which contains 'minimal' images and 'httpd-container' logs

phracek:
[test]

hhorak:
[test]

hhorak:
> Good catch. Let's try it in CI, where it works.
> 
> What about 'micro' images? LIke in https://github.com/httpd-container/2.4-micro?
> 
> Before merging this pull request we need to analyse logs at least for s2i-nodejs-container which contains 'minimal' images and 'httpd-container' logs

I think micro might be included as well, good point.

hhorak:
[test]

phracek (approved):

PR #351, opened 2023-09-05
allow building for any specified $OS

zmiklank:
Partly fixes https://github.com/sclorg/s2i-base-container/issues/279.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -212,22 +212,10 @@ echo "Built versions are: $dirs"
 
 for dir in ${dirs}; do
   pushd "${dir}" > /dev/null
-  if [ "$OS" == "rhel8" ] || [ "$OS" == "rhel8-candidate" ]; then
-    docker_build_with_version Dockerfile.rhel8
-  elif [ "$OS" == "rhel9" ] || [ "$OS" == "rhel9-candidate" ]; then
-    docker_build_with_version Dockerfile.rhel9
-  elif [ "$OS" == "rhel7" ] || [ "$OS" == "rhel7-candidate" ]; then
-    docker_build_with_version Dockerfile.rhel7
-  elif [ "$OS" == "fedora" ] || [ "$OS" == "fedora-candidate" ]; then
-    docker_build_with_version Dockerfile.fedora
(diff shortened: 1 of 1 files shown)


zmiklank:
[test]

zmiklank:
@phracek Do we still use the *-candidate $OSes? If yes I will rework this PR to work with them. Current proposal supports only standard targets (rhel7, rhel8, fedora, ...) for which the Dockerfiles exist.

build.sh
```diff
@@ -212,22 +212,10 @@ echo "Built versions are: $dirs"
...
-  elif [ "$OS" == "fedora" ] || [ "$OS" == "fedora-candidate" ]; then
-    docker_build_with_version Dockerfile.fedora
-  elif [ "$OS" == "centos6" ] || [ "$OS" == "centos6-candidate" ]; then
-    docker_build_with_version Dockerfile.centos6
-  elif [ "$OS" == "c8s" ] || [ "$OS" == "c8s-candidate" ]; then
-    docker_build_with_version Dockerfile.c8s
-  elif [ "$OS" == "c9s" ] || [ "$OS" == "c9s-candidate" ]; then
-    docker_build_with_version Dockerfile.c9s
-  else
+  if [ "$OS" == "centos7" ] || [ "$OS" == "centos7-candidate" ]; then
```
phracek, line 215: I don't know why `-candidate` were used?. Even `centos7` can be used like the rest.
I guess only one row is needed `docker_build_with_version Dockerfile."$OS"`

I guess we alreade have a Dockerfile as symlink to Dockerfile.centos7, right?

phracek (changes requested):
Thanks for this improvement.

See me comment.

build.sh
zmiklank: > I guess we alreade have a Dockerfile as symlink to Dockerfile.centos7, right?

AFAIK this is valid only for Python container. So we need to support centos7 separately.

zmiklank: @pkubatrh Do you know, if *-candidate targets are still used somewhere, please?

phracek (approved):
LGTM. Let's get merge it.

Issue #353, opened 2023-11-14
grep: warning: stray \ before -


frenzymadness:
It seems to me that the warning (grep: warning: stray \ before -) we are getting in the CI of Python container comes from this line:

https://github.com/sclorg/container-common-scripts/blob/250ae7c6b9312e2389b08d349a1a95c59ed1dde9/test-lib.sh#L928

It seems to be a new feature of grep 3.8. I'm not an expert here but it seems that a single slash at the beginning of the string is enough

zmiklank:
++ I came to this warning too, came to conclusion that it makes no harm and decided to let it be as it is.
However...now it seems that I am not the only person who spent a time on this which means it would be benefitial to either fix the warning or silence it.

Your suggestion seems correct, however we should probably check how this behave in older bashes too. Do you want to create PR for this or shall I?

frenzymadness:
I'm on PTO now so feel free to take it over.

zmiklank:
tested ` echo "$s2i_args" | grep -q "\--incremental" && incremental=true ` on CentOS7, and it works. PR will follow shortly.

PR #354, opened 2023-11-15
fix stray backslash in grep argument

zmiklank:
fixes grep warning:  (grep: warning: stray \ before -)

fixes: https://github.com/sclorg/container-common-scripts/issues/353
-->

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -925,7 +925,7 @@ ct_s2i_build_as_df()
         return 1
     fi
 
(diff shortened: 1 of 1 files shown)


zmiklank:
[test]

frenzymadness (approved):
I think it's good and I believe you've tested it at all supported systems.

zmiklank:
I tested it on Centos7 and it worked. I am not sure why the tests failed, but we no longer support testing of Centos7 images upstream so I've just created different PR which disables the tests.

PR #355, opened 2023-11-15
test c9s compose instead of deprecated centos7

zmiklank:
We no longer test centos7 compose in any of our images, thus it does not make sense here anymore.

@phracek could you please take a look at this PR, to check if I did not miss anything? Thanks.


phracek (approved):
LGTM

PR #368, opened 2024-03-27
Add build_args also to `ct_s2i_build_as_df` function.

phracek:
Add build_args also to `ct_s2i_build_as_df` function.
We have to functions to build image. This one and `ct_test_app_dockerfile`

In case of --ulimit is defined in s2i_args variable we will add this parameter to docker/podman command line.

This pull request blocks https://github.com/sclorg/s2i-nodejs-container/pull/425.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -813,7 +813,7 @@ ct_get_public_image_name() {
   elif [ "$os" == "rhel9" ]; then
     public_image_name=$registry/rhel9/$base_image_name-${version//./}
   elif [ "$os" == "centos7" ]; then
-    public_image_name=$registry/centos7/$base_image_name-${version//./}-centos7
+    public_image_name=$registry/centos7/$base_image_name-${version//./}-centos7:centos7
   elif [ "$os" == "c8s" ]; then
     public_image_name=$registry/sclorg/$base_image_name-${version//./}-c8s
   elif [ "$os" == "c9s" ]; then
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -989,6 +989,8 @@ EOF
     # Check if -v parameter is present in s2i_args and add it into docker build command
     read -ra mount_options <<< "$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*' || true)"
 
+    # Check if s2i_args contains --ulimit
+    # if so, add it into docker build command. The format is like --ulimit nofile=4096:4096
     echo "$s2i_args" | grep -q '\--ulimit' && build_args=$(echo "$s2i_args" | grep -o -e '\--ulimit[[:space:]]\S*\w=*')
     # Run the build and tag the result
     ct_build_image_and_parse_id "$df_name" "${mount_options[*]+${mount_options[*]}} -t $dst_image . $build_args"
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -883,6 +883,9 @@ ct_s2i_usage()
 # ct_s2i_build_as_df APP_PATH SRC_IMAGE DST_IMAGE [S2I_ARGS]
 # ----------------------------
 # Create a new s2i app image from local sources in a similar way as source-to-image would have used.
+# This function is wrapper for ct_s2i_build_as_df_build_args in case user do not want to add build args
+# This function is used in all https://github.com/sclorg/*-container test cases and we do not
+# want to break functionality
 # Argument: APP_PATH - local path to the app sources to be used in the test
 # Argument: SRC_IMAGE - image to be used as a base for the s2i build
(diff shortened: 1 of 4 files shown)


phracek:
[test]

test-lib.sh
```diff
@@ -988,8 +989,9 @@ EOF
     # Check if -v parameter is present in s2i_args and add it into docker build command
     read -ra mount_options <<< "$(echo "$s2i_args" | grep -o -e '\(-v\)[[:space:]]\.*\S*' || true)"
 
+    echo "$s2i_args" | grep -q '\--ulimit' && build_args=$(echo "$s2i_args" | grep -o -e '\--ulimit[[:space:]]\S*\w=*')
```
zmiklank, line 992: What is purpose of this line? Can you make a comment about it in the code please?

phracek: @zmiklank It gets s2i-args and parse if `--ulimit` is present. If so then get all `--ulimit parameters` a pass them into docker/podman build command.
But the comment is correct.

phracek: @zmiklank The comment is added into code and also into commit message https://github.com/sclorg/container-common-scripts/pull/368/commits/cb248e0eb51c785f812d432820d64f178c3eedad

phracek:
[test]

phracek:
[test]

test-lib.sh
phracek: Yes, but which one? There are a lot of them. For case of `s2i-nodejs-container` there is ```--ulimit nofile=4096:4096```. Passing build parameters to function `ct_s2i_build_as_df` is done at the end.

In this case, we should define a new variable and specify it, like --build-args, and the rest of the parameters will be used by podman. But this is used only by one specific container. Does it make sense either?

zmiklank: What if we pass build args to the function like this?
```
diff --git a/test-lib.sh b/test-lib.sh
index c38647e..c0be7ab 100644
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -880,6 +880,17 @@ ct_s2i_usage()
     docker run --rm "$img_name" bash -c "$usage_command"
 }
 
+ct_s2i_build_as_df()
+{
+    local app_path=$1; shift
+    local src_image=$1; shift
+    local dst_image=$1; shift
+    local s2i_args="$*";
+
+    ct_s2i_build_as_df_build_args $app_path $src_image $dst_image "" $s2i_args
+
+}
+
 # ct_s2i_build_as_df APP_PATH SRC_IMAGE DST_IMAGE [S2I_ARGS]
 # ----------------------------
 # Create a new s2i app image from local sources in a similar way as source-to-image would have used.
@@ -888,11 +899,12 @@ ct_s2i_usage()
 # Argument: DST_IMAGE - image name to be used during the tagging of the s2i build result
 # Argument: S2I_ARGS - Additional list of source-to-image arguments.
 #                      Only used to check for pull-policy=never and environment variable definitions.
-ct_s2i_build_as_df()
+ct_s2i_build_as_df_build_args()
 {
     local app_path=$1; shift
     local src_image=$1; shift
     local dst_image=$1; shift
+    local build_args=$1; shift
     local s2i_args="$*";
     local local_app=upload/src/
     local local_scripts=upload/scripts/
```

phracek: There is a typo. I guess the argument "" is empty and it will not work. I have to clarify it by real tests. Give me mmnt.

phracek:
I have tested this updated PR with s2i-nodejs-container repo and it works

[test]

test-lib.sh
```diff
@@ -894,6 +894,26 @@ ct_s2i_build_as_df()
     local src_image=$1; shift
     local dst_image=$1; shift
     local s2i_args="$*";
+
+    ct_s2i_build_as_df_build_args "$app_path" "$src_image" "$dst_image" "" "$s2i_args"
+}
+
+# ct_s2i_build_as_df_build_args APP_PATH SRC_IMAGE DST_IMAGE [S2I_ARGS]
```
zmiklank, line 901:
```
-  # ct_s2i_build_as_df_build_args APP_PATH SRC_IMAGE DST_IMAGE [S2I_ARGS]
+  # ct_s2i_build_as_df_build_args APP_PATH SRC_IMAGE DST_IMAGE BUILD_ARGS [S2I_ARGS]
```

Just a nitpick in comment.

Also could you please explain in comment of `ct_s2i_build_as_df` function why we created the wrapper?

phracek: It should be explained by https://github.com/sclorg/container-common-scripts/pull/368/commits/ca4f716c7b500c2ad9796a7c79b94f42e9848474

phracek:
[test]

zmiklank (approved):

PR #369, opened 2024-05-20
Remove set -e and add to the end FAILED message

phracek:
Remove set -e and add to the end FAILED message in case test failed.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -6,8 +6,6 @@
 # TEST_OPENSHIFT_MODE - If set, run OpenShift tests (if present)
 # VERSIONS - Must be set to a list with possible versions (subdirectories)
 
-trap 'echo "errexit on line $LINENO, $0" >&2' ERR
-
 [ -n "${DEBUG:-}" ] && set -x
 
 FAILED_VERSIONS=""
--- a/test.sh
+++ b/test.sh
@@ -82,7 +82,7 @@ for dir in ${VERSIONS}; do
   popd > /dev/null
 done
 
-if [[ "$FAILED_VERSIONS" != "" ]]; then
+if [[ -n "$FAILED_VERSIONS" ]]; then
     echo "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
     echo "Test for image ${IMAGE_NAME} FAILED in these versions ${FAILED_VERSIONS}."
     echo "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
--- a/test.sh
+++ b/test.sh
@@ -6,8 +6,6 @@
 # TEST_OPENSHIFT_MODE - If set, run OpenShift tests (if present)
 # VERSIONS - Must be set to a list with possible versions (subdirectories)
 
-set -E
-
 trap 'echo "errexit on line $LINENO, $0" >&2' ERR
 
 [ -n "${DEBUG:-}" ] && set -x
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -6,12 +6,28 @@
 # TEST_OPENSHIFT_MODE - If set, run OpenShift tests (if present)
 # VERSIONS - Must be set to a list with possible versions (subdirectories)
 
-set -eE
+set -E
```
zmiklank, line 9: `-E` is not needed if `-e` is not set IIUC.

test.sh
```diff
@@ -56,9 +74,16 @@ for dir in ${VERSIONS}; do
...
       echo "-> Upstream tests are not present, skipping"
     fi
   fi
   popd > /dev/null
 done
+
+if [[ "$FAILED_VERSIONS" != "" ]]; then
+    echo "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
+    echo "Test for image ${IMAGE_NAME} FAILED in these versions ${FAILED_VERSIONS}."
+    echo "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
```
zmiklank, line 88: We should exit the script with non-zero value, if some test fails.

test.sh
```diff
@@ -56,9 +74,16 @@ for dir in ${VERSIONS}; do
...
       VERSION=$dir test/run-upstream
+      failed_version "$?" "$dir"
     else
       echo "-> Upstream tests are not present, skipping"
     fi
   fi
   popd > /dev/null
 done
+
+if [[ "$FAILED_VERSIONS" != "" ]]; then
```
zmiklank, line 85:
```suggestion
if [[ -n "$FAILED_VERSIONS" ]]; then
```

zmiklank (changes requested):
Thanks for the PR, looks good overall, but I've added few comments.

zmiklank:
And please, check the shellcheck output.

test.sh
phracek: https://github.com/sclorg/container-common-scripts/pull/369/commits/fee1f6a17fc7867f6d7e9fb3d3e36c2cf1340edc

test.sh
phracek: https://github.com/sclorg/container-common-scripts/pull/369/commits/fee1f6a17fc7867f6d7e9fb3d3e36c2cf1340edc

phracek:
[test]

test.sh
```diff
@@ -6,12 +6,26 @@
 # TEST_OPENSHIFT_MODE - If set, run OpenShift tests (if present)
 # VERSIONS - Must be set to a list with possible versions (subdirectories)
 
-set -eE
-
 trap 'echo "errexit on line $LINENO, $0" >&2' ERR
```
zmiklank, line 9: Also this line no longer has any effect.

zmiklank:
However, I think this PR does not really solve the problem you want to solve. That would require to remove exit on error from build.sh and tag.sh. However, that is not safe to do, as all of testsuites on our container repos rely on that.
And if any image is not even buildable - as it the case in this PR - I see no reason to try to run rest of the tests anyway, even for other versions, as unbuildable image is a big problem by itself.

phracek:
> However, I think this PR does not really solve the problem you want to solve. That would require to remove exit on error from build.sh and tag.sh. However, that is not safe to do, as all of testsuites on our container repos rely on that. And if any image is not even buildable - as it the case in this PR - I see no reason to try to run rest of the tests anyway, even for other versions, as unbuildable image is a big problem by itself.

Currently, in case of `build.sh` and `tag.gs` failed then test is not run anymore. Because in both script there is present `set -e` where it should be.
In case one test has failed, then the next version has to be tested.
This pull request does not belong to building and tagging. Only for tests.

test.sh
phracek: https://github.com/sclorg/container-common-scripts/pull/369/commits/2229acbe720e89bae02bb70256af4efca888e5d7

zmiklank (approved):

PR #373, opened 2024-05-28
Add support PyTest test suite

phracek:
This pull request adds support for testing sclorg container images by container-ci-suite, and especially by PyTest.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -110,8 +110,8 @@ test-openshift: script_env += TEST_OPENSHIFT_MODE=true
 test-openshift: tag
 	VERSIONS="$(VERSIONS)" BASE_IMAGE_NAME="$(BASE_IMAGE_NAME)" $(script_env) $(test)
 
-.PHONY: test-pytest
-test-pytest: script_env += TEST_PYTEST=true
+.PHONY: test-openshift-pytest
+test-pytest: script_env += TEST_OPENSHIFT_PYTEST=true
 test-pytest: tag
 	VERSIONS="$(VERSIONS)" BASE_IMAGE_NAME="$(BASE_IMAGE_NAME)" $(script_env) $(test)
 
--- a/common.mk
+++ b/common.mk
@@ -110,6 +110,12 @@ test-openshift: script_env += TEST_OPENSHIFT_MODE=true
 test-openshift: tag
 	VERSIONS="$(VERSIONS)" BASE_IMAGE_NAME="$(BASE_IMAGE_NAME)" $(script_env) $(test)
 
+.PHONY: test-pytest
+test-pytest: script_env += TEST_PYTEST=true
+test-pytest: tag
+	VERSIONS="$(VERSIONS)" BASE_IMAGE_NAME="$(BASE_IMAGE_NAME)" $(script_env) $(test)
+
+
 .PHONY: shellcheck
 shellcheck:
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -75,6 +75,12 @@ for dir in ${VERSIONS}; do
       echo "-> Upstream tests are not present, skipping"
     fi
   fi
+
+  if [ -n "${TEST_PYTEST}" ]; then
+    VERSION=$dir python3 -m pytest -vv --showlocals test/test_*.py
```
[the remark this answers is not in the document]
zmiklank: Well, yes, I would say more optimal solution would be to create a separate file for this. However, I would consider different naming, because this is only about OCP4 tests, isn't it?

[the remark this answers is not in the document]
phracek: > solution



> Well, yes, I would say more optimal solution would be to create a separate file for this. However, I would consider different naming, because this is only about OCP4 tests, isn't it?

For now yes, but maybe later on, we will migrate also container test ;)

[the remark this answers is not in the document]
phracek: Fixed by https://github.com/sclorg/container-common-scripts/pull/373/commits/a14b9f2f919e2b86e7f678e9425d4eb682fdc8c6. Good point.

common.mk
```diff
@@ -114,6 +114,12 @@ test-openshift: script_env += TEST_OPENSHIFT_MODE=true
 test-openshift: tag
 	VERSIONS="$(VERSIONS)" BASE_IMAGE_NAME="$(BASE_IMAGE_NAME)" $(script_env) $(test)
 
+.PHONY: test-openshift-pytest
+test-pytest: script_env += TEST_OPENSHIFT_PYTEST=true
```
zmiklank, line 118: The makefile target is still called test-pytest, not test-openshift-pytest.
Also is there any reason for duplicating it?

phracek: No. Stupid typo from me. Fixed by https://github.com/sclorg/container-common-scripts/pull/373/commits/d8cc7304106a0abe6bfdd65611cd9f601e0e90a4

phracek:
[test]

phracek:
Rebased against master

[test]

test.sh
```diff
@@ -75,6 +75,16 @@ for dir in ${VERSIONS}; do
       echo "-> Upstream tests are not present, skipping"
     fi
   fi
+
+  if [ -n "${TEST_OPENSHIFT_PYTEST}" ]; then
+    if [[ -x test/run-openshift-pytest ]]; then
```
pkubatrh, line 80: We shouldnt need double brackets over here

pkubatrh:
otherwise lgtm

test.sh
phracek: The double brackets are used in the whole script it makes sense to remove it in the entire script. We do not use any enhanced functionality.

phracek: https://github.com/sclorg/container-common-scripts/pull/373/commits/9fdd56c9924ec202f8e8a57c8fabcb984cde5599

phracek:
[test]

test.sh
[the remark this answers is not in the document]
zmiklank: What about the separate file for test command?

zmiklank (approved):

PR #378, opened 2024-07-01
Add support for building, testing C10S in container-common-scripts

phracek:

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -53,6 +53,10 @@ else ifeq ($(TARGET),c9s)
 	OS := c9s
 	DOCKERFILE ?= Dockerfile.c9s
 	REGISTRY := quay.io/
+else ifeq ($(TARGET),c10s)
+	OS := c10s
+	DOCKERFILE ?= Dockerfile.c10s
+	REGISTRY := quay.io/
 else
 	OS := centos7
 	DOCKERFILE ?= Dockerfile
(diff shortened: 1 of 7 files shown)


phracek:
This is blocked by https://github.com/sclorg/tfaga-wrapper/pull/28

test.sh
```diff
@@ -40,7 +40,7 @@ for dir in ${VERSIONS}; do
   # Kept also IMAGE_NAME as some tests might still use that.
   IMAGE_NAME="$(docker inspect -f "{{.Config.Labels.name}}" "$IMAGE_ID"):$IMAGE_VERSION"
     # shellcheck disable=SC2268
-  if [ "${OS}" == "centos7" ] || [ "${OS}" == "c9s" ] || [ "${OS}" == "fedora" ]; then
+  if [ "${OS}" == "centos7" ] || [ "${OS}" == "c9s" ] || [ "${OS}" == "c10s" ] || [ "${OS}" == "fedora" ]; then
```
fila43, line 43: What about to remove centos7?

test-lib.sh
```diff
@@ -816,6 +816,8 @@ ct_get_public_image_name() {
     public_image_name=$registry/centos7/$base_image_name-${version//./}-centos7:centos7
```
fila43, line 816: Same here

common.mk
```diff
@@ -53,6 +53,10 @@ else ifeq ($(TARGET),c9s)
 	OS := c9s
```
fila43, line 53: I would remove obsolete conditions to make it more readable

test.sh
phracek: This will be removed by a different PR. It will remove RHEL7 and CentOS 7 functionality, as soon as our containers will stop support RHEL7

test-lib.sh
phracek: This will be part of different PR, that does not exist yet ;) RHEL-7 and CentOS 7 deprecation.

fila43:
Then LGTM

PR #380, opened 2024-08-27
Use timeout utility instead of double-sub-shell hell

hhorak:
It turned out that for incremental builds for Node.js, the test always waited 10m no matter that the build finished sooner.

It's not clear why it happened, but when debugging, it seemed like some weired race-condition happened that let some sleeps hang.

Anyway, it's not clear whether there is a reason to not use timeout utility that seems to serve exactly for this reason.

### DIFF
--- a/test-lib.sh
+++ b/test-lib.sh
@@ -74,30 +74,14 @@ ct_build_image_and_parse_id() {
   local ret_val
   local dockerfile
   local command
-  local pid_build
-  local pid_sleep
   local sleep_time
   log_file="$(mktemp)"
   sleep_time="10m"
   [ -n "$1" ] && dockerfile="-f $1"
   command="$(echo "docker build --no-cache $dockerfile $2" | tr -d "'")"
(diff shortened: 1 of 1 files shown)


hhorak:
@SlouchyButton Any idea why `timeout` could be a problem here?

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

SlouchyButton (approved):
Tested on Node.js container on all supported targets and versions. All tests seem to be working.

Code is cleaner and more readable.

phracek (approved):
LGTM. Let's spread it into containers and we will see the updates.

phracek:
[test]

phracek:
[test]

hhorak:
[test]

phracek:
Let's get merge it and distribute to s2i-nodejs-container as first.

PR #394, opened 2025-05-13
Remove deployement config

phracek:
This pull request deletes getting information about DeploymentConfigs.

The deploymentconfigs are deprecated since OpenShift 4.14+.

Information about image_id is not used in the project alone https://github.com/search?q=org%3Asclorg+%3CSAME_IMAGE%3E&type=code.
It is really safe do remove it.

Also function `ct_os_service_image_info` is not used in the whole https://github.com/sclorg organization
See https://github.com/search?q=org%3Asclorg+ct_os_service_image_info&type=code

PR #396, opened 2025-05-28
Remove usage `ct_os_service_image_info` that was removed by

phracek:
This fixes regression caused by https://github.com/sclorg/container-common-scripts/pull/394

The function does not exist in whole this repository now.


phracek:
Bump mypy to version 1.2.0

[test]

PR #398, opened 2025-07-22
Try to push image to OpenShift internal registry more times

phracek:
Try to push image to OpenShift internal registry more times

The error in case push failed.

This fixes error like:

```
Copying blob sha256:415d74f354df4c83b246f01be89785ea693ee130c55b856b2d4d868103e8582e
Error: trying to reuse blob sha256:44212c8bb03bfcbaa3e281221507f684d16690b8d059ac051d6a21652a596210 at destination: checking whether a blob sha256:44212c8bb03bfcbaa3e281221507f684d16690b8d059ac051d6a21652a596210 exists in default-route-openshift-image-registry.apps.ocp4-apps-stacks.hosted.psi.rdu2.redhat.com/sclorg-test-31062/python-312: received unexpected HTTP status: 500 Internal Server Error
warning: Cannot check if git requires authentication.
```

### DIFF
--- a/test-lib-openshift.sh
+++ b/test-lib-openshift.sh
@@ -461,9 +461,15 @@ function ct_os_upload_image() {
     return 1
   fi
   output_name="$OCP4_REGISTER/$namespace/$image_name"
-
-  docker tag "${source_name}" "${output_name}"
-  docker push "${output_name}"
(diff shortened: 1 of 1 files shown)


phracek:
[test]

PR #404, opened 2025-09-04
Remove RHEL7, UBI7 and  scl enable from test suite

phracek:
This pull request is separated to several commit:

- Remove RHEL7, UBI7 and centos sentences in this test suite
- Update compose from RHEL-9.4-Nightly to RHEL-9.6-Nightly
- Remove two functions `ct_check_exec_env_vars` annd `ct_check_scl_enable_vars`
- Turn off failed in case of analysis is sent to logdetective. We do not want to fail in case of something happens
- Replace RHEL8 -> RHEL10 in check_imagestreams.sh

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -59,7 +59,7 @@ jobs:
               tmt_plan="rhel8-docker$"
             else
               if [ "${{ matrix.os }}" == "rhel9" ]; then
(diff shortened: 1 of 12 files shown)


phracek:
[test]

frenzymadness (approved):
The diff looks good to me. What catched my attention is the fact that we have almost  the same function `analyze_logs_by_logdetective()` in `build.sh` and `test.sh`.

PR #406, opened 2025-09-10
Remove usage .version and replace it with $dir variable.

phracek:
Remove usage .version and replace it with $dir variable.

The variable has format like:
1.24, 2.4-micro, 13.

So it can be used for tagging and testing.

As well as for test.sh it can be used


phracek:
[test]

frenzymadness (approved):
The changes look good to me, and the tests in https://github.com/sclorg/s2i-python-container/pull/755 proved that they work.

